PR #285, opened 2026-06-03
docs: Env File Sync guide (crown-jewel rewrite/rename + top-level) + remove dead env_file_pattern

jainal09:
## What & why

**Vault sync is envdrift's crown jewel** — securely syncing `.env` files across a team via your cloud vault + git. This PR makes the guide say that, name itself intuitively, and sit where a newcomer will actually find it. (Tracker issues #6, #7, #8, #9, #10.)

## Guide: `guides/vault-sync.md` → `guides/env-file-sync.md`

- **Full rewrite, two-tier framing**: config-free `vault-push`/`vault-pull` (2-min on-ramp) vs toml-driven team `sync`/`pull` (the real workflow).
- **Documents `vault-pull`** — it was missing from this guide entirely.
- **Precise dotenvx-vs-SOPS scope note** (verified against both backends): Vault Sync distributes the dotenvx `.env.keys` private key, so it's dotenvx-specific; SOPS delegates decryption to its own KMS/age/PGP and doesn't use Vault Sync.
- **Collapsed duplicated cloud-provider setup** (#7): the per-provider CLI-install/login bloat for Azure/AWS/HashiCorp/GCP becomes **one table of the envdrift⇄provider seam** — credential chain envdrift resolves + minimum permissions + links to each provider's own auth docs.
- **Single-provider example configs** (#6) — no stacked `azure`+`gcp` in one example.

## Discoverability

- **Promoted to a top-level nav entry** right after Getting Started (was buried mid-list under "Guides").
- **Old URL preserved**: added the `mkdocs-redirects` plugin so `/guides/vault-sync/` → `/guides/env-file-sync/` (no broken bookmarks). Dep added to `docs.yml` install + `docs`/`dev` extras.
- **All 10 cross-links + labels** updated to the new name.

## Also bundled

- **README**: crown-jewel CTA pointing at the guide (#10, minimal slice) + minor "Solution" wording.
- **playground/envdrift.toml**: single-provider (azure) example.

## Verification

- `mkdocs build --strict` → **exit 0** (no broken links).
- Confirmed the redirect stub is generated: `site/guides/vault-sync/index.html` → `../env-file-sync/`.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Env File Sync Guide covering secure team sync of environment files.
  * Reorganized guides, replacing Vault Sync references with Env File Sync throughout documentation.
  * Updated README with clearer description of envdrift's zero-trust encryption and sync capabilities.

* **Chores**
  * Added URL redirect support to preserve existing documentation links during guide reorganization.
  * Updated build dependencies for documentation generation.

<h3>Greptile Summary</h3>

This PR renames the Vault Sync guide to Env File Sync (`guides/env-file-sync.md`), promotes it to top-level navigation, preserves the old URL via `mkdocs-redirects`, and bundles a small but meaningful code fix: `ephemeral_keys` was being silently dropped when `load_sync_config_and_client` rebuilt `SyncConfig`/`ServiceMapping` from the auto-discovered config path.

- **Guide rewrite**: full two-tier framing (config-free `vault-push`/`vault-pull` vs toml-driven `sync`/`pull`), collapsed per-provider setup into one permissions table, single-provider examples, SOPS scope clarification; all 10 cross-links updated.
- **`ephemeral_keys` propagation fix** in `sync.py` (three call sites: `load_sync_config_and_client`, `pull`, `lock`) with a regression test that exercises both the central flag and per-mapping override.
- **Dead-code removal**: `env_file_pattern` config field and `get_env_file_path()` helper, which were documented but never consumed by any production code path, are removed along with their test assertions.

Safe to merge after the open thread items (dotenvx --rotate dependency and key rotation section clarity) are resolved; the code fix is correct and the redirect infrastructure is properly wired.

The code change (ephemeral_keys propagation) is a genuine bug fix with a regression test that covers both the central flag and per-mapping override. The guide rewrite and navigation restructuring are documentation-only. Two issues flagged in earlier review rounds — the undocumented dotenvx binary dependency in the key-rotation section and a possible config-key mismatch in the HashiCorp tab — remain open and affect user-facing correctness of the guide if unaddressed.

docs/guides/env-file-sync.md — key rotation section and HashiCorp provider tab have open thread items that should be resolved before merging.

| Filename | Overview |
|----------|----------|
| docs/guides/env-file-sync.md | New 560-line guide covering config-free and toml-driven vault sync workflows; comprehensive and well-structured, with previously-flagged issues around dotenvx --rotate dependency and a possible vault_addr/url discrepancy still noted in the review thread. |
| src/envdrift/cli_commands/sync.py | Adds ephemeral_keys propagation to three call sites (load_sync_config_and_client, pull, lock) that previously silently dropped the field when rebuilding SyncConfig from auto-discovered config. |
| src/envdrift/config.py | Removes env_file_pattern field and get_env_file_path() helper — both were documented but never called by any production code path; clean dead-code removal with no behavioral impact. |
| tests/unit/test_sync_config.py | Adds TestLoadSyncConfigEphemeralPropagation regression test covering central flag and per-mapping ephemeral_keys override through the auto-discovery code path; mock patching is correct. |
| mkdocs.yml | Promotes Env File Sync to top-level nav, removes old Vault Sync entry, and configures mkdocs-redirects to preserve /guides/vault-sync/ → /guides/env-file-sync/; syntax is correct. |
| pyproject.toml | Adds mkdocs-redirects>=1.2 to both docs and dev extras, consistent with the docs.yml workflow change. |
| .github/workflows/docs.yml | Adds mkdocs-redirects to the pip install step so the strict build doesn't fail on the new redirects plugin. |
| playground/envdrift.toml | Appends a single-provider Azure vault sync example block; default_vault_name matches the subdomain in vault_url, and the mapping uses a single environment as intended by the guide. |
| README.md | Adds a CTA blockquote pointing to the new Env File Sync guide and tightens the solution description wording; hardcoded absolute URL is appropriate for a README. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[envdrift pull / sync] --> B{Config source}
    B -->|--config path.toml| C[Load explicit TOML]
    B -->|No --config| D[Auto-discover envdrift.toml\nor pyproject.toml]
    B -->|--config pair.txt| E[Legacy pair file]

    C --> F[load_sync_config_and_client]
    D --> F
    E --> F

    F --> G[Build SyncConfig\nwith ephemeral_keys]
    G --> H{ephemeral_keys?}
    H -->|true| I[Fetch key from vault\nPass via env var\nNever write .env.keys]
    H -->|false/nil| J[Fetch key from vault\nWrite to .env.keys]

    I --> K[dotenvx decrypt in place]
    J --> K

    subgraph nav [Nav / URL changes]
        N1["/guides/vault-sync/ (old)"] -->|redirect| N2["/guides/env-file-sync/ (new)"]
        N3["Top-level nav entry\nright after Getting Started"] --> N2
    end
```

<sub>Reviews (4): Last reviewed commit: ["docs(playground): align default\_vault\_na..."](https://github.com/jainal09/envdrift/commit/0d2e0046c46af45738d903d111e3d1921b0aff67) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35378158)</sub>


jainal09:
### Added on this branch: remove dead `env_file_pattern` / `get_env_file_path` (tracker #11)

Investigation while writing the guide surfaced a config option that **does nothing**:

- `env_file_pattern` (default `.env.{environment}`) is parsed from toml and documented, but its only consumer — `get_env_file_path()` — has **zero callers**.
- Sync/vault hardcode `.env.<env>` inline ([sync.py:674](../blob/main/src/envdrift/cli_commands/sync.py#L674)); guard's leak detection is hardcoded to `.env*` ([native.py:236-242](../blob/main/src/envdrift/scanner/native.py#L236)).

Making it *real* would mean threading a custom pattern through the **security-critical scanner** too — otherwise users could name secret files anything and silently lose guard protection. So this removes the dead knob and keeps `.env*` as the single convention the whole system already shares (sync, encrypt, **and** guard).

- Drops the field + toml parsing + `EXAMPLE_CONFIG` line, deletes `get_env_file_path()`, removes matching docs rows/examples and test assertions.
- Backwards-safe: a leftover `env_file_pattern` key was already ignored (unknown keys aren't errors).
- ✅ 1120 unit tests pass; `config.py` 88% coverage; ruff/pyrefly/bandit green.

README.md
```diff
@@ -23,10 +23,10 @@ Sync environment variables across your team. No more "it works on my machine."
 
 ## The Solution
 
-envdrift is an **open-source** CLI that syncs encrypted `.env` files using **your existing cloud vault**.
+envdrift is an **open-source** CLI that encrypts `env` files and syncs them using **your existing cloud vault** and git.
```
jainal09: Typo introduced by this edit: `encrypts \`env\` files` lost the leading dot — the previous wording was `.env` files, and `.env` is the term used everywhere else in the README and guides. Suggest: `encrypts \`.env\` files and syncs them ...`.

docs/guides/env-file-sync.md
```diff
@@ -0,0 +1,550 @@
...
+# (get vault access from your team lead)
+envdrift pull        # syncs every key AND decrypts every .env file — done
+```
+
+That's the whole promise: one command, no Slacked secrets.
+
+### Key rotation
+
+```bash
+dotenvx encrypt .env.production --rotate     # new key in .env.keys
```
jainal09: This rotation snippet drops to raw `dotenvx encrypt .env.production --rotate`, whereas the rest of the guide consistently uses the `envdrift` wrapper (`envdrift encrypt .env.production`). Two issues: (1) it silently assumes the `dotenvx` binary is on PATH, an external dependency the reader hasn't been told to install directly; (2) it's inconsistent with the surrounding flow. If envdrift has no `--rotate` equivalent, add a one-line note like "Key rotation is a dotenvx-native operation (requires the `dotenvx` CLI); envdrift then re-pushes the new key." so the reader knows why the command shape changes.

jainal09 (commented):
## Review — Env File Sync guide (rewrite/rename) + remove dead `env_file_pattern`

**Overview:** Renames/rewrites `guides/vault-sync.md` → `guides/env-file-sync.md` with a clean two-tier framing (config-free `vault-push`/`vault-pull` vs toml-driven `sync`/`pull`), promotes it to top-level nav, preserves the old URL via `mkdocs-redirects`, and updates all 10 cross-links. Bundles a code cleanup removing the unused `env_file_pattern` field + `get_env_file_path()`.

### Code change — verified safe
The `env_file_pattern` / `get_env_file_path` removal is a genuine dead-code cleanup. I grepped the whole tree (src + tests + docs) and there are **zero remaining references** outside the lines this PR deletes — including the example-config docstring, the benchmark fixture, and the unit-test assertions, all of which were updated in lockstep. Nothing reads `config.env_file_pattern` at runtime, so this is clean. Good catch removing it rather than leaving a config knob that does nothing.

### Docs
- Two-tier structure and the collapsed provider credential-chain table read well; documenting `vault-pull` here closes a real gap.
- Redirect wiring is consistent across `mkdocs.yml`, both `pyproject.toml` extras, and `docs.yml`. PR confirms `mkdocs build --strict` exits 0.
- `playground/envdrift.toml` now shows a single-provider vault.sync example — cleaner for the guide, though it drops the only playground example of `[guard]`/`[partial_encryption]`. Intentional per the PR, just noting those features lose their sole runnable sample.

### Two small nits (inline) — neither blocks merge
Seconding Greptile on both: a `.env`→`env` typo in the README and a raw-`dotenvx` invocation in the rotation snippet. Details inline.

Docs-only behavior change + a verified dead-code removal. LGTM after the nits.

docs/guides/env-file-sync.md
```diff
@@ -0,0 +1,550 @@
...
+credential chain**. If your CLI is already logged in, envdrift is already
+authenticated. The table below is the entire envdrift-specific contract: which
+credentials it resolves, and the *minimum* permissions it needs. For installing
+CLIs, creating vaults, and logging in, follow the provider's own docs (linked).
+
+| Provider | envdrift authenticates via | Minimum permissions | Provider auth docs |
+|:--|:--|:--|:--|
+| **Azure Key Vault** | `DefaultAzureCredential` — env vars (`AZURE_CLIENT_ID`/`TENANT_ID`/`CLIENT_SECRET`) → `az login` → managed identity | Secrets: **Get**, **List** | [Azure auth](https://learn.microsoft.com/azure/key-vault/general/authentication) |
+| **AWS Secrets Manager** | boto3 default chain — env vars → `~/.aws/credentials` → IAM role (EC2/ECS/Lambda) | `secretsmanager:GetSecretValue`, `secretsmanager:ListSecrets` | [AWS credentials](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) |
+| **HashiCorp Vault** | `VAULT_ADDR` + `VAULT_TOKEN` (or `~/.vault-token`) | `read`, `list` on the secret path | [Vault auth](https://developer.hashicorp.com/vault/docs/auth) |
```
jainal09: This row describes a credential chain envdrift does not currently implement for HashiCorp. The CLI still requires the Vault URL from `--vault-url` or `[vault.hashicorp].url`, and `HashiCorpVaultClient` only reads the token from `VAULT_TOKEN`; it does not consume `VAULT_ADDR` or `~/.vault-token`. A user with `VAULT_ADDR` set and the Vault CLI logged in will still hit the missing `--vault-url` / missing token path. Please document the actual contract here (`url`/`--vault-url` plus `VAULT_TOKEN`) or add the env/file fallback in the client.

docs/guides/env-file-sync.md
```diff
@@ -0,0 +1,550 @@
...
+
+### Ephemeral keys mode
+
+Fetch keys from the vault and pass them straight to dotenvx via environment
+variables — **never writing `.env.keys` to disk**. Ideal for CI/CD and
+security-sensitive or short-lived environments.
+
+```toml
+[vault.sync]
+ephemeral_keys = true   # central: applies to all mappings
```
jainal09, line 337: This central setting is documented as applying to all mappings, but the documented zero-flag workflow does not preserve it today. In the auto-discovered config path, `load_sync_config_and_client()` rebuilds `SyncConfig(...)` without `ephemeral_keys=vault_sync.ephemeral_keys`; then `pull()` rebuilds `filtered_config` without `ephemeral_keys` again. Result: `envdrift pull` (and auto-discovered `envdrift sync`) can still write `.env.keys` even when `[vault.sync] ephemeral_keys = true`, unless every mapping sets `ephemeral_keys = true` explicitly. Please either carry the central flag through those code paths or document only the per-mapping setting here.

jainal09:
Reviewed PR #285.

Added two inline comments:
- HashiCorp provider setup documents `VAULT_ADDR` / `~/.vault-token`, but the current client requires `url`/`--vault-url` and only reads `VAULT_TOKEN`: https://github.com/jainal09/envdrift/pull/285#discussion_r3349150774
- Central `[vault.sync] ephemeral_keys = true` is documented for `pull`, but the auto-discovered config path and `pull()` filtered config drop the central flag, so `.env.keys` can still be written: https://github.com/jainal09/envdrift/pull/285#discussion_r3349150960

I did not duplicate existing inline comments for the `[vault.hashicorp] vault_addr` typo, playground partial-encryption config regression, README `.env` typo, or the markdownlint failure.

Validation run locally in a temp worktree:
- `uv run --extra dev pytest tests/unit/test_config.py tests/unit/test_sync_config.py tests/unit/test_sync_engine.py -q` -> 119 passed
- `uv run --extra docs mkdocs build --strict` -> passed; redirect page generated at `site/guides/vault-sync/index.html`

Current GitHub checks: `Tests & Coverage` is failing in `make lint-docs` with 16 `MD046/code-block-style` errors in `docs/guides/env-file-sync.md`.

jainal09:
### Round-2 review — all comments addressed (commits `b724c6a` → `0d2e004`)

| Comment | Fix |
|:--|:--|
| markdownlint **MD046** failing CI (coderabbit) | Disabled MD046 — mkdocs content tabs/admonitions indent code blocks, which the rule misreads (consistent with the repo already disabling MD041/MD060/MD024). `b724c6a` |
| HashiCorp **`vault_addr` → `url`** + credential-chain overclaim (codex P2, greptile P1, jainal09) | Config key fixed to `url`; row now states the real contract — URL from `--vault-url`/`[vault.hashicorp].url`, token from `VAULT_TOKEN` only (no `VAULT_ADDR`/`~/.vault-token`). `b724c6a`, `4dce7ce` |
| AWS **`ListSecrets`** not needed (Copilot ×2) | Dropped from the table + IAM snippet — AWS auths via STS `get_caller_identity`, reads via `GetSecretValue`. Azure keeps Get+List (its `authenticate()` lists). `b724c6a` |
| "minimum permissions" **read vs write** (Copilot) | Added a note: table = minimum **read** for sync/pull; `vault-push` needs write. `b724c6a` |
| README **`env` → `.env`** typo (greptile, Copilot, coderabbit, jainal09) | Leading dot restored. `b724c6a` |
| Key rotation uses raw **`dotenvx --rotate`** (greptile, jainal09) | Added a note that rotation is a dotenvx-native op (no `envdrift --rotate`; the `dotenvx` binary is required for that step). `b724c6a` |
| Playground **partial-encryption config** clobbered (codex P2) | Restored `[guard]` + `[partial_encryption]` (Feature 6 depends on them) alongside the single-provider `[vault]`. `b724c6a` |
| Playground **vault identifier mismatch** (coderabbit) | `default_vault_name` aligned to `env-drift`. `0d2e004` |
| **`ephemeral_keys`** central flag silently ignored (jainal09) | **Code fix**: carried the flag through all three `sync.py` rebuild sites (only `from_toml_file` honored it before) + regression test. `4dce7ce` |

CI: markdownlint clean, `mkdocs build --strict` passes, 1121 unit tests pass, changed `sync.py` lines covered.

Issue #348, opened 2026-06-05
Deep main-branch review: security, release, and reliability findings


jainal09:
## Context

A read-only deep review of the `main` branch found several high-confidence issues across guard path handling, sync output, release/install integrity, agent fallback behavior, schema loading, scanner detection, and VS Code extension behavior.

> **✅ Verification audit (main `211999e` / release 10.13.6):** every finding below was re-checked against the current code. **27 of 28 are VALID and present on `main`; 1 is already fixed** (the guard finding, struck below). Duplicate/related cross-references to dedicated issues are annotated inline as `→ #NNN`. Full per-finding evidence is in the verification comment below.

### P1 / high

- ~~`src/envdrift/cli_commands/guard.py:268` and `src/envdrift/cli_commands/guard.py:308` treat `git diff` output as cwd-relative paths. Git returns repo-root-relative paths, so running `envdrift guard --staged` or `envdrift guard --pr-base` from a subdirectory can filter out real changed files and exit successfully without scanning staged/PR secrets.~~ — **✅ ALREADY FIXED on `main`.** `guard.py` resolves `git diff` output against the repo root via `_git_toplevel()` (`git rev-parse --show-toplevel`) for both `--staged` and `--pr-base`, then re-relativizes to cwd. Fixed by #337 (closed #302); the `removeprefix("origin/")` also reflects #319.
- `src/envdrift/sync/operations.py:134` with `src/envdrift/output/rich.py:347` prints the first 32 characters of local and vault values during mismatches. For dotenvx private keys and other vault-backed secrets, this leaks sensitive material into terminals and CI logs.
- `envdrift-agent/internal/encrypt/encrypt.go:79` says Python module fallback is available, but then executes `python lock <file>` instead of `python -m envdrift lock <file>`. Agents on systems without a standalone `envdrift` binary can pass availability checks but fail encryption.
- `.github/workflows/publish.yml:5` uses the tag glob `v*`, which also matches component tags such as `vscode-v*`. A VS Code extension release tag can therefore trigger the Python PyPI publish workflow.
- `.github/workflows/publish.yml:6` allows manual publish dispatch from arbitrary refs and `uv publish` runs unconditionally after build. A manual dispatch can publish a non-release/dev version unintentionally.
- `.github/workflows/agent-release.yml:108` uploads agent binaries but does not generate the `checksums.txt` file that installers expect, causing checksum verification to be skipped in practice. **→ related #334** (producer side).
[...]

Issue #359, opened 2026-06-05
Azure & GCP authenticate() probe with List permission → breaks get/set-only least-privilege identities


jainal09:
## Summary
The Azure and GCP vault backends verify authentication by **listing** secrets. An identity granted only get/access on secrets (a standard least-privilege policy) lacks the list permission, so `authenticate()` fails with a permission error even though `get_secret` — all that sync actually needs — would succeed. AWS deliberately avoids this by probing with STS `get_caller_identity` (see comment at `vault/aws.py:78-79`).

## Location
`src/envdrift/vault/azure.py:83-88`
`src/envdrift/vault/gcp.py:93-105`

## Details
```python
# azure.py
secrets_iter = self._client.list_properties_of_secrets()
next(iter(secrets_iter), None)   # requires "List" → 403 for get/set-only identity
```
```python
# gcp.py
secrets_iter = self._client.list_secrets(request={"parent": self._project_path(), "page_size": 1})
next(iter(secrets_iter), None)   # requires secretmanager.secrets.list → PermissionDenied
```
For Azure a `Get`/`Set`-only access policy and for GCP a service account with only `secretmanager.versions.access` will fail the probe and be reported as an auth failure.

## Impact
Least-privilege credentials that are perfectly capable of `get_secret`/`set_secret` are rejected at `authenticate()`/`ensure_authenticated()`, forcing users to over-grant `List` permissions — contrary to the security posture of the tool and inconsistent with the AWS backend.

## Suggested fix
Probe with a non-list operation (a token/identity check), or treat a permission error on the list-probe as "authenticated but cannot list" rather than a hard failure.

---
*Read-only deep review of `main`. Distinct from #304 (Azure `is_authenticated` after failure) and #305 (HashiCorp exception wrapping).*

jainal09:
Azure half fixed in PR #340 (commit c81e753): authenticate() now treats a 403/Forbidden on the secret-List probe as authenticated-but-cannot-list (least-privilege Get/Set identities accepted), while ClientAuthenticationError still fails. GCP half tracked in PR #393. Load-bearing regression tests added.

jainal09:
GCP half fixed in PR #393 (commit 1d0b53d): authenticate() now treats PermissionDenied on the secrets.list probe as authenticated-but-cannot-list (least-privilege versions.access-only service accounts accepted), while Unauthenticated/DefaultCredentialsError still fail. Azure half tracked in PR #340. Load-bearing regression tests added.

[...]

PR #393, opened 2026-06-06
fix(security): redact secret previews in sync output + enforce GCP project boundary (#348)

jainal09:
## Summary

Two `#348` security findings — a secret leak and a project-boundary bypass.

### Secret-preview leak in sync output
On a value mismatch, the sync diff output **and** the interactive overwrite prompt printed the first 32 chars of both the local and vault values. The vault value is the actual secret — for a 64-hex dotenvx private key that's **half the key** dumped into terminals/CI logs.

**Fix:** replace `preview_value` (raw 32-char prefix) with `redact_value`, emitting a non-reversible, **per-process-salted** BLAKE2b discriminator: `<redacted len=N sha=XXXXXXXX>`.
- No plaintext of either value is emitted (both are masked — the local value can be a secret too).
- Two *different* values still produce *different* output, so the mismatch stays visible; equal values never reach this branch.
- The salt is per-process and not persisted, so the digest is a within-run discriminator only — not an offline brute-force / rainbow target.

### GCP cross-project read/write bypass
`_version_path` (and `_secret_id`/`_secret_path`) accepted a fully-qualified `projects/<other>/secrets/...` name **unchanged**, bypassing the bound `project_id` — a caller could read (or silently write) another project's secret if credentials allowed.

**Fix:** a shared `_validate_project` guard raises `VaultError` when a fully-qualified name targets a project other than the bound one (case-variant `projects/mine`, prefix-collision `projects/MINE-EVIL`, and malformed `projects/` inputs all rejected). `get_secret` validates **before** any `access_secret_version` call — no cross-project read reaches the network.

## Verification
- **No-leak proof:** rendered output contains no 8+ char substring of a 64-hex / AWS-key / JWT secret, yet a genuine mismatch is still conveyed.
- **Boundary proof:** cross-project names raise on every path builder; `get_secret` never calls the API for a foreign project; same-project access still works.
- Changed-line coverage: `gcp.py` 99%, `operations.py` 93%, `engine.py` 81%. Gates green (ruff/format/pyrefly/bandit); 47 focused tests pass.

Addresses the secret-leak + cross-project findings of #348.

## Summary by CodeRabbit

* **Security Improvements**
  * Non-reversible, per-run redaction for secret previews that preserves length info and explicitly marks empty/missing values to prevent leaking substrings.

* **Bug Fixes**
  * GCP Secret Manager integration now validates resource name formats and enforces bound-project restrictions; authentication error handling refined to distinguish permission vs credential failures.

* **Tests**
  * Added coverage ensuring console output and prompts never expose raw secret substrings.

<h3>Greptile Summary</h3>

This PR closes two security findings from #348: a secret-preview leak in sync diff/prompt output, and a GCP cross-project boundary bypass in the Secret Manager client.

- **Secret redaction (`operations.py`, `engine.py`):** `preview_value` (raw 32-char prefix) is replaced by `redact_value`, which emits a per-process-salted BLAKE2b-4 discriminator (`<redacted len=N sha=XXXXXXXX>`). The per-process salt means digests cannot be correlated or brute-forced across runs; distinct values still produce distinct output so mismatches remain visible. The `local_value` guard in `engine.py` is also tightened from a truthy check to `is not None`, correctly treating an empty local value as a value rather than \"no key.\"
- **GCP project boundary (`gcp.py`):** A new `_validate_project` guard, keyed off a compiled `_QUALIFIED_NAME_RE` regex, is inserted into all three path-builder methods and gates every `get_secret` call before any network I/O. Malformed or cross-project names raise `VaultError` immediately. The `authenticate()` method is also corrected to accept `PermissionDenied` (least-privilege SA holding only `secretmanager.versions.access`) while still failing hard on `Unauthenticated`.
- **Tests:** Fake GCP exception classes are restructured to mirror the real SDK's MRO, a new ordering-regression test pins the `except` clause sequence, and sliding-window no-leak assertions cover both the renderer and the interactive prompt path.

Safe to merge. Both security fixes are narrowly scoped, internally consistent, and backed by thorough tests covering the exact failure paths they close.

The secret-redaction change is a straightforward function swap with no new state; the only behavioral difference beyond the security fix is the `is not None` guard in `engine.py`, which correctly handles empty-string local values that the old truthy check would have silently dropped. The GCP boundary enforcement inserts validation at all three path-builder entry points and in `get_secret`, so no call path to the GCP API can bypass the project check. The fake exception hierarchy in the tests now mirrors the real SDK's MRO, and the new ordering-regression test would catch a future reordering of the `except` clauses. No pre-existing behaviour is regressed.

| Filename | Overview |
|----------|----------|
| src/envdrift/sync/operations.py | Replaces `preview_value` (raw 32-char prefix) with `redact_value`: a per-process-salted BLAKE2b-4 discriminator that emits `<redacted len=N sha=XXXXXXXX>` for non-empty values and `<empty>` for the empty-string case. None passes through unchanged. Logic is correct. |
| src/envdrift/sync/engine.py | Swaps `preview_value` for `redact_value` at the two call sites; the `local_value` guard is tightened from the truthy check `if local_value` to the explicit `if local_value is not None`, correctly treating the empty-string case as a value rather than "no local key". |
| src/envdrift/vault/gcp.py | Adds `_validate_project` guarding all three path-builder helpers (`_secret_id`, `_secret_path`, `_version_path`) and `get_secret`; the compiled `_QUALIFIED_NAME_RE` regex enforces canonical shape and project match before any GCP call. `PermissionDenied` in `authenticate()` is now accepted (least-privilege SA fix) while `Unauthenticated` still raises `AuthenticationError`. |
| tests/unit/test_vault_gcp_paths.py | New file: pure-string path-logic regression tests using a real `GCPSecretManagerClient` (no mocking); covers bare name, same-project FQN, cross-project rejection in all three path builders, malformed names, and a spy-client proof that `get_secret` never reaches `access_secret_version` for a foreign project. |
| tests/unit/test_vault_gcp.py | Fake exception hierarchy fixed to mirror the real SDK MRO (specific errors now subclass `DummyGoogleAPICallError`); `test_authenticate_except_clause_ordering` pins that specific-before-generic ordering is preserved. Splits the old `test_authenticate_permission_denied` into two tests with clear PermissionDenied-accepted / Unauthenticated-rejected semantics. |
| tests/unit/coverage/test_cov_sync_engine.py | Adds regression test driving the real `SyncEngine` mismatch → `prompt_callback` path with 64-hex secrets; asserts no 8-char window of either secret leaks into the prompt message and that the redacted discriminator tokens are present. |
| tests/unit/test_rich_output.py | Adds renderer no-leak test that feeds `redact_value`-processed previews into `print_service_sync_status` and confirms no 8-char substring of either raw secret appears in rendered output while the mismatch is still signalled. |
| tests/unit/test_sync_operations.py | Replaces `TestPreviewValue` with `TestRedactValue`; new suite covers None pass-through, empty-string marker, length in output, no-8-char-window leak, distinct-value discrimination, same-value idempotency, and same-length-different-value discrimination. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Caller
    participant SyncEngine
    participant redact_value
    participant GCPClient as GCPSecretManagerClient
    participant _validate_project
    participant GCP_API as GCP Secret Manager API

    Caller->>SyncEngine: sync_all()
    SyncEngine->>GCPClient: get_secret(name)
    GCPClient->>GCPClient: ensure_authenticated()
    GCPClient->>GCPClient: _version_path(name)
    GCPClient->>_validate_project: validate(name)
    alt cross-project or malformed name
        _validate_project-->>GCPClient: raise VaultError
        GCPClient-->>SyncEngine: VaultError (no network call)
    else same-project / bare name
        _validate_project-->>GCPClient: OK
        GCPClient->>GCP_API: access_secret_version(version_path)
        GCP_API-->>GCPClient: SecretValue
        GCPClient-->>SyncEngine: SecretValue
    end
    SyncEngine->>redact_value: redact_value(vault_value)
    redact_value-->>SyncEngine: "<redacted len=N sha=XXXXXXXX>"
    SyncEngine->>redact_value: redact_value(local_value)
    redact_value-->>SyncEngine: "<redacted len=M sha=YYYYYYYY>"
    SyncEngine-->>Caller: ServiceSyncResult (no plaintext in previews)
```

<sub>Reviews (8): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/security-le..."](https://github.com/jainal09/envdrift/commit/a013c44601eef82c62a792bab24ae20e687b89e2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36000134)</sub>

### DIFF
--- a/src/envdrift/sync/engine.py
+++ b/src/envdrift/sync/engine.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING
 
 from envdrift.env_files import detect_env_file, resolve_mapping_env_file
 from envdrift.sync.config import ServiceMapping, SyncConfig
-from envdrift.sync.operations import EnvKeysFile, ensure_directory, preview_value
+from envdrift.sync.operations import EnvKeysFile, ensure_directory, redact_value
 from envdrift.sync.result import (
     DecryptionTestResult,
     ServiceSyncResult,
@@ -110,7 +110,7 @@ class SyncEngine:
 
             # Fetch secret from vault
             vault_value = self._fetch_vault_secret(mapping)
-            vault_preview = preview_value(vault_value)
+            vault_preview = redact_value(vault_value)
 
             # Check for ephemeral mode - skip local file operations
             is_ephemeral = self.config.get_effective_ephemeral(mapping)
@@ -142,7 +142,7 @@ class SyncEngine:
             env_keys_path = mapping.folder_path / self.config.env_keys_filename
             env_keys_file = EnvKeysFile(env_keys_path)
             local_value = env_keys_file.read_key(effective_key_name)
-            local_preview = preview_value(local_value) if local_value else None
+            local_preview = redact_value(local_value) if local_value else None
 
             # Compare values
             if local_value is None:
--- a/src/envdrift/sync/engine.py
+++ b/src/envdrift/sync/engine.py
@@ -142,7 +142,7 @@ class SyncEngine:
             env_keys_path = mapping.folder_path / self.config.env_keys_filename
             env_keys_file = EnvKeysFile(env_keys_path)
             local_value = env_keys_file.read_key(effective_key_name)
-            local_preview = redact_value(local_value) if local_value else None
+            local_preview = redact_value(local_value) if local_value is not None else None
 
             # Compare values
             if local_value is None:
--- a/src/envdrift/vault/gcp.py
+++ b/src/envdrift/vault/gcp.py
@@ -130,12 +130,21 @@ class GCPSecretManagerClient(VaultClient):
             )
             next(iter(secrets_iter), None)
         except DefaultCredentialsError as e:
+            # No usable credentials at all (no ADC, bad GOOGLE_APPLICATION_CREDENTIALS,
+            # etc.) — a genuine authentication failure.
             self._client = None
             raise AuthenticationError(f"GCP authentication failed: {e}") from e
-        except (
-            google_exceptions.PermissionDenied,
-            google_exceptions.Unauthenticated,
-        ) as e:
+        except google_exceptions.PermissionDenied:
+            # The credential authenticated successfully but lacks
+            # `secretmanager.secrets.list`. A least-privilege service account that
+            # only holds `secretmanager.versions.access` (enough for get_secret,
+            # which is all sync needs) hits this on the list probe. Treat it as
+            # authenticated-but-cannot-list: keep the client. If a specific secret
+            # genuinely can't be read, get_secret() surfaces a clear error later.
+            pass
+        except google_exceptions.Unauthenticated as e:
+            # The credential itself is invalid/expired — a genuine auth failure,
+            # distinct from PermissionDenied (authenticated, missing list permission).
             self._client = None
             raise AuthenticationError(f"GCP authentication failed: {e}") from e
         except google_exceptions.GoogleAPICallError as e:
--- a/src/envdrift/vault/gcp.py
+++ b/src/envdrift/vault/gcp.py
@@ -62,7 +62,29 @@ class GCPSecretManagerClient(VaultClient):
     def _project_path(self) -> str:
         return f"projects/{self.project_id}"
 
+    def _validate_project(self, name: str) -> None:
+        """Reject fully-qualified resource names that target a different project.
+
+        A bare secret name is left to resolve under the bound project. A
+        ``projects/<P>/secrets/...`` name is only allowed when ``<P>`` matches the
+        project this backend is bound to, preventing a caller-supplied name from
+        crossing the configured project boundary.
+        """
+        if not name.startswith("projects/"):
+            return
+        parts = name.split("/")
+        if len(parts) < 2 or not parts[1]:
+            raise VaultError(f"Malformed GCP secret resource name: {name!r}")
+        requested_project = parts[1]
+        if requested_project != self.project_id:
+            raise VaultError(
+                f"Secret resource name targets project {requested_project!r}, "
+                f"but this backend is bound to project {self.project_id!r}. "
+                f"Cross-project access is not allowed."
+            )
+
     def _secret_id(self, name: str) -> str:
+        self._validate_project(name)
         if name.startswith("projects/"):
             parts = name.split("/")
             if "secrets" in parts:
@@ -72,9 +94,11 @@ class GCPSecretManagerClient(VaultClient):
         return name
 
     def _secret_path(self, name: str) -> str:
+        self._validate_project(name)
         return f"{self._project_path()}/secrets/{self._secret_id(name)}"
 
     def _version_path(self, name: str, version: str = "latest") -> str:
+        self._validate_project(name)
         if name.startswith("projects/") and "/versions/" in name:
             return name
         if name.startswith("projects/") and "/secrets/" in name:
--- a/src/envdrift/vault/gcp.py
+++ b/src/envdrift/vault/gcp.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import contextlib
+import re
 from typing import Any
 
 from envdrift.vault.base import (
@@ -36,6 +37,13 @@ def _get_gcp_modules() -> tuple[Any, Any]:
     return _secretmanager, _google_exceptions
 
 
+# Canonical GCP Secret Manager resource name: projects/<P>/secrets/<S> optionally
+# followed by /versions/<V>. Each segment is non-empty and slash-free. A bare
+# secret name (no `projects/` prefix) is handled separately and resolves under the
+# bound project. Declaring the shape once here keeps validation in a single place.
+_QUALIFIED_NAME_RE = re.compile(r"^projects/(?P<project>[^/]+)/secrets/[^/]+(?:/versions/[^/]+)?$")
+
+
 class GCPSecretManagerClient(VaultClient):
     """GCP Secret Manager implementation.
 
@@ -63,19 +71,22 @@ class GCPSecretManagerClient(VaultClient):
         return f"projects/{self.project_id}"
 
     def _validate_project(self, name: str) -> None:
-        """Reject fully-qualified resource names that target a different project.
+        """Reject resource names that are malformed or target a different project.
 
-        A bare secret name is left to resolve under the bound project. A
-        ``projects/<P>/secrets/...`` name is only allowed when ``<P>`` matches the
-        project this backend is bound to, preventing a caller-supplied name from
-        crossing the configured project boundary.
+        A bare secret name (no ``projects/`` prefix) is left to resolve under the
+        bound project. A fully-qualified name must match the canonical shape
+        ``projects/<P>/secrets/<S>`` optionally followed by ``/versions/<V>``, and
+        ``<P>`` must match the project this backend is bound to. This prevents a
+        caller-supplied name from being silently rewritten into a synthetic path
+        (e.g. ``projects/<P>`` or ``projects/<P>/other/<S>``) or from crossing the
+        configured project boundary.
         """
         if not name.startswith("projects/"):
             return
-        parts = name.split("/")
-        if len(parts) < 2 or not parts[1]:
+        match = _QUALIFIED_NAME_RE.match(name)
+        if match is None:
             raise VaultError(f"Malformed GCP secret resource name: {name!r}")
-        requested_project = parts[1]
+        requested_project = match.group("project")
         if requested_project != self.project_id:
             raise VaultError(
                 f"Secret resource name targets project {requested_project!r}, "
--- a/tests/unit/test_vault_gcp.py
+++ b/tests/unit/test_vault_gcp.py
@@ -18,23 +18,33 @@ class DummyGcpError(Exception):
 
 
 class DummyGoogleAPICallError(DummyGcpError):
-    """Fake GoogleAPICallError."""
+    """Fake GoogleAPICallError.
+
+    The real ``google.api_core.exceptions.GoogleAPICallError`` is the generic
+    base of the concrete status errors below (PermissionDenied, Unauthenticated,
+    NotFound, AlreadyExists all subclass it via ClientError). Mirroring that MRO
+    here is load-bearing: ``authenticate()`` relies on the *specific* clauses
+    (PermissionDenied -> accept, Unauthenticated -> reject) preceding the generic
+    ``except GoogleAPICallError`` clause. If these fakes were flat siblings, a
+    reordering regression (generic clause first) would still pass the tests while
+    re-breaking #359 in the real SDK. See test_authenticate_except_clause_ordering.
+    """
 
 
-class DummyPermissionDeniedError(DummyGcpError):
-    """Fake PermissionDenied."""
+class DummyPermissionDeniedError(DummyGoogleAPICallError):
+    """Fake PermissionDenied (a GoogleAPICallError subclass, as in the real SDK)."""
 
 
-class DummyUnauthenticatedError(DummyGcpError):
-    """Fake Unauthenticated."""
+class DummyUnauthenticatedError(DummyGoogleAPICallError):
+    """Fake Unauthenticated (a GoogleAPICallError subclass, as in the real SDK)."""
 
 
-class DummyNotFoundError(DummyGcpError):
-    """Fake NotFound."""
+class DummyNotFoundError(DummyGoogleAPICallError):
+    """Fake NotFound (a GoogleAPICallError subclass, as in the real SDK)."""
 
 
-class DummyAlreadyExistsError(DummyGcpError):
-    """Fake AlreadyExists."""
+class DummyAlreadyExistsError(DummyGoogleAPICallError):
+    """Fake AlreadyExists (a GoogleAPICallError subclass, as in the real SDK)."""
 
 
 class DummyDefaultCredentialsError(Exception):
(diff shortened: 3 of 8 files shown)


src/envdrift/sync/engine.py
```diff
@@ -142,7 +142,7 @@ def _sync_service(self, mapping: ServiceMapping) -> ServiceSyncResult:
             env_keys_path = mapping.folder_path / self.config.env_keys_filename
             env_keys_file = EnvKeysFile(env_keys_path)
             local_value = env_keys_file.read_key(effective_key_name)
-            local_preview = preview_value(local_value) if local_value else None
+            local_preview = redact_value(local_value) if local_value else None
```
jainal09: 🤖 **envdrift review-bot** · medium · test-gap


Issue #348 names TWO secret-leak surfaces in sync: the diff/status output AND the interactive overwrite prompt ('the sync diff output and the interactive overwrite prompt printed the first 32 chars of both the local and vault values'). The PR fixes both source sites (engine.py:113 vault_preview=redact_value(...), engine.py:145 local_preview=redact_value(...); the prompt_msg at engine.py ~193-199 interpolates these vars). But the new tests only cover the redact_value function in isolation (test_sync_operations.py) and the rich RENDERER (test_rich_output.py). No new test drives the engine's prompt_msg construction. The only existing prompt-path test, tests/unit/coverage/test_cov_sync_engine.py::test_prompt_accepts_update, uses short values ('new_secret'/'old_secret', <32 chars) and asserts only `'Value mismatch for k' in received[0]` — it never asserts the previews are redacted. So if the prompt path regressed (e.g. someone re-introduced a raw preview at the prompt site, or a future change interpolated local_value/vault_value instead of the redacted vars), no test in the suite would catch it. Per CLAUDE.md ('every bug fix ships a real regression test'), the second leak surface from #348 is effectively unverified. The fix itself is correct; only its verification is incomplete. Recommend a test driving SyncEngine with prompt_callback and realistic 64-hex secrets that asserts no >=8-char window of either secret appears in the captured prompt message.

**Evidence:** OLD operations.py preview_value: `if len(value) <= length: return value` (short values returned verbatim) and `return f"{value[:length]}..."` (long values leak first 32 chars). The pre-existing prompt test test_cov_sync_engine.py::test_prompt_accepts_update feeds value='new_secret'/'old_secret' (both <32 chars) so on OLD code the prompt message contained those secrets verbatim, yet the test only asserts `assert "Value mismatch for k" in received[0]` — it would PASS on old (leaking) code AND new (redacted) code, so it does not exercise the prompt-leak bug. The PR adds NO new test that captures engine prompt_msg output; new tests touch only test_sync_operations.py (function), test_rich_output.py (renderer), and test_vault_gcp*.py (GCP).

jainal09:
🤖 **envdrift review-bot** — test-fidelity re-review of commit 39c0bb5


Automated re-review focused on whether the #348 fix ships real regression coverage. 1 finding, posted inline.

- **medium · test-gap** — [Interactive-prompt leak surface (one of the two named in #348) has no regression test](https://github.com/jainal09/envdrift/pull/393#discussion_r3367165501) (`src/envdrift/sync/engine.py:145`). The PR fixes both leak surfaces from #348, but only the diff/status renderer and the `redact_value` function get new tests. The interactive overwrite prompt path is only touched by the pre-existing `test_cov_sync_engine.py::test_prompt_accepts_update`, which is byte-identical to main, feeds <32-char values, and asserts only `'Value mismatch for k' in received[0]` — so it passes on both leaking (old) and redacted (new) code and does not prove the prompt fix. Per CLAUDE.md ("every bug fix ships a real regression test"), the second leak surface is effectively unverified.

tests/unit/test_vault_gcp_paths.py
```diff
@@ -0,0 +1,88 @@
...
+    without live credentials -- the guard runs before ``access_secret_version``.
+
+    ``ensure_authenticated`` would normally need a client; assert the validator
+    fires either via that path or the resource-name guard. We isolate the guard
+    by calling the version-path builder ``get_secret`` uses.
+    """
+    c = _client()
+    cross = "projects/victim-project/secrets/db-password/versions/latest"
+    with pytest.raises(VaultError):
+        c._version_path(cross)
```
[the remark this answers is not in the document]
jainal09: Fixed in ce0ca54: the test now drives the real `get_secret` with a spy client and asserts `access_secret_version` is never reached for a cross-project name — not just `_version_path`.

src/envdrift/sync/engine.py
jainal09: Both #348 leak surfaces are closed: the interactive overwrite prompt (engine.py ~197) embeds the SAME `local_preview`/`vault_preview`, which are now `redact_value(...)` outputs — so no plaintext reaches the prompt either. The redaction is unit-tested in test_sync_operations.py.

src/envdrift/sync/engine.py
```diff
@@ -142,7 +142,7 @@ def _sync_service(self, mapping: ServiceMapping) -> ServiceSyncResult:
             env_keys_path = mapping.folder_path / self.config.env_keys_filename
             env_keys_file = EnvKeysFile(env_keys_path)
             local_value = env_keys_file.read_key(effective_key_name)
-            local_preview = preview_value(local_value) if local_value else None
+            local_preview = redact_value(local_value) if local_value is not None else None
```
jainal09, line 145: 🤖 **envdrift review-bot** · medium · test-gap


This commit's core engine change is `local_preview = redact_value(local_value) if local_value is not None else None` at `src/envdrift/sync/engine.py:145` (was `if local_value`). The bug being fixed: when a key exists locally with an empty value (`""`), the old truthiness guard set `local_preview = None`, so on a mismatch the rich renderer (`src/envdrift/output/rich.py:345`: `if result.local_value_preview and result.vault_value_preview:`) skipped printing BOTH the Local and Vault discriminators — hiding the empty-vs-secret mismatch entirely.

Per CLAUDE.md ("every bug fix ships a real regression test" that FAILS without the fix and PASSES with it), this fix is unverified at the layer it was made. The PR adds only a function-level test, `tests/unit/test_sync_operations.py::TestRedactValue::test_empty_string_marked_empty`, which asserts `redact_value("") == "<empty>"`. That test never touches `engine.py:145` and passes regardless of the truthiness-vs-`is not None` guard, so it cannot catch a regression of this branch. No test in `tests/unit/coverage/test_cov_sync_engine.py` (14 `_sync_service` tests) drives `_sync_service` with an empty-string local value: I grepped the file and integration tests and found no empty-local-value case.

Impact: a future refactor reverting `is not None` to `if local_value` would silently reintroduce the leak-discriminator-hiding bug with a green suite. Direction: add a focused test that runs `_sync_service` (mismatch path) with a local `.env.keys` value of `""` against a non-empty vault value and asserts `result.local_value_preview == "<empty>"` (it is `None` on pre-fix code, so the assertion fails without the fix and passes with it). This is distinct from cubic's already-addressed bug comment and from my prior edbot prompt-leak comment.

**Evidence:** engine.py:145 (head): `local_preview = redact_value(local_value) if local_value is not None else None`. rich.py:345: `if result.local_value_preview and result.vault_value_preview:` (None preview => mismatch lines skipped). Only new empty-value test is test_sync_operations.py::test_empty_string_marked_empty: `assert redact_value("") == "<empty>"` — does not exercise engine.py. grep of tests/unit/coverage/test_cov_sync_engine.py shows no empty-string local_value scenario.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit ce0ca54


The empty-value redaction fix at `engine.py:145` is correct, but its only new test exercises `redact_value` in isolation — nothing drives `_sync_service` with an empty-string local value, so a future revert of the `is not None` guard would pass the suite while re-hiding the empty-vs-secret mismatch.

**Findings (1):**
- [medium] Empty-value redaction fix (engine.py:145 `is not None`) has no engine-level regression test that fails without it — `src/envdrift/sync/engine.py:145` (inline)

src/envdrift/sync/engine.py
jainal09: 🤖 **envdrift review-bot**


Agreed on the fix, and verified against ce0ca54: the prompt at `engine.py:193-198` interpolates `local_preview`/`vault_preview`, both `redact_value(...)` outputs, so no plaintext reaches the prompt. That part is correct. My finding was about *verification*, not the fix: `redact_value`'s unit tests in `test_sync_operations.py` exercise the function in isolation, not the engine's `prompt_msg` construction. The PR touches no engine prompt test — `gh pr diff 393 --name-only` doesn't include `tests/unit/coverage/test_cov_sync_engine.py`, and the only prompt test there (`test_prompt_accepts_update`) feeds short values and asserts solely `"Value mismatch for k" in received[0]`, so a regression that re-interpolated `local_value`/`vault_value` at the prompt site would go uncaught. Marking **partial**: code correct, residual (low-severity) gap is one engine-level test driving `prompt_callback` with a realistic 64-hex secret that asserts no >=8-char window of either secret appears in the captured message.

jainal09: Fixed in 4b48c3e. Added `test_prompt_message_redacts_realistic_secrets` (in `tests/unit/coverage/test_cov_sync_engine.py`) that drives the real `SyncEngine` mismatch → `prompt_callback` path with distinct 64-hex secrets, captures the prompt message, and asserts no ≥8-char window of either raw secret reaches the prompt while the `<redacted len=… sha=…>` marker is present for both previews. Proven load-bearing: when the prompt site interpolates the raw `value[:32]` it FAILS (`vault secret window 'a0123456' leaked into prompt message`), and PASSES with the redacted form on HEAD.

src/envdrift/vault/gcp.py
```diff
@@ -62,7 +62,29 @@ def __init__(self, project_id: str):
...
+        parts = name.split("/")
+        if len(parts) < 2 or not parts[1]:
+            raise VaultError(f"Malformed GCP secret resource name: {name!r}")
+        requested_project = parts[1]
+        if requested_project != self.project_id:
+            raise VaultError(
+                f"Secret resource name targets project {requested_project!r}, "
+                f"but this backend is bound to project {self.project_id!r}. "
+                f"Cross-project access is not allowed."
+            )
```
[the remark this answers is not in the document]
jainal09: Fixed in 350aa83. `_validate_project` now enforces the full canonical shape `projects/<P>/secrets/<S>[/versions/<V>]`: it hard-fails when the name isn't `projects` / non-empty project / `secrets` / non-empty secret, and (when a version tail is present) anything other than exactly `/versions/<non-empty>`. So `projects/my-project`, `projects/my-project/other/x`, `projects/my-project/secrets/`, and `projects/my-project/secrets/x/versions` all raise `VaultError` instead of being rewritten into a synthetic path. Added a parametrized regression test (`test_malformed_fully_qualified_names_are_rejected`) proven to fail on the pre-fix guard (6 failed → 6 passed).

src/envdrift/sync/engine.py
jainal09: 🤖 **envdrift review-bot**


You're right — resolved. Verified against `350aa83`: `test_prompt_message_redacts_realistic_secrets` (tests/unit/coverage/test_cov_sync_engine.py:200) drives the real `SyncEngine` mismatch → `prompt_callback` path (not `redact_value` in isolation or the renderer), captures the actual `prompt_msg` built at engine.py:194-200, and runs an 8-char sliding-window check over both raw 65-char secrets plus asserting both `<redacted len=…>` markers are present. I confirmed it's load-bearing: replaying the message under the old `preview_value` (`value[:32]`) leaks `'a0123456'` / `'bfedcba9'` and fails the window assertion, while HEAD's redacted form passes — exactly the leak surface #2 from #348 that my finding said was unverified. Thanks for closing it with a true regression test.

src/envdrift/vault/gcp.py
```diff
@@ -62,7 +62,41 @@ def __init__(self, project_id: str):
...
+        # Canonical shape: projects/<P>/secrets/<S>[/versions/<V>]
+        if (
+            len(parts) < 4
+            or parts[0] != "projects"
+            or not parts[1]
+            or parts[2] != "secrets"
+            or not parts[3]
+        ):
+            raise VaultError(f"Malformed GCP secret resource name: {name!r}")
+        if len(parts) > 4 and (len(parts) != 6 or parts[4] != "versions" or not parts[5]):
```
jainal09: Addressed in 5351c73: `_validate_project` no longer uses the two multi-clause conditionals — the canonical resource shape `projects/<P>/secrets/<S>[/versions/<V>]` is now expressed as a single anchored regex (`_QUALIFIED_NAME_RE`) declared once at module scope, leaving just `if match is None: raise` and the unchanged cross-project check. Complexity is well under threshold and the shape lives in one place. (All 32 gcp tests + full non-integration suite still green.)

tests/unit/test_vault_gcp.py
```diff
@@ -199,11 +202,37 @@ def test_authenticate_default_credentials_error(self, mock_gcp):
...
+        )
+        mock_gcp._secretmanager.SecretManagerServiceClient.return_value = mock_client
+
+        client = mock_gcp.GCPSecretManagerClient(project_id="my-project")
+        client.authenticate()
+
+        assert client.is_authenticated() is True
+        assert client._client is mock_client
+
+    def test_authenticate_unauthenticated_still_fails(self, mock_gcp):
```
jainal09, line 236: 🤖 **envdrift review-bot** · medium · test-gap


This commit makes `authenticate()`'s exception ordering security-critical: it splits the old single `except (PermissionDenied, Unauthenticated)` block into three branches with semantically OPPOSITE outcomes — `PermissionDenied` -> accept (`pass`, keep client), `Unauthenticated` -> reject (`AuthenticationError`), `GoogleAPICallError` -> `VaultError`. In the real `google.api_core.exceptions`, both `PermissionDenied` (via `Forbidden`/`ClientError`) and `Unauthenticated` (via `ClientError`) subclass `GoogleAPICallError`, so this only works because the two specific clauses precede the generic `GoogleAPICallError` clause.

The new regression tests can't protect that ordering. In the `mock_gcp` fixture the fakes are flat siblings — `DummyPermissionDeniedError(DummyGcpError)` and `DummyUnauthenticatedError(DummyGcpError)` are NOT subclasses of `DummyGoogleAPICallError(DummyGcpError)`. So if someone reordered the production `except` clauses to put `GoogleAPICallError` first, the real code would catch `PermissionDenied` as a generic error (re-breaking #359 by raising `VaultError`) and — more dangerously — catch `Unauthenticated` as `VaultError` instead of `AuthenticationError`, mis-categorizing a genuine credential failure. Yet `test_authenticate_permission_denied_on_list_probe_accepted` and `test_authenticate_unauthenticated_still_fails` would both keep passing, because under the mock hierarchy a `GoogleAPICallError`-first clause does not match the flat-sibling Dummy exceptions. I reproduced this divergence in a scratch script: under the mock hierarchy a `GoogleAPICallError`-first order returns "accepted" for PermissionDenied, while under the real subclass hierarchy it returns "VaultError". Impact: the load-bearing tests for the accept-vs-reject split give false confidence against a reordering regression. Direction: make the Dummy fakes mirror the SDK MRO (`class DummyPermissionDeniedError(DummyGoogleAPICallError)`, `class DummyUnauthenticatedError(DummyGoogleAPICallError)`) so the except-ordering contract is actually exercised.

**Evidence:** test_vault_gcp.py:20-29 — `class DummyGoogleAPICallError(DummyGcpError)`, `class DummyPermissionDeniedError(DummyGcpError)`, `class DummyUnauthenticatedError(DummyGcpError)` (all subclass DummyGcpError, NOT DummyGoogleAPICallError). gcp.py authenticate(): `except google_exceptions.PermissionDenied: ... pass` then `except google_exceptions.Unauthenticated as e: raise AuthenticationError(...)` then `except google_exceptions.GoogleAPICallError as e: raise VaultError(...)`. Real SDK: PermissionDenied and Unauthenticated both subclass GoogleAPICallError.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 1d0b53d


The production `authenticate()` exception-ordering split (PermissionDenied -> accept, Unauthenticated -> reject, GoogleAPICallError -> VaultError) is correct at HEAD, but the new regression tests cannot protect that ordering: the mock fakes are flat siblings instead of mirroring the real SDK MRO, so a clause-reorder regression would slip past a green suite.

**Findings (1):**
- [medium] GCP regression tests can't catch an except-clause reordering: mock exceptions don't subclass GoogleAPICallError like the real SDK — `tests/unit/test_vault_gcp.py:226` (inline)

tests/unit/test_vault_gcp.py
jainal09: Fixed in fdabb22.

The `mock_gcp` fixture's fake exceptions now mirror the real `google.api_core.exceptions` MRO: `DummyPermissionDeniedError`, `DummyUnauthenticatedError`, `DummyNotFoundError`, and `DummyAlreadyExistsError` all subclass `DummyGoogleAPICallError` (matching the SDK, where they descend from `GoogleAPICallError` via `ClientError`) instead of being flat siblings.

Added `test_authenticate_except_clause_ordering`: with both `PermissionDenied` and `Unauthenticated` now `GoogleAPICallError` subclasses, it asserts `PermissionDenied` is still ACCEPTED (client retained) and `Unauthenticated` still maps to `AuthenticationError` (not `VaultError`) — proving the specific-before-generic clause order in `authenticate()` is load-bearing.

Proved load-bearing: temporarily reordering the production `except` clauses to put `except GoogleAPICallError` first (NOT committed — `src/` diff is clean) makes the new ordering test plus the existing `..._permission_denied_on_list_probe_accepted` and `..._unauthenticated_still_fails` tests fail (PermissionDenied swallowed as VaultError):
- with reorder: `3 failed, 18 deselected`
- restored order: `3 passed, 18 deselected`

Full non-integration suite stays green: `2158 passed, 2 skipped, 280 deselected`. Production code unchanged (test-only).

jainal09: 🤖 **envdrift review-bot**


You're right — verified against fdabb22 and conceding. The `mock_gcp` fakes now mirror the real SDK MRO: `DummyPermissionDeniedError`/`DummyUnauthenticatedError`/`DummyNotFoundError`/`DummyAlreadyExistsError` all subclass `DummyGoogleAPICallError` (test_vault_gcp.py:34-47), and the new `test_authenticate_except_clause_ordering` (test_vault_gcp.py:269) pins the contract with `issubclass(PermissionDenied, GoogleAPICallError)` sanity asserts plus the accept/reject outcomes. I reproduced the protection independently: with these subclass fakes, a generic-first `except GoogleAPICallError` reorder makes `PermissionDenied -> VaultError` and `Unauthenticated -> VaultError`, failing both assertions — whereas the old flat siblings would have let a generic-first clause fall through and still return "accepted" (the false confidence I flagged). The fix commit is test-only (production `authenticate()` ordering at gcp.py:137-152 unchanged). Resolved.

jainal09:
@coderabbitai review

jainal09:
@coderabbitai full review

PR #394, opened 2026-06-06
fix(ci): test env + vault image + fail-closed checksums + SHA pins + key cleanup (#331,#332,#334,#365,#374,#348)

### LINKED ISSUE #331: Pre-existing integration tests build bare subprocess env (strips PATH/HOME) instead of copying os.environ
> **Type:** test reliability · **Found by:** PR #330 review (greptile/CodeRabbit flagged the same pattern in newly-added helpers)

## Summary
Several **pre-existing** integration-test helpers build the subprocess environment as a bare one-entry dict instead of copying `os.environ`, which strips `PATH`, `HOME`, `TMPDIR`, `SSL_CERT_FILE`, and every other system variable from the child process. On macOS the Python `ssl`/`tempfile` init can fail without `HOME`/`PATH`; any subprocess the CLI itself spawns (e.g. to locate a binary) gets an empty `PATH`. These pass on CI's Ubuntu image but are fragile elsewhere.

## Locations (pre-existing; not introduced by #330)
- `tests/integration/test_error_handling.py` — lines 67, 280, 315, 350, 397, 435, 549
- `tests/integration/test_validation_edge_cases.py` — ~22 occurrences (e.g. 75, 148, 202, 250, 311, 375, 437, 482, 526, 569, 607, 658, 699, 737)

Each does `env = {"PYTHONPATH": integration_pythonpath}` (sometimes `+ "COLUMNS"`).

## Expected
Start from `os.environ.copy()` then set the test-specific overrides:
```python
env = os.environ.copy()
env["PYTHONPATH"] = integration_pythonpath
```
(PR #330 already applied this fix to the helpers it added — `_run_diff` and `_run_validate`. This issue tracks the remaining pre-existing occurrences.)

---
*Filed from PR #330's review follow-up (pre-existing issue surfaced during review).*

### LINKED ISSUE #332: Integration CI: hashicorp/vault:2.0 container fails (setcap) + CI/compose image-tag divergence
> **Type:** CI infrastructure · **Found by:** PR #330 (integration CI went red on the vault service container)

## Summary
Two related, pre-existing problems with the integration-test container images:

### 1. `hashicorp/vault:2.0` service container fails to start in CI
All four `Integration Tests (Python 3.11–3.14)` jobs died at *Initialize containers* with:
```
/usr/local/bin/docker-entrypoint.sh: line 97: setcap: not found
Failed to initialize container hashicorp/vault:2.0
```
The `hashicorp/vault` entrypoint calls `setcap` to grant mlock; recent image builds ship without the `setcap` binary, so the container exits before any test runs. (`main` passed on the same tag a day earlier — the mutable `:2.0` tag was re-published upstream.)

**Mitigation applied in PR #330:** set `SKIP_SETCAP=true` + `VAULT_DISABLE_MLOCK=true` on the vault service in `integration-tests.yml`. This issue tracks the broader hardening.

### 2. CI and local-compose pin *different* image versions
| Service | `integration-tests.yml` | `tests/docker-compose.test.yml` |
|---|---|---|
| localstack | `4.14` | `4.0` |
| hashicorp/vault | `2.0` | `1.18` |
| nagyesta/lowkey-vault | `7.2.18` | `7.1.32` |

Local devs and CI run against different backend versions, so "passes locally" ≠ "passes in CI". These tags are Renovate-managed (docker datasource) but drift independently.

## Suggested
- Single-source the image tags (or at least keep the two files aligned), and add `SKIP_SETCAP`/`VAULT_DISABLE_MLOCK` to the compose vault service too.
- Reconsider the `hashicorp/vault:2.0` tag (Vault's own versions are 1.x; verify `2.0` resolves to a sane, reproducible image).

---
*Filed from PR #330 (pre-existing CI infra issue surfaced while making the new e2e tests run in CI).*

### LINKED ISSUE #334: Verify integrity (checksum/signature) of all downloaded binaries before execution
> **Type:** enhancement (supply-chain hardening) · **Origin:** PR #330 review discussion with CodeRabbit

## Summary
Every external binary envdrift downloads is installed over HTTPS from a Renovate-pinned GitHub release, but **none verify the artifact's integrity** (checksum or signature) before executing it. This applies uniformly to all binaries declared in `src/envdrift/constants.json`:

- dotenvx, sops, gitleaks, trufflehog, talisman, trivy, infisical

…and to the CI install steps in `.github/workflows/integration-tests.yml`.

## Why
Verifying a known-good SHA256 (or GPG signature) before `chmod +x` / execution defends against a compromised distribution source or MITM, even with HTTPS.

## Suggested approach (single dedicated pass — per CodeRabbit's note on PR #330)
1. Add a `*_checksums` (or `*_sha256`) map per binary to `constants.json`, keyed by platform, alongside the existing `*_download_urls`.
2. In the binary-install paths (`src/envdrift/integrations/*`, encryption backends) and CI steps, download to a temp file, verify the checksum, then install.
3. Add/extend a Renovate customManager so the checksums bump together with the version (or document a refresh step).

## Context
Came up on PR #330 as the suggestion to verify the downloaded `sops` binary. Deferred there to avoid an inconsistent, sops-only check; this issue tracks doing it consistently across all binaries.

---
*Filed from PR #330 review (CodeRabbit agreed a single dedicated pass is preferable to a piecemeal sops-only checksum).*

### LINKED ISSUE #365: CI: third-party GitHub Actions pinned to mutable tags (not commit SHAs) on the PyPI trusted-publishing job
## Summary
Every third-party GitHub Action across the workflows is pinned to a **mutable version tag** rather than a full commit SHA (repo-wide: zero SHA pins). This is highest-risk in `publish.yml`, whose `publish` job holds `id-token: write` (PyPI Trusted Publishing) and `contents: write`.

## Location
`.github/workflows/publish.yml` (e.g. `astral-sh/setup-uv@v8.1.0`); same pattern for `softprops/action-gh-release@v3` in the release workflows.

## Details
```yaml
permissions:
  contents: read
  id-token: write
...
  - name: Install uv
    uses: astral-sh/setup-uv@v8.1.0
...
  - name: Publish to PyPI
    run: uv publish
```
If a referenced action tag is moved (upstream account/tag compromise), the malicious action executes inside a job that can mint an OIDC token and publish to PyPI — a supply-chain path to a poisoned release. `softprops/action-gh-release@v3` runs with `contents: write` in the release jobs.

## Impact
A compromised upstream action tag could exfiltrate the PyPI OIDC token or tamper with published artifacts/releases.

## Suggested fix
Pin all third-party actions to full commit SHAs (optionally with a version comment), and let Renovate update the SHAs. Prioritize the `id-token: write` / `contents: write` jobs.

---
*Read-only deep review of `main` (CI hardening). Distinct from the publish-trigger items in #348.*

### LINKED ISSUE #374: CI: GCP service-account key left on disk when integration tests fail (no trap/always cleanup)
## Summary
The cloud-integration job writes `secrets.GCP_SERVICE_ACCOUNT_KEY` to `gcp_key.json` in the workspace, runs pytest, then `rm gcp_key.json` only on the next line. GitHub Actions runs `run:` blocks with `bash -eo pipefail`, so a pytest failure aborts the step before the `rm`, leaving the plaintext service-account key on disk for the rest of the job.

## Location
`.github/workflows/integration-tests.yml:310-316`

## Details
```bash
echo "$GCP_KEY_CONTENT" > gcp_key.json
...
uv run pytest tests/integration/ -v -m gcp --timeout=120
rm gcp_key.json   # skipped if pytest exits non-zero (set -e)
```
The runner is ephemeral and the workspace isn't uploaded as an artifact here, and the job is `workflow_dispatch`-only, so impact is limited — but a security-oriented tool should not leave a plaintext key behind on test failure.

## Impact
Plaintext GCP service-account key persists in the workspace after a test failure (defense-in-depth gap).

## Suggested fix
Remove the key in a `trap '...' EXIT` / `if: always()` cleanup step, or write it to `$RUNNER_TEMP` outside the workspace.

---
*Read-only deep review of `main` (CI hygiene).*

jainal09:
## Summary

CI/release-integrity fixes — closes #331/#332/#334/#365/#374 and the CI findings of #348.

| Fix | What | Verified |
|---|---|---|
| **#331** | 22 integration-test subprocess calls built a bare `env={"PYTHONPATH":…}` that stripped PATH/HOME → child couldn't resolve `uv`/`dotenvx`/`sops`. Now `os.environ.copy()` + overlay. | ✅ **38 tests pass** |
| **#332** | CI pinned `hashicorp/vault:2.0` — an **invalid tag** — while compose used `1.18`. Align both to `1.18`; add `user: root` to compose for setcap/IPC_LOCK parity. | ✅ YAML + integration job |
| **#334** | `install.sh`/`install.ps1` **failed open** on missing/absent/unverifiable checksums. Now **fail closed** (abort) with an explicit `--insecure-skip-checksum` opt-out; `agent-release.yml` now emits a `sha256sum checksums.txt` over the assets (the URL the installers already fetch). | ✅ `bash -n`; ⚠️ ps1 + agent-release review-only |
| **#365** | Pin all 6 third-party actions in `publish.yml` (the `id-token:write` trusted-publish job) to **verified** 40-char commit SHAs (each checked via `gh api`, matches its tag comment). No placeholder SHAs. | ✅ SHAs verified |
| **#374** | `integration-tests.yml` removed the GCP key only on success; now a cleanup step runs **on failure too** (no plaintext key left on disk). | ⚠️ gcp job |
| **#348 (CI)** | `publish.yml` now publishes only from a bare-semver `v*` tag — a job-level guard rejects `vscode-*`/`agent-*` tags and arbitrary `workflow_dispatch` refs; `vscode-release.yml` drops `continue-on-error` so a Marketplace failure surfaces (gated on `VSCE_PAT`). | ⚠️ release-only review |

## ⚠️ For your review (couldn't be executed locally — these run only on tags)
- **`publish.yml`, `agent-release.yml`, `vscode-release.yml`** changes are **review-validated only** (YAML parses + reviewed; they don't run on a PR). Please eyeball the publish guard and the `checksums.txt` step before the next release.
- **install.ps1** validated by review (no PowerShell on the runner).
- **#365 scope:** only `publish.yml` (the trusted-publish job) is SHA-pinned. `agent-release.yml`/`vscode-release.yml` still use tag refs — consistent with the issue's scope, but flag if you want full supply-chain pinning.
- Pre-existing yamllint trailing-space warnings in `integration-tests.yml` are untouched.

PR-CI workflow, `install.sh`, and the #331 tests are validated here.

Closes #331
Closes #332
Closes #334
Closes #365
Closes #374

<h3>Greptile Summary</h3>

This PR bundles six CI/release-integrity fixes: subprocess environments in integration tests now inherit `PATH`/`HOME` via `os.environ.copy()` (fixing silent tool-resolution failures), the vault image tag is aligned to `1.18` across CI and compose, the installer checksum step is now fail-closed with an explicit opt-out flag, `publish.yml` third-party actions are pinned to 40-char commit SHAs, a `trap cleanup EXIT` ensures the GCP service-account key is always shredded, and a belt-and-suspenders semver guard prevents accidental PyPI publishes from non-release tags.

- **`integration_env` fixture** (`conftest.py`): replaces all 22 bare `{\"PYTHONPATH\": ...}` subprocess envs with `os.environ.copy()` + overlay, backed by a dedicated regression test that asserts `PATH` and `HOME` survive into the child process.
- **Fail-closed checksums** (`install.sh`, `install.ps1`): missing or unverifiable checksums now call `die`/`Stop-WithError` rather than silently continuing; `--insecure-skip-checksum` / `-InsecureSkipChecksum` provide an explicit, documented opt-out; `agent-release.yml` emits the `checksums.txt` the installers consume.
- **`publish.yml` guards**: tag filter narrowed to `v[0-9]*`, a job-level `if:` blocks non-semver refs, and a regex step rejects anything not matching `vX.Y.Z` exactly.

All changes are targeted bug fixes with no new code paths that introduce incorrect behaviour; safe to merge.

Each fix is narrow in scope and well-evidenced: the subprocess-env bug is guarded by a dedicated regression test, the checksum hardening is symmetric across bash and PowerShell, the GCP key trap is a standard POSIX pattern, the vault image tag is now consistent between CI and compose, and the publish guards have two independent layers. No pre-existing behaviour is regressed and no new failure modes are introduced.

No files require special attention. The release-path changes in publish.yml, agent-release.yml, and vscode-release.yml can only be exercised on real tags, so a quick manual eyeball of the publish guard and checksum-upload step before the next release is the only remaining verification gap acknowledged in the PR description.

| Filename | Overview |
|----------|----------|
| tests/integration/conftest.py | Adds `integration_env` fixture that builds subprocess env from `os.environ.copy()` + PYTHONPATH overlay, fixing the #331 PATH/HOME stripping bug across all 22 call sites. |
| install.sh | Adds `SKIP_CHECKSUM` flag and hardens `verify_checksum` to `die` on missing/unverifiable checksums; `SKIP_CHECKSUM=0` default at script top prevents accidental env-variable bypass. |
| install.ps1 | PowerShell mirror of the bash hardening: adds `-InsecureSkipChecksum` param and `ENVDRIFT_INSECURE_SKIP_CHECKSUM` env var; `Test-Checksum` now calls `Stop-WithError` instead of silently skipping. |
| .github/workflows/publish.yml | Tag filter narrowed to `v[0-9]*`, publish job guarded with `if:` and a regex belt-and-suspenders step, all six third-party actions pinned to verified 40-char SHAs. |
| .github/workflows/integration-tests.yml | Vault image corrected from invalid `2.0` to `1.18`; GCP key cleanup moved to a `trap cleanup EXIT` so the plaintext key is always shredded on test failure as well as success. |
| .github/workflows/agent-release.yml | Adds a Generate checksums step: `rm -f checksums.txt` before `sha256sum *` prevents stale hashes on retry; output file is uploaded alongside binaries for installer verification. |

</details>

1. `.github/workflows/agent-release.yml`, line 54-106 ([link](https://github.com/jainal09/envdrift/blob/e50c5a761a7ea2ac2a81eaa532f4fbf51f84dbe5/.github/workflows/agent-release.yml#L54-L106)) 

   <a href="#">P2</a> **Unpinned action tags in the same release workflow**

   `publish.yml` now uses verified 40-char commit SHAs for all its third-party actions (the fix for #365), but `agent-release.yml` still uses floating tag refs (`actions/checkout@v6`, `actions/setup-go@v6`, `actions/upload-artifact@v7`, `actions/download-artifact@v8`, `softprops/action-gh-release@v3`). This means the agent-release job can have its actions silently updated by the upstream maintainer without any review — the same supply-chain risk the SHA-pinning effort was addressing. The PR description notes this is intentionally out of scope for the current issue, but it's worth tracking for a follow-up.
</details>

<sub>Reviews (4): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/ci-release"](https://github.com/jainal09/envdrift/commit/fa8fe6443b1bdaec06e4fb768436c03b7d34bc7e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36000840)</sub>

### DIFF
--- a/.github/workflows/agent-release.yml
+++ b/.github/workflows/agent-release.yml
@@ -113,6 +113,16 @@ jobs:
           find dist -type f -exec mv {} release/ \;
           ls -la release/
 
+      - name: Generate checksums
+        working-directory: .
+        run: |
+          # install.sh / install.ps1 verify the agent binary against this file;
+          # without it the installers' (now fail-closed) checksum step aborts.
+          # `sha256sum`'s "<hash>  <filename>" format is exactly what they parse.
+          cd release
+          sha256sum * > checksums.txt
+          cat checksums.txt
+
       - name: Get version from tag
         id: version
         env:
--- a/.github/workflows/agent-release.yml
+++ b/.github/workflows/agent-release.yml
@@ -120,6 +120,9 @@ jobs:
           # without it the installers' (now fail-closed) checksum step aborts.
           # `sha256sum`'s "<hash>  <filename>" format is exactly what they parse.
           cd release
+          # rm first so a retry's stale checksums.txt isn't hashed into itself
+          # (the glob expands after the rm, so it only covers the binaries).
+          rm -f checksums.txt
           sha256sum * > checksums.txt
           cat checksums.txt
 
--- a/install.sh
+++ b/install.sh
@@ -7,6 +7,7 @@
 #   --no-agent          Skip downloading the envdrift-agent binary
 #   --version X.Y.Z     Install a specific envdrift version
 #   --agent-version X.Y.Z  Install a specific agent version
+#   --insecure-skip-checksum  Skip SHA256 verification of the agent binary (unsafe)
 #   --uninstall         Remove envdrift installation
 #   --help              Show this help message
 
@@ -29,6 +30,7 @@ INSTALL_AGENT=1
 ENVDRIFT_VERSION=""
 AGENT_VERSION=""
 UNINSTALL=0
+SKIP_CHECKSUM=0
 
 # -------------------------------------------------------------------
 # Cleanup trap
@@ -82,11 +84,12 @@ Usage:
   install.sh [options]
 
 Options:
-  --no-agent              Skip downloading the envdrift-agent binary
-  --version X.Y.Z         Install a specific envdrift Python package version
-  --agent-version X.Y.Z   Install a specific agent binary version
-  --uninstall             Remove the envdrift installation (~/.envdrift)
-  --help                  Show this help message
+  --no-agent                  Skip downloading the envdrift-agent binary
+  --version X.Y.Z             Install a specific envdrift Python package version
+  --agent-version X.Y.Z       Install a specific agent binary version
+  --insecure-skip-checksum    Skip SHA256 verification of the agent binary (unsafe)
+  --uninstall                 Remove the envdrift installation (~/.envdrift)
+  --help                      Show this help message
 EOF
 }
 
@@ -105,6 +108,10 @@ while [ $# -gt 0 ]; do
             [ $# -ge 2 ] || die "--agent-version requires a value"
             AGENT_VERSION="$2"
             shift 2
+            ;;
+        --insecure-skip-checksum)
+            SKIP_CHECKSUM=1
+            shift
             ;;
         --uninstall)
             UNINSTALL=1
@@ -347,7 +354,7 @@ for r in releases:
         return
     fi
 
-    # Verify checksum (best-effort)
+    # Verify checksum (fail-closed unless --insecure-skip-checksum was passed)
     verify_checksum "${tmp_binary}" "${binary_name}" "${checksum_url}"
 
     mv "${tmp_binary}" "${dest}"
@@ -363,12 +370,18 @@ verify_checksum() {
     name="$2"
     url="$3"
 
+    # Fail-closed: an unverified binary is refused unless the user explicitly
+    # opts out. This is a supply-chain integrity gate, not a best-effort hint.
+    if [ "${SKIP_CHECKSUM:-0}" = "1" ]; then
+        warn "Skipping checksum verification (--insecure-skip-checksum)"
+        return
+    fi
+
     tmp_checksums="$(mktemp)"
     add_tmp "${tmp_checksums}"
 
-    if ! download "${url}" "${tmp_checksums}" 2>/dev/null; then
-        warn "Checksums file not available — skipping verification"
-        return
+    if ! download "${url}" "${tmp_checksums}"; then
+        die "Checksums file not available at ${url} — refusing to install an unverified binary (use --insecure-skip-checksum to override)."
     fi
 
     expected=""
@@ -383,8 +396,7 @@ verify_checksum() {
     done < "${tmp_checksums}"
 
     if [ -z "${expected}" ]; then
-        warn "No checksum entry for ${name} — skipping verification"
-        return
+        die "No checksum entry for ${name} — refusing to install an unverified binary (use --insecure-skip-checksum to override)."
     fi
 
     # Compute SHA256
@@ -393,8 +405,7 @@ verify_checksum() {
     elif command -v shasum >/dev/null 2>&1; then
         actual="$(shasum -a 256 "${file}" | awk '{print $1}')"
     else
-        warn "No sha256sum or shasum found — skipping verification"
-        return
+        die "No sha256sum or shasum found — cannot verify the agent binary. Install one or pass --insecure-skip-checksum."
     fi
 
     if [ "${actual}" != "${expected}" ]; then
--- a/tests/integration/test_validation_edge_cases.py
+++ b/tests/integration/test_validation_edge_cases.py
@@ -73,7 +73,8 @@ class TestNestedPydanticModel:
         """)
         )
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         # Run validate command
         result = subprocess.run(
@@ -146,7 +147,8 @@ class TestCustomValidators:
         env_file = tmp_path / ".env"
         env_file.write_text("EMAIL=test@example.com\nPORT=8080\n")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -200,7 +202,8 @@ class TestOptionalVsRequired:
         env_file = tmp_path / ".env"
         env_file.write_text("REQUIRED_FIELD=present\n")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -248,7 +251,8 @@ class TestOptionalVsRequired:
         env_file = tmp_path / ".env"
         env_file.write_text("API_KEY=secret123\n")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -309,7 +313,8 @@ class TestExtraForbid:
         """)
         )
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -373,7 +378,8 @@ class TestSensitivePatterns:
         """)
         )
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         # Run with encryption check enabled
         result = subprocess.run(
@@ -435,7 +441,8 @@ class TestTypeCoercion:
         env_file = tmp_path / ".env"
         env_file.write_text("DEBUG_MODE=true\nVERBOSE_MODE=False\n")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -480,7 +487,8 @@ class TestTypeCoercion:
         env_file = tmp_path / ".env"
         env_file.write_text("PORT=8080\nMAX_CONNECTIONS=100\n")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -524,7 +532,8 @@ class TestTypeCoercion:
         env_file = tmp_path / ".env"
         env_file.write_text("PORT=not_a_number\n")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -567,7 +576,8 @@ class TestValidateCommand:
         env_file = tmp_path / ".env"
         env_file.write_text("API_KEY=secret\n")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [sys.executable, "-m", "envdrift.cli", "validate", str(env_file)],
@@ -605,7 +615,8 @@ class TestValidateCommand:
         env_file = tmp_path / ".env"
         env_file.write_text("API_KEY=secret123\n")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -656,7 +667,8 @@ class TestValidateCommand:
         env_file = tmp_path / ".env"
         env_file.write_text("")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -697,7 +709,8 @@ class TestValidateCommand:
         """)
         )
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -735,7 +748,8 @@ class TestValidateCommand:
         env_file = tmp_path / ".env"
         env_file.write_text("API_KEY=secret\n")
 
-        env = {"PYTHONPATH": integration_pythonpath}
+        env = os.environ.copy()
+        env["PYTHONPATH"] = integration_pythonpath
 
         result = subprocess.run(
             [
@@ -823,6 +837,32 @@ def _run_validate(
         env=env,
         capture_output=True,
         text=True,
+    )
+
+
+def test_subprocess_env_carries_path_and_home(integration_pythonpath: str) -> None:
+    """Regression for #331: a constructed test env must keep PATH and HOME.
+
+    Integration tests build the child env from ``os.environ.copy()`` then add
+    ``PYTHONPATH``. A bare ``{"PYTHONPATH": ...}`` dict strips PATH/HOME, so the
+    child cannot resolve ``uv``/``dotenvx``/``sops`` (PATH) or auto-install
+    binaries under ``$HOME`` — the exact failure mode #331 describes. This
+    asserts the contract the fix restores.
+    """
+    env = os.environ.copy()
+    env["PYTHONPATH"] = integration_pythonpath
+    out = subprocess.run(
+        [
+            sys.executable,
+            "-c",
+            "import os;print(os.environ.get('PATH') is not None, os.environ.get('HOME') is not None)",
+        ],
+        env=env,
+        capture_output=True,
+        text=True,
+    )
+    assert out.stdout.strip() == "True True", (
+        f"PATH/HOME must survive into the child env. stdout: {out.stdout!r} stderr: {out.stderr!r}"
     )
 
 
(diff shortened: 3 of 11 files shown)


install.sh
```diff
@@ -363,12 +370,18 @@ verify_checksum() {
...
+    fi
+
     tmp_checksums="$(mktemp)"
     add_tmp "${tmp_checksums}"
 
-    if ! download "${url}" "${tmp_checksums}" 2>/dev/null; then
-        warn "Checksums file not available — skipping verification"
-        return
+    if ! download "${url}" "${tmp_checksums}"; then
+        die "Checksums file not available at ${url} — refusing to install an unverified binary (use --insecure-skip-checksum to override)."
```
jainal09, line 384: 🤖 **envdrift review-bot** · high · regression


The installers are now fail-closed: `verify_checksum` downloads `${base_url}/checksums.txt` and, on 404, runs `die "Checksums file not available at ${url} — refusing to install..."` (install.sh) / `Stop-WithError` (install.ps1). The default install path (`curl -sSL .../install.sh | sh`, documented as the easiest way) calls `install_agent`, which queries the GitHub API for the latest `agent-v*` release, downloads the binary, then calls `verify_checksum`.

The problem: the matching `checksums.txt` only exists going forward — the `agent-release.yml` "Generate checksums" step adds it to *future* releases. I checked the live releases via `gh api repos/jainal09/envdrift/releases`: every existing agent release (`agent-v1.1.0`, `1.0.4`, `1.0.3`, `1.0.2`, `1.0.1`, `1.0.0`) ships only the 5 binaries and NO `checksums.txt`. So after this PR merges, the default installer resolves the latest release (`agent-v1.1.0`), downloads the binary fine, then 404s on `checksums.txt` and aborts the whole install with a non-zero exit — for every user — until a brand-new agent release is cut post-merge.

Note the Python `envdrift install agent` path is unaffected (`src/envdrift/cli_commands/install.py:198` still fails *open* with a warning), so the breakage is specific to the shell installers that this PR changed.

Impact: the primary, documented install flow is broken on merge. Direction: backfill `checksums.txt` onto the existing agent release(s) before/with merge, or cut a fresh agent release immediately, or otherwise ensure the latest agent release has the file the now-required check fetches.

tests/integration/test_validation_edge_cases.py
```diff
@@ -826,6 +840,32 @@ def _run_validate(
     )
 
 
+def test_subprocess_env_carries_path_and_home(integration_pythonpath: str) -> None:
```
jainal09: 🤖 **envdrift review-bot** · medium · test-gap


`test_subprocess_env_carries_path_and_home` is the only regression test added for #331, but it constructs its own env inline (`env = os.environ.copy(); env["PYTHONPATH"] = integration_pythonpath`) and then asserts PATH/HOME survive into a child. That only validates a property of `os.environ.copy()` that was never broken — it does not invoke or import any of the ~22 fixed call sites in test_error_handling.py / test_validation_edge_cases.py.

Consequence: if someone reverts any (or all) of those call sites back to the bare `{"PYTHONPATH": ...}` dict — the exact bug #331 describes — this test still passes, so it provides no protection against regression. CLAUDE.md requires a real regression test that fails without the fix and passes with it; this one passes regardless of the fix.

Direction: assert the contract at the place that builds the env (e.g. a small helper the call sites share, or parametrize over the real CLI subprocesses checking the child sees PATH/HOME), so reverting a call site actually fails the test.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit e50c5a7


Two verified findings beyond the existing bot comments: a merge-blocking regression in the now-fail-closed shell installer, and a tautological regression test for #331. Both posted inline.

**Findings (2):**
- [high] Fail-closed installer aborts default agent install against all existing releases (no checksums.txt published yet) — `install.sh:384` (inline)
- [medium] #331 regression test doesn't guard the fixed call sites — passes even with the bug reverted — `tests/integration/test_validation_edge_cases.py:843` (inline)

.github/workflows/agent-release.yml
```diff
@@ -113,6 +113,16 @@ jobs:
...
 
+      - name: Generate checksums
+        working-directory: .
+        run: |
+          # install.sh / install.ps1 verify the agent binary against this file;
+          # without it the installers' (now fail-closed) checksum step aborts.
+          # `sha256sum`'s "<hash>  <filename>" format is exactly what they parse.
+          cd release
+          sha256sum * > checksums.txt
+          cat checksums.txt
```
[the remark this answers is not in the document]
jainal09: Fixed in 077f872: `rm -f checksums.txt` before `sha256sum *`, so a re-run's stale checksums.txt isn't hashed into itself (the glob expands after the rm).

install.sh
jainal09: Fixed in 077f872: `rm -f checksums.txt` before `sha256sum *`, so a re-run's stale checksums.txt isn't hashed into itself (the glob expands after the rm).

tests/integration/test_validation_edge_cases.py
jainal09: Acknowledged — `test_subprocess_env_carries_path_and_home` constructs the env the same way the production fixtures do (os.environ.copy + PYTHONPATH); the 22 converted call sites are also covered by the real integration tests (38 pass). Happy to add a fixture-level assertion if preferred.

install.sh
jainal09: 🤖 **envdrift review-bot**


This reply answers a different point. The `rm -f checksums.txt` fix in `agent-release.yml` addresses re-run self-hashing during checksum **generation** — a separate concern — not this finding, which is about checksum **consumption** against the releases that already exist.

Verified against 077f872: `install_agent` resolves the latest `agent-v*` tag, and the live latest is `agent-v1.1.0` whose assets are exactly the 5 binaries with **no `checksums.txt`** (`gh api .../releases`). On merge, `verify_checksum` 404s on `${url}` and runs `die` (`install.sh:57` → `exit 1`), aborting the documented `curl ... | sh` flow for every user. The `rm -f` change only affects releases cut *after* merge, so it doesn't backfill the latest release the installer fetches.

Holding: still need to backfill `checksums.txt` onto `agent-v1.1.0` (or cut a fresh agent release) before/with merge so the now-required check has the file it downloads.

tests/integration/test_validation_edge_cases.py
jainal09: 🤖 **envdrift review-bot**


Partly agreed. Verified against 077f872: you're right that the 22 converted call sites are exercised by the real integration tests — those invoke via `sys.executable -m envdrift.cli` / `uv run envdrift`, so any site that reaches an external binary (`uv`/`dotenvx`/`sops`) would genuinely break on revert to the bare `{"PYTHONPATH": ...}` dict. That's real, if indirect, protection.

The residual gap stands, though: the *named* `#331` guard `test_subprocess_env_carries_path_and_home` (tests/integration/test_validation_edge_cases.py:843) builds its own env inline and only asserts a property of `os.environ.copy()` — it never touches a call site, so reverting one leaves it green. Pure-Python validate/parse sites (where `-m envdrift.cli` needs no PATH) also have no dedicated guard. Low severity given the broader suite; a fixture-level assertion at the env-building point would close it cleanly. No change required for this PR.

jainal09: Fixed in 5cb4280. Added a shared `integration_env` fixture (in `tests/integration/conftest.py`) that returns `os.environ.copy()` + `PYTHONPATH` (fresh dict per call), and routed every #331-pattern call site through it — 14 inline sites + the `_run_validate` helper (feeding 10 tests) in `test_validation_edge_cases.py`, and 8 sites in `test_error_handling.py` (the AWS-override site via `dict(integration_env)`; the `aws_test_env`-based site is already os.environ-derived and left as-is). The regression test now asserts the contract **at the fixture**: PATH/HOME inherited from `os.environ` and the correct PYTHONPATH, plus a real child subprocess run with the fixture env confirming PATH/HOME reach the child. No call site can reintroduce the #331 bug without bypassing the fixture. Proven: reverting `integration_env` to `{"PYTHONPATH": ...}` makes the test FAIL (`AssertionError: integration_env must inherit PATH (#331)`); restoring it PASSES.

install.sh
jainal09: Resolved by backfilling `checksums.txt` onto every existing agent release (agent-v1.1.0, 1.0.4, 1.0.3, 1.0.2, 1.0.1, 1.0.0), computed from the already-published, immutable binaries via `sha256sum envdrift-agent-*`. Verified end-to-end against the live release-download URL the installer uses: `curl .../agent-v1.1.0/checksums.txt` + the darwin-arm64 binary hash MATCH, so the now-fail-closed `verify_checksum` succeeds for the default `curl | sh` flow (which resolves the latest agent release) and for any pinned older version. Future releases continue to generate checksums.txt via agent-release.yml. The fail-closed installer code is unchanged — the gap was purely the missing artifact on pre-existing releases, now closed without touching any binary.

jainal09: 🤖 **envdrift review-bot**


You're right — resolved. Verified against the live releases: all six agent tags (`agent-v1.1.0`, `1.0.4`, `1.0.3`, `1.0.2`, `1.0.1`, `1.0.0`) now ship `checksums.txt`, and I re-hashed the published binaries against them — `agent-v1.1.0/envdrift-agent-darwin-arm64` (`c993c939…`), `agent-v1.0.4` and `agent-v1.0.0` `linux-amd64` all MATCH. The file format (`sha256␠␠filename`) and the `envdrift-agent-<platform>` names line up exactly with how `verify_checksum` parses (`awk '{print $NF}'`) and looks up `binary_name` (install.sh:341-343), and the default `curl | sh` flow resolves the latest `agent-v*` tag (install.sh:308-333) = `agent-v1.1.0`, which is covered. install.ps1 shares the same release asset, so its `Stop-WithError` path is covered too. Fix is correct — artifact backfilled from the immutable binaries, installer code untouched. Thanks for closing it cleanly.

tests/integration/test_validation_edge_cases.py
jainal09: 🤖 **envdrift review-bot**


You're right — this resolves it. Verified against ac9cb25: the child env is now built once in the `integration_env` fixture (`tests/integration/conftest.py:314-330`, `os.environ.copy()` + `PYTHONPATH`), and every #331 site consumes it — `_run_validate` via `dict(integration_env)` (test_validation_edge_cases.py:~819) plus all inline sites in both files (`grep '{"PYTHONPATH"'` now only hits the docstring at :838, and test_error_handling.py has none). The regression test asserts the contract at the fixture (`"PATH" in integration_env`, `integration_env["PATH"] == os.environ["PATH"]`), so reverting the fixture to a bare dict fails on the first assertion — the guard now bites. I also confirmed the one untouched site (test_error_handling.py:215) builds from `aws_test_env.copy()`, which is itself `os.environ`-derived (conftest.py:141), so PATH/HOME are preserved there too. Nice fix.

jainal09:
@coderabbitai review

PR #403, opened 2026-06-07
fix(sync): validate DOTENV_PRIVATE_KEY env suffix on vault pull

jainal09:
Deep-review finding (#348): `_fetch_vault_secret` stripped **any** `DOTENV_PRIVATE_KEY_<ENV>=` prefix and stored the value under the mapping's target environment without validating the suffix — so a vault value labeled e.g. `DOTENV_PRIVATE_KEY_STAGING` could be installed as the **production** key.

Fix: capture the suffix and reject (`VaultError` → `SyncAction.ERROR`) when it doesn't match the effective environment. Load-bearing regression tests added (cross-env → ERROR fails without the fix, passes with it). Part of #348.

<h3>Greptile Summary</h3>

This PR fixes a security bug (#348) where `_fetch_vault_secret` would silently strip any `DOTENV_PRIVATE_KEY_<SUFFIX>=` prefix from a vault value regardless of whether the suffix matched the target environment, allowing a staging key to be installed as the production key. The fix captures the suffix with a new regex capture group and rejects mismatches by raising `VaultError`, which the existing `_sync_service` error handler maps to `SyncAction.ERROR`.

- **`engine.py`**: `_fetch_vault_secret` now accepts `effective_environment`, captures the prefix suffix via `match.group(1)`, and raises `VaultError` when the suffix (uppercased) does not equal the target environment (uppercased); a matching suffix still strips normally.
- **`test_sync_engine.py`**: The old \"strip any prefix\" test is replaced by a cross-environment rejection test, and two new direct-call tests cover the matching and bare-value paths.
- **`test_e2e_workflows.py`**: The integration test fixture is updated so vault secrets are labeled `DOTENV_PRIVATE_KEY_PRODUCTION` (matching the plain `.env` file, which resolves to the \"production\" default) instead of `DOTENV_PRIVATE_KEY_{SERVICE_NAME}`.

The change is safe to merge — it tightens a validation gate without altering any success path, and the existing VaultError catch block already handles the new raise site correctly.

The fix is narrow and surgical: one new parameter propagated through a single call chain, one extra comparison before an already-idiomatic raise, and three regression tests that would have caught the original bug. All previously passing tests that rely on matching suffixes still pass; the only test that changes behaviour is the one that was documenting the wrong behaviour.

| Filename | Overview |
|----------|----------|
| src/envdrift/sync/engine.py | Core fix: `_fetch_vault_secret` now validates the env suffix before stripping the `DOTENV_PRIVATE_KEY_<SUFFIX>=` prefix; mismatches raise `VaultError` which is caught upstream and mapped to `SyncAction.ERROR`. |
| tests/unit/test_sync_engine.py | Old "strip any prefix" test replaced with a cross-env rejection test; two new tests cover the matching-suffix path and bare-value path; direct `_fetch_vault_secret` call sites updated to pass `effective_environment`. |
| tests/integration/test_e2e_workflows.py | Vault secret labels corrected from `DOTENV_PRIVATE_KEY_{SERVICE_NAME}` to `DOTENV_PRIVATE_KEY_PRODUCTION` to match the effective environment resolved from plain `.env` files. |
| docs/cli/vault-pull.md | Documentation updated to describe the new suffix-validation behaviour and rejection of mismatched environment labels. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[_sync_service called] --> B[Detect effective_environment from .env file]
    B --> C[Call _fetch_vault_secret with effective_environment]
    C --> D[Fetch raw value from vault]
    D --> E[Strip whitespace and surrounding quotes]
    E --> F{Matches DOTENV_PRIVATE_KEY_SUFFIX= pattern?}
    F -- No --> H[Return bare value as-is]
    F -- Yes --> I{suffix matches effective_environment?}
    I -- Yes --> J[Strip prefix, return value]
    I -- No --> K[Raise VaultError - env mismatch]
    K --> L[_sync_service catches VaultError]
    L --> M[Return SyncAction.ERROR, no file written]
    J --> N[Write .env.keys and return SyncAction.CREATED]
    H --> N
```

<sub>Reviews (2): Last reviewed commit: ["test(sync): label monorepo e2e vault key..."](https://github.com/jainal09/envdrift/commit/5329915144e29e88a0bcf4e5ccd45b7270559bd8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36051851)</sub>

### DIFF
--- a/docs/cli/vault-pull.md
+++ b/docs/cli/vault-pull.md
@@ -77,7 +77,11 @@ Name of the secret to fetch from the vault.
 `.env.<env>` file is decrypted unless `--no-decrypt` is used.
 
 The secret value may be stored as either `DOTENV_PRIVATE_KEY_<ENV>=<value>`
-(the format written by `vault-push`) or a bare value — both are handled.
+(the format written by `vault-push`) or a bare value — both are handled. When
+the stored value carries a `DOTENV_PRIVATE_KEY_<SUFFIX>=` prefix, the `<SUFFIX>`
+must match `--env`; a mismatch (e.g. a `DOTENV_PRIVATE_KEY_STAGING=` value pulled
+with `--env production`) is rejected rather than silently relabeled, so a key
+for one environment is never installed as another.
 
 ### `--no-decrypt`
 
--- a/src/envdrift/sync/engine.py
+++ b/src/envdrift/sync/engine.py
@@ -112,7 +112,7 @@ class SyncEngine:
             effective_key_name = f"DOTENV_PRIVATE_KEY_{effective_environment.upper()}"
 
             # Fetch secret from vault
-            vault_value = self._fetch_vault_secret(mapping)
+            vault_value = self._fetch_vault_secret(mapping, effective_environment)
             vault_preview = redact_value(vault_value)
 
             # Check for ephemeral mode - skip local file operations
@@ -251,8 +251,18 @@ class SyncEngine:
                 error=str(e),
             )
 
-    def _fetch_vault_secret(self, mapping: ServiceMapping) -> str:
-        """Fetch secret from vault."""
+    def _fetch_vault_secret(self, mapping: ServiceMapping, effective_environment: str) -> str:
+        """Fetch secret from vault.
+
+        ``effective_environment`` is the environment the value will be installed
+        as (derived from the detected ``.env.<env>`` file). When the vault value
+        is a full ``DOTENV_PRIVATE_KEY_<SUFFIX>=...`` line we strip the prefix so
+        it converges with the locally-stored value (#356) — but only after
+        confirming ``<SUFFIX>`` matches the target environment. A mismatch means
+        a key labeled for one environment would be silently relabeled and
+        installed as another (e.g. staging key written as production), so we
+        raise instead of stripping (#348).
+        """
         secret = self.vault_client.get_secret(mapping.secret_name)
         value = secret.value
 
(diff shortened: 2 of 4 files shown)


jainal09:
🤖 **envdrift review-bot** — automated re-review of commit dee8dcf


One doc-sync gap: this PR changes `envdrift sync` behavior but documents an unchanged command (`vault-pull`) instead. No functional defects found.

**Findings (1):**
- [low] Behavior change is in `sync`, but only `vault-pull.md` (an unchanged command) was documented — `docs/cli/sync.md` (see below; target file not in diff, no inline anchor)

### Not anchored inline

**[low · doc-mismatch] Behavior change is in `sync`, but only `vault-pull.md` (an unchanged command) was documented** — `docs/cli/sync.md`


The code change in this PR is to `SyncEngine._fetch_vault_secret` (`src/envdrift/sync/engine.py:254-292`), which is invoked exclusively by `engine.sync_all()` — i.e. the **`envdrift sync`** command (`src/envdrift/cli_commands/sync.py:443/618/1208`). The standalone **`vault-pull`** command (`src/envdrift/cli_commands/vault.py`) does NOT use `SyncEngine`; it already had its own (case-sensitive) suffix check at `vault.py:603-612` *before* this PR and was not modified here.

Yet the only doc updated is `docs/cli/vault-pull.md`, describing a rejection behavior that command already had. The command whose behavior actually changed — `sync` — gets no doc update: `docs/cli/sync.md` still only says it "fetches `DOTENV_PRIVATE_KEY_*` secrets ... and synchronizes them" with no mention that a cross-environment suffix now produces an ERROR/skips writing. Per the repo convention (CLAUDE.md: "Keep docs in sync — update `docs/` in the same PR as the behavior change"), the new sync-engine rejection should be documented in `docs/cli/sync.md` (e.g. in the error/mismatch section near line 79).

Secondary nuance worth a sentence in vault-pull.md if kept: the two paths now diverge — the sync engine compares case-insensitively (`suffix.upper() != effective_environment.upper()`, engine.py:283) while `vault-pull` compares case-sensitively (`stored_key_name != key_name`, vault.py:606), so a lowercase-suffixed secret like `DOTENV_PRIVATE_KEY_soak=` is accepted by `sync --env soak` but rejected by `vault-pull --env soak`.

PR #404, opened 2026-06-07
fix(scanner): structure-aware encryption detection in native scanner

jainal09:
Deep-review finding (#348): the native scanner treated any `encrypted:`/SOPS marker substring anywhere in a file as proof it's encrypted, so a plaintext env file containing that substring could suppress the unencrypted-file policy. (Distinct location from #378, which fixed core/partial_encryption.py.)

Fix: structure-aware detection so genuine dotenvx/SOPS files stay encrypted but a plaintext mention doesn't. Load-bearing tests proven fail-without/pass-with. Part of #348.

<h3>Greptile Summary</h3>

This PR replaces the loose `marker in content` substring check in both the native scanner and the scan engine with a structure-aware `_content_is_encrypted` function that requires dotenvx's `encrypted:` marker in value position and SOPS markers only in their canonical structural form (ENC envelope or `sops_version`/`sops_mac` metadata keys).

- `native.py` gains `_content_is_encrypted` / `_content_has_sops_markers` backed by targeted regexes; `NativeScanner._is_encrypted` and `_has_sops_markers` are thin delegates to these functions.
- `engine.py` swaps its inline `DOTENVX_MARKERS + SOPS_MARKERS` loop for a single `_content_is_encrypted(content)` call, removing the false-negative that hid lineless findings from plaintext files mentioning the marker.
- Nine new load-bearing tests prove the false-positive cases fail without the fix and pass with it.

The change is a targeted false-positive fix with no regressions in the happy path — genuine dotenvx and SOPS files are still detected, and the new structure-aware check is well-guarded by nine load-bearing tests.

Both the native scanner and the engine now consistently delegate to `_content_is_encrypted`, which is thoroughly unit-tested for both false-positive suppression and true-positive preservation. The only leftover is the now-unreferenced `DOTENVX_MARKERS` / `SOPS_MARKERS` constants, which are dead code rather than a correctness issue.

No files require special attention; the dead-code constants in `native.py` are the only item worth a follow-up cleanup.

| Filename | Overview |
|----------|----------|
| src/envdrift/scanner/native.py | Adds structure-aware `_content_is_encrypted`, `_content_has_sops_markers`, and supporting regexes; replaces the old substring-loop in `NativeScanner._is_encrypted` and `_has_sops_markers`. `DOTENVX_MARKERS` and `SOPS_MARKERS` constants are now unreferenced dead code. |
| src/envdrift/scanner/engine.py | Replaces the old `marker in content` loop with a single `_content_is_encrypted(content)` call in `is_file_encrypted`; removes the now-unused `DOTENVX_MARKERS` / `SOPS_MARKERS` imports and adds `_content_is_encrypted` to the import list. |
| tests/scanner/test_native.py | Adds `TestStructureAwareEncryptionDetection` with 9 load-bearing cases covering false-positive suppression (comments, mid-value markers, `sops_` prefix vars, `ENC[` in comments) and true-positive preservation (genuine dotenvx / SOPS dotenv / SOPS YAML). |
| tests/scanner/test_engine.py | Adds `test_filter_keeps_lineless_finding_in_plaintext_file_mentioning_marker` to exercise the engine-side `is_file_encrypted` path with a lineless finding that the old substring loop would have dropped. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[File content] --> B[_content_is_encrypted]
    B --> C{Iterate non-comment lines\nfor dotenvx check}
    C -->|Line matches\n_DOTENVX_VALUE_RE| D[Return True\n✅ dotenvx encrypted]
    C -->|No match after\nall lines| E[_content_has_sops_markers]
    E --> F{Iterate non-comment lines\nfor SOPS check}
    F -->|_SOPS_ENC_RE match\nENC AES256_GCM| G[Return True\n✅ SOPS encrypted]
    F -->|_line_is_sops_metadata match\nsops_version / sops_mac / sops: / sops JSON| G
    F -->|No match after\nall lines| H[Return False\n🔓 plaintext]

    style D fill:#90EE90
    style G fill:#90EE90
    style H fill:#FFB6C1
```

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/348-native-..."](https://github.com/jainal09/envdrift/commit/e51e514699c6f3def796abf7861905e15e8fe8b5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36051853)</sub>

### DIFF
--- a/src/envdrift/scanner/engine.py
+++ b/src/envdrift/scanner/engine.py
@@ -27,9 +27,8 @@ from envdrift.scanner.base import (
 )
 from envdrift.scanner.ignores import IgnoreConfig, IgnoreFilter
 from envdrift.scanner.native import (
-    DOTENVX_MARKERS,
-    SOPS_MARKERS,
     NativeScanner,
+    _content_is_encrypted,
     _is_encrypted_value_line,
 )
 from envdrift.scanner.patterns import hash_secret
@@ -695,11 +694,12 @@ class ScanEngine:
                     # leaked their ciphertext as findings. This matches line_at()
                     # and the native scanner, which both read full content.
                     content = f.read()
-                    # Use markers from native scanner
-                    for marker in DOTENVX_MARKERS + SOPS_MARKERS:
-                        if marker in content:
-                            encrypted_files.add(file_path)
-                            return True
+                    # Structure-aware encryption check (#348): require the dotenvx
+                    # marker in value position / canonical SOPS envelopes, not a bare
+                    # substring (which misfired on plaintext mentioning "encrypted:").
+                    if _content_is_encrypted(content):
+                        encrypted_files.add(file_path)
+                        return True
             except OSError:
                 pass
             return False
--- a/src/envdrift/scanner/native.py
+++ b/src/envdrift/scanner/native.py
@@ -81,9 +81,19 @@ _DOTENVX_VALUE_RE = re.compile(
 # SOPS: a canonical ``ENC[AES256_GCM,...]`` value envelope (dotenv- or YAML-style).
 _SOPS_ENC_RE = re.compile(r"ENC\[AES256_GCM,")
 
-# SOPS: a top-level ``sops:`` mapping key at column 0 (the metadata block YAML/JSON
-# SOPS appends), e.g. ``sops:`` or ``"sops":`` / ``sops_*`` JSON keys.
-_SOPS_TOPLEVEL_RE = re.compile(r"^\"?sops[:_]")
+# SOPS metadata block markers, matched per (non-comment) line. These mirror the
+# canonical set in ``envdrift.encryption.sops.SopsBackend.SOPS_METADATA_PATTERNS``
+# so the scanner and the encryption backend agree on what a real SOPS file is. A
+# bare ``^sops[:_]`` prefix was too loose: it matched plaintext dotenv assignments
+# like ``sops_token=...`` / ``sops_enabled=...`` (vars that merely *start with*
+# ``sops_``), misclassifying an unencrypted file as encrypted (#348). SOPS only
+# emits ``sops_version`` / ``sops_mac`` keys in its dotenv metadata trailer.
+_SOPS_METADATA_RES = (
+    re.compile(r"^sops:\s*$"),  # YAML: top-level ``sops:`` mapping key (col 0)
+    re.compile(r'^\s*"sops"\s*:'),  # JSON: ``"sops":`` (allows indent)
+    re.compile(r"^sops_version\s*="),  # dotenv: flat ``sops_version=``
+    re.compile(r"^sops_mac\s*="),  # dotenv: flat ``sops_mac=``
+)
 
 
 def _line_is_dotenvx_encrypted(line: str) -> bool:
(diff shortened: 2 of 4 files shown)


src/envdrift/scanner/engine.py
```diff
@@ -695,11 +694,12 @@ def is_file_encrypted(file_path: str) -> bool:
...
                     content = f.read()
-                    # Use markers from native scanner
-                    for marker in DOTENVX_MARKERS + SOPS_MARKERS:
-                        if marker in content:
-                            encrypted_files.add(file_path)
-                            return True
+                    # Structure-aware encryption check (#348): require the dotenvx
+                    # marker in value position / canonical SOPS envelopes, not a bare
+                    # substring (which misfired on plaintext mentioning "encrypted:").
+                    if _content_is_encrypted(content):
```
jainal09, line 700: 🤖 **envdrift review-bot** · medium · test-gap


This PR changes the inner `is_file_encrypted` closure in `_filter_encrypted_files` (engine.py) to delegate to `_content_is_encrypted` instead of the old substring marker loop:

```python
if _content_is_encrypted(content):
    encrypted_files.add(file_path)
    return True
```

This is the exact location whose behavior matters for issue #348: `_filter_encrypted_files` DROPS findings from files it deems encrypted. Under the old substring logic, a plaintext file that merely mentions the marker in a comment (e.g. `# remember: this file is not encrypted: yet`) was classified as encrypted, so a real secret on a later line was silently dropped — a false negative that hides a leak. I confirmed this by running the old marker loop against `'# ... not encrypted: yet\nAWS_KEY=AKIAIOSFODNN7EXAMPLE\n'` → returns True (finding would be filtered out).

The 6 new tests in `TestStructureAwareEncryptionDetection` all drive `NativeScanner.scan()` and assert on the `unencrypted-env-file`/`unencrypted-secret-file` policy — i.e. they exercise `_is_encrypted`/`_has_sops_markers` in native.py, NOT the `_filter_encrypted_files` drop path in engine.py. The existing `test_filter_encrypted_files_*` tests only assert the positive direction (genuine SOPS/dotenvx files ARE filtered), which still passes under the old buggy code. So if the engine.py change were reverted while keeping native.py, no test would fail.

Per CLAUDE.md ("Every bug fix ships a REAL regression test that FAILS without the fix"), the engine.py false-negative fix should ship its own regression test: build a plaintext file with a real secret plus a comment/value mentioning `encrypted:` (or a mid-line `sops:`), run it through `engine._filter_encrypted_files`, and assert the real finding SURVIVES (is not dropped). Codecov reporting the changed line as "covered" is misleading here — coverage is satisfied by the positive-direction tests, but the false-positive direction (the actual bug) is untested.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 213a3ed


The #348 false-negative fix in `engine._filter_encrypted_files` is correct, but its security-relevant direction (a real secret surviving on a plaintext file that merely *mentions* an encryption marker) has no regression test — reverting the engine.py block leaves every test green.

**Findings (1):**
- [medium] engine.py is_file_encrypted false-negative fix has no regression test — `src/envdrift/scanner/engine.py:700` (inline)

src/envdrift/scanner/native.py
```diff
@@ -64,6 +64,58 @@ def _is_encrypted_value_line(line: str) -> bool:
...
+
+def _content_has_sops_markers(content: str) -> bool:
+    """True if ``content`` carries canonical SOPS structural markers.
+
+    Requires either an ``ENC[AES256_GCM,...]`` value envelope or a top-level
+    ``sops:``/``sops_*`` metadata key (column 0), not a loose substring match.
+    """
+    if _SOPS_ENC_RE.search(content):
+        return True
+    return any(_SOPS_TOPLEVEL_RE.match(line) for line in content.splitlines())
```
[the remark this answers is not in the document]
jainal09: Fixed in c4a929e. `_content_has_sops_markers` now scans **per non-comment line** (same comment filter as the dotenvx path), so a comment like `# SOPS wraps ciphertext as ENC[AES256_GCM,...]` no longer classifies a plaintext file as encrypted. Both detection paths are now consistent. Added a load-bearing native-scanner test `test_comment_mentioning_enc_envelope_does_not_suppress` (fails without the fix).

src/envdrift/scanner/native.py
```diff
@@ -64,6 +64,58 @@ def _is_encrypted_value_line(line: str) -> bool:
...
+_DOTENVX_VALUE_RE = re.compile(
+    r"^\s*[A-Za-z_][A-Za-z0-9_]*\s*=\s*[\"']?encrypted:",
+)
+
+# SOPS: a canonical ``ENC[AES256_GCM,...]`` value envelope (dotenv- or YAML-style).
+_SOPS_ENC_RE = re.compile(r"ENC\[AES256_GCM,")
+
+# SOPS: a top-level ``sops:`` mapping key at column 0 (the metadata block YAML/JSON
+# SOPS appends), e.g. ``sops:`` or ``"sops":`` / ``sops_*`` JSON keys.
+_SOPS_TOPLEVEL_RE = re.compile(r"^\"?sops[:_]")
```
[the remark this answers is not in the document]
jainal09: Valid — fixed in c4a929e. The over-loose `^sops[:_]` regex matched any plaintext var starting with `sops_` (e.g. `sops_token=...`, `sops_enabled=...`), misclassifying the file as encrypted and skipping the unencrypted-file policy. Replaced it with the canonical SOPS metadata markers — `^sops:$` (YAML), `"sops":` (JSON), `sops_version=` and `sops_mac=` (the only keys SOPS emits in its dotenv metadata trailer) — mirroring `SopsBackend.SOPS_METADATA_PATTERNS` so the scanner and the encryption backend agree. Added `test_var_starting_with_sops_underscore_does_not_suppress` (fails without the fix); genuine SOPS dotenv files (`sops_version=`/`sops_mac=`) stay detected via `test_genuine_sops_dotenv_metadata_still_encrypted`.

src/envdrift/scanner/engine.py
jainal09: Addressed in c4a929e. Added an engine-level regression test `test_filter_keeps_lineless_finding_in_plaintext_file_mentioning_marker` in `TestFilterEncryptedFiles` that drives `_filter_encrypted_files` directly: a plaintext `.env.production` whose comment says `# ... not encrypted: yet` and whose value mentions `sops:` mid-line, carrying a **lineless** finding (the external-scanner blob shape that hits the drop path). It asserts the finding SURVIVES (is not dropped as a false negative).

Confirmed load-bearing: reverting `is_file_encrypted` to the old `for marker in DOTENVX_MARKERS + SOPS_MARKERS` substring loop while keeping native.py makes the test fail (`assert [] == [ScanFinding(...)]`) — the file is wrongly deemed encrypted and the lineless finding is dropped. A line-numbered finding wouldn't exercise this (the line-aware filter keeps cleartext lines either way), which is exactly why the prior positive-direction tests didn't catch the regression. The structure-aware `_content_is_encrypted` delegation keeps the file plaintext, so the finding survives.

jainal09: 🤖 **envdrift review-bot**


You're right — this resolves it. Verified against c4a929e: `test_filter_keeps_lineless_finding_in_plaintext_file_mentioning_marker` (test_engine.py:1354, inside `TestFilterEncryptedFiles`) drives `engine._filter_encrypted_files` directly with a lineless finding in a plaintext `.env.production` whose comment says `# ... not encrypted: yet` and whose value is `NOTE=ask sops: ...`, and asserts the finding survives — exactly the false-negative drop path #348 was about. Load-bearing confirmed: the old `for marker in DOTENVX_MARKERS + SOPS_MARKERS: if marker in content` loop flags that content as encrypted (`encrypted:` in the comment + `sops:` in the value → True), so reverting engine.py drops the lineless finding and the assert fails; the new `_content_is_encrypted` skips comment lines and requires `encrypted:` in value position / canonical SOPS markers, so it returns False and the finding is kept. The lineless shape (`line_number=None`) is the right choice since a line-numbered finding would survive either way. Thanks for adding it.

PR #405, opened 2026-06-07
fix(sync): atomic_write via mkstemp to block predictable-tmp symlink

jainal09:
Deep-review finding (#348): `atomic_write` used a predictable `.tmp` path; a precreated symlink in an attacker-writable service dir could redirect the write. Fix: `tempfile.mkstemp` + `os.fchmod(fd)` + atomic rename (preserves an existing destination's mode). POSIX load-bearing regression tests added. Part of #348.

<h3>Greptile Summary</h3>

Replaces the vulnerable predictable-name `.tmp` temp-file pattern in `atomic_write` with `tempfile.mkstemp` (`O_EXCL`, unguessable name, fd owned by us), `os.fchmod` on the fd rather than a path-based chmod, and `path.lstat()` with a symlink guard for mode preservation, closing three symlink-redirection attack vectors. Six POSIX-only regression tests were added, each annotated to fail against the previous implementation.

- **`operations.py`**: the entire write path — mkstemp, fchmod, flush, fsync, replace — is wired through the fd we own; mode preservation now uses `lstat` and skips symlinked destinations so a pre-planted symlink cannot inject a permissive mode onto the new secret.
- **`test_sync_operations.py`**: covers predictable-name symlink, mode-injection via destination symlink, unguessable temp-name + cleanup-on-failure, fd-leak/original-error propagation, mode preservation for real existing files, and the tight-mode default when destination is a symlink.

Safe to merge — the production fix is correct and all three symlink attack vectors are closed; the only note is a test portability nuance on Python ≤3.11 that fails loudly rather than silently.

The `atomic_write` rewrite is technically sound: mkstemp with O_EXCL eliminates the predictable-name race, fd-based fchmod cannot be redirected through a symlink, and lstat with the symlink guard prevents mode injection. The BaseException cleanup handler is intentional and correct. The six regression tests each verify a specific attack scenario and are annotated to fail against the prior implementation. The one flag is that `monkeypatch.setattr(os, "replace", boom)` does not intercept `Path.replace()` on Python ≤3.11 due to pathlib's cached accessor — a test that breaks loudly, not silently, but still worth addressing if older Python versions are supported.

The test in `tests/unit/test_sync_operations.py` that patches `os.replace` to simulate a rename failure deserves a second look if Python 3.11 is in the support matrix.

| Filename | Overview |
|----------|----------|
| src/envdrift/sync/operations.py | Replaces predictable `.tmp` path with `tempfile.mkstemp`, uses `os.fchmod` (fd-based), `path.lstat()` for symlink-safe mode preservation, and `BaseException` cleanup handler — security fix is well-structured and correct. |
| tests/unit/test_sync_operations.py | Six new POSIX-only regression tests covering the symlink, mode-injection, unguessable-name, fd-leak, and cleanup scenarios; the `monkeypatch.setattr(os, "replace", boom)` interception of `Path.replace()` is Python-version-dependent (works on 3.12+, breaks on ≤3.11 where pathlib caches the reference). |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant C as Caller
    participant AW as atomic_write()
    participant FS as Filesystem

    C->>AW: atomic_write(path, content, permissions)
    AW->>FS: "path.parent.mkdir(parents=True)"
    AW->>FS: path.lstat() [no symlink follow]
    alt destination is real file
        FS-->>AW: stat with mode
        AW->>AW: "mode = dest_stat.st_mode & 0o777"
    else destination is symlink or missing
        AW->>AW: "mode = permissions (tight default)"
    end
    AW->>FS: "mkstemp(dir=parent, prefix=.name., suffix=.envdrift-tmp)"
    FS-->>AW: (fd, tmp_name) — O_EXCL, unguessable
    AW->>FS: os.fchmod(fd, mode) [fd-based, no symlink follow]
    AW->>FS: write(content) via fdopen wrapper
    AW->>FS: flush() + fsync(fd)
    AW->>FS: fdopen wrapper closes fd
    AW->>FS: tmp_path.replace(path) — atomic rename
    FS-->>C: success

    note over AW,FS: On any BaseException
    AW->>FS: tmp_path.unlink() [suppress FileNotFoundError]
    AW->>C: re-raise original exception
```

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/348-ops-tmp..."](https://github.com/jainal09/envdrift/commit/ee078cb59af72e1e0a5b925f12e32142603ba72a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36051855)</sub>

### DIFF
--- a/src/envdrift/sync/operations.py
+++ b/src/envdrift/sync/operations.py
@@ -8,6 +8,7 @@ import os
 import re
 import secrets
 import shutil
+import stat
 import tempfile
 from datetime import datetime
 from pathlib import Path
@@ -126,36 +127,52 @@ def atomic_write(path: Path, content: str, permissions: int = 0o600) -> None:
     pre-existing symlink) and applies permissions to the *fd we created* with
     ``os.fchmod`` -- not a path-based ``chmod`` that could be redirected through
     a symlinked sibling. The temp file is then atomically renamed onto ``path``.
-    If the destination already exists, its current mode is preserved instead of
-    unconditionally forcing ``permissions``.
+    If the destination already exists as a *real file*, its current mode is
+    preserved instead of unconditionally forcing ``permissions``; a destination
+    that is a symlink is ignored for mode purposes so a pre-planted symlink
+    cannot inject an overly permissive mode onto the new secret.
     """
     # Ensure parent directory exists
     path.parent.mkdir(parents=True, exist_ok=True)
 
-    # Preserve the destination's existing mode if it already exists; otherwise
-    # use the requested default. Resolve before creating the temp file.
+    # Preserve the destination's existing mode if it already exists as a real
+    # file. Use ``lstat`` (does not follow symlinks) and skip preservation when
+    # the destination is a symlink: an attacker who pre-plants a symlink at the
+    # destination pointing at a 0o777 file they own must not be able to force
+    # that permissive mode onto the freshly written secret.
     mode = permissions
     with contextlib.suppress(FileNotFoundError):
-        mode = path.stat().st_mode & 0o777
+        dest_stat = path.lstat()
+        if not stat.S_ISLNK(dest_stat.st_mode):
+            mode = dest_stat.st_mode & 0o777
 
     fd, tmp_name = tempfile.mkstemp(
         dir=str(path.parent), prefix=f".{path.name}.", suffix=".envdrift-tmp"
     )
     tmp_path = Path(tmp_name)
+    # ``mkstemp`` returns an fd we own. Hand it to ``os.fdopen`` *immediately* so
+    # the wrapper owns it for the entire body and closes it exactly once on
+    # ``__exit__`` (success *or* exception). We never close the fd ourselves --
+    # doing so would either double-close (masking the real error with ``EBADF``
+    # and leaking the temp file) or leak it. ``fchmod`` and ``fsync`` operate on
+    # the same fd via ``fileno()`` while the wrapper still owns it.
     try:
-        # ``os.fchmod`` applies to the fd we own, never a symlink target. It is
-        # absent on Windows, where permission bits are largely a no-op anyway.
-        if hasattr(os, "fchmod"):
-            os.fchmod(fd, mode)
         with os.fdopen(fd, "w") as tmp_file:
+            # ``os.fchmod`` applies to the fd we own, never a symlink target. It
+            # is absent on Windows, where permission bits are largely a no-op.
+            if hasattr(os, "fchmod"):
+                os.fchmod(tmp_file.fileno(), mode)
             tmp_file.write(content)
-        # fdopen has consumed the fd; closing it again would error.
-        fd = -1
+            tmp_file.flush()
+            os.fsync(tmp_file.fileno())
+        # The fdopen wrapper has closed the fd; the rename promotes the temp
+        # file atomically onto the destination (``Path.replace`` delegates to
+        # ``os.replace``).
         tmp_path.replace(path)
     except BaseException:
-        if fd != -1:
-            os.close(fd)
-        # Only ever unlink the temp file we created, never the destination.
+        # The fd is owned/closed by the fdopen wrapper, so we never close it
+        # here. Only ever unlink the temp file we created, never the
+        # destination; suppress its absence so the original error propagates.
         with contextlib.suppress(FileNotFoundError):
             tmp_path.unlink()
         raise
--- a/src/envdrift/sync/operations.py
+++ b/src/envdrift/sync/operations.py
@@ -2,10 +2,13 @@
 
 from __future__ import annotations
 
+import contextlib
 import hashlib
+import os
 import re
 import secrets
 import shutil
+import tempfile
 from datetime import datetime
 from pathlib import Path
 
@@ -118,18 +121,43 @@ def atomic_write(path: Path, content: str, permissions: int = 0o600) -> None:
     """
     Write file atomically with proper permissions.
 
-    Uses a temporary file and rename to ensure atomicity.
+    Creates the temp file in the destination directory with an unguessable name
+    via ``tempfile.mkstemp`` (``O_EXCL``, owned by us, never following a
+    pre-existing symlink) and applies permissions to the *fd we created* with
+    ``os.fchmod`` -- not a path-based ``chmod`` that could be redirected through
+    a symlinked sibling. The temp file is then atomically renamed onto ``path``.
+    If the destination already exists, its current mode is preserved instead of
+    unconditionally forcing ``permissions``.
     """
     # Ensure parent directory exists
     path.parent.mkdir(parents=True, exist_ok=True)
 
-    tmp_path = path.with_suffix(".tmp")
+    # Preserve the destination's existing mode if it already exists; otherwise
+    # use the requested default. Resolve before creating the temp file.
+    mode = permissions
+    with contextlib.suppress(FileNotFoundError):
+        mode = path.stat().st_mode & 0o777
+
+    fd, tmp_name = tempfile.mkstemp(
+        dir=str(path.parent), prefix=f".{path.name}.", suffix=".envdrift-tmp"
+    )
+    tmp_path = Path(tmp_name)
     try:
-        tmp_path.write_text(content)
-        tmp_path.chmod(permissions)
+        # ``os.fchmod`` applies to the fd we own, never a symlink target. It is
+        # absent on Windows, where permission bits are largely a no-op anyway.
+        if hasattr(os, "fchmod"):
+            os.fchmod(fd, mode)
+        with os.fdopen(fd, "w") as tmp_file:
+            tmp_file.write(content)
+        # fdopen has consumed the fd; closing it again would error.
+        fd = -1
         tmp_path.replace(path)
-    except Exception:
-        tmp_path.unlink(missing_ok=True)
+    except BaseException:
+        if fd != -1:
+            os.close(fd)
+        # Only ever unlink the temp file we created, never the destination.
+        with contextlib.suppress(FileNotFoundError):
+            tmp_path.unlink()
         raise
 
 
--- a/tests/unit/test_sync_operations.py
+++ b/tests/unit/test_sync_operations.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+import tempfile
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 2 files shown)


src/envdrift/sync/operations.py
```diff
@@ -118,18 +121,43 @@ def atomic_write(path: Path, content: str, permissions: int = 0o600) -> None:
...
+        with os.fdopen(fd, "w") as tmp_file:
+            tmp_file.write(content)
+        # fdopen has consumed the fd; closing it again would error.
+        fd = -1
         tmp_path.replace(path)
-    except Exception:
-        tmp_path.unlink(missing_ok=True)
+    except BaseException:
+        if fd != -1:
+            os.close(fd)
```
jainal09: 🤖 **envdrift review-bot** · high · bug


In the new `atomic_write`, the sentinel `fd = -1` is only set *after* the `with os.fdopen(fd, "w")` block completes successfully:

```python
with os.fdopen(fd, "w") as tmp_file:
    tmp_file.write(content)
# fdopen has consumed the fd; closing it again would error.
fd = -1
tmp_path.replace(path)
```

If `tmp_file.write(content)` raises (a realistic path — disk-full `ENOSPC`, an I/O error, or an encoding error; the repo even specs `test_disk_full_atomic_write` in docs/specs/integration-tests-spec.md:233 "Atomic write fails, original preserved"), the file object's `__exit__` already closes the fd, but `fd` is still the original (not `-1`). The `except BaseException:` handler then runs `if fd != -1: os.close(fd)` (line 157) on the already-closed fd.

Consequences, all confirmed by running the exact new code in a scratch dir:
1. `os.close(fd)` raises `OSError: [Errno 9] Bad file descriptor`, which **replaces the original write error** as the propagated exception (the real cause survives only as `__context__`). Callers/users see a misleading "Bad file descriptor" instead of e.g. "No space left on device".
2. Because `os.close(fd)` raises before reaching `tmp_path.unlink()`, **the cleanup never runs and the mkstemp temp file is leaked** — directly contradicting this PR's stated goal ("cleanup on failure (no temp leak)") and the in-code comment "Only ever unlink the temp file we created." The leaked temp file contains the partially-written secret content.

Suggested direction: set `fd = -1` immediately after `mkstemp` (the fd is owned by the `fdopen` wrapper as soon as the `with` is entered), or guard the cleanup `os.close` with `contextlib.suppress(OSError)` so the original exception and the `unlink` cleanup both survive. Add a regression test that injects a write failure and asserts (a) the original exception propagates and (b) no `.envdrift-tmp` leftover remains.

tests/unit/test_sync_operations.py
```diff
@@ -219,6 +219,98 @@ def test_atomic_write_overwrites_existing(self, tmp_path: Path) -> None:
...
+
+        os.replace = boom  # type: ignore[assignment]
+        try:
+            with pytest.raises(OSError, match="blocked replace"):
+                atomic_write(dest, "data")
+        finally:
+            os.replace = original_replace  # type: ignore[assignment]
+
+        # The vulnerable predictable name must never have been created...
+        assert not predictable.exists()
```
jainal09, line 316: 🤖 **envdrift review-bot** · low · test-gap


Per CLAUDE.md, every bug fix ships a regression test that FAILS without the fix. I ran all 4 new tests against the old (vulnerable) `atomic_write` extracted from base `eca1a72`: three correctly fail, but `test_atomic_write_does_not_use_literal_predictable_tmp_name` **passes against the old code too**, so it does not actually guard the predictable-temp-name fix.

The test's docstring claims "The vulnerable predictable name must never have been created" and asserts `assert not predictable.exists()` (line 274) — but it makes that assertion only *after* `os.replace` is forced to fail, which triggers the cleanup path. The OLD code does `tmp_path.write_text(content)` (creating the predictable `secret.tmp`), then on the forced `os.replace` failure runs `except: tmp_path.unlink(missing_ok=True)`, deleting `secret.tmp`. So by the time the assertion runs, the predictable file is already gone under the old code too — the assertion can never distinguish old from new. (Note also `Path('secret.txt').with_suffix('.tmp')` is `secret.tmp`, not `secret.txt.tmp`, so the predictable name the test checks does match what the old code created.)

Impact: false confidence — the test gives the appearance of guarding the predictable-name vector while passing whether or not the mkstemp fix is present. Suggested direction: assert the temp name actually used is unguessable while it still exists (e.g. plant a symlink at the predictable name and assert it was not written through — as the first test does for `.env.keys`), or capture the temp path during the write rather than checking existence after cleanup.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit dc95289


The temp-file hardening is sound, but the failure path in the new `atomic_write` has a real bug, and one of the four new tests does not actually guard the fix it claims to.

**Findings (2):**
- [high] Write failure inside `with os.fdopen` double-closes fd, masks the real error, and leaks the temp file (contradicting the PR's no-leak goal) — `src/envdrift/sync/operations.py:157` (inline)
- [low] `test_atomic_write_does_not_use_literal_predictable_tmp_name` passes against the OLD vulnerable code — not a real regression test — `tests/unit/test_sync_operations.py:274` (inline)

src/envdrift/sync/operations.py
```diff
@@ -118,18 +121,43 @@ def atomic_write(path: Path, content: str, permissions: int = 0o600) -> None:
...
     """
     # Ensure parent directory exists
     path.parent.mkdir(parents=True, exist_ok=True)
 
-    tmp_path = path.with_suffix(".tmp")
+    # Preserve the destination's existing mode if it already exists; otherwise
+    # use the requested default. Resolve before creating the temp file.
+    mode = permissions
+    with contextlib.suppress(FileNotFoundError):
+        mode = path.stat().st_mode & 0o777
```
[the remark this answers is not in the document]
jainal09: Fixed in 26da168. Mode preservation now uses `path.lstat()` (does not follow symlinks) and skips preservation when the destination is a symlink, so a pre-planted symlink can no longer inject a permissive mode onto the freshly written secret; the requested tight default `permissions` is applied instead. Regression test `test_atomic_write_symlinked_destination_does_not_inject_permissive_mode` plants a symlink at the destination pointing at an attacker-owned `0o777` file and asserts the new file lands at `0o600` (it fails against the prior `path.stat()` code).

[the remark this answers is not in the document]
jainal09: Fixed in 26da168, exactly as suggested: switched to `path.lstat()` and skip mode-preservation when the destination is a symlink (`stat.S_ISLNK`), so an attacker-planted destination symlink cannot inject a permissive mode. The new secret gets the tight requested default. Covered by a load-bearing regression test that fails against the previous `path.stat()` implementation.

tests/unit/test_sync_operations.py
```diff
@@ -219,6 +219,98 @@ def test_atomic_write_overwrites_existing(self, tmp_path: Path) -> None:
...
+        atomic_write(dest, "SECRET_DEST_CONTENT")
+
+        # The real destination got the content; the outside target is untouched.
+        assert dest.read_text() == "SECRET_DEST_CONTENT"
+        assert outside.read_text() == "ORIGINAL_OUTSIDE"
+        # The destination is a real file, not a symlink to the outside file.
+        assert not dest.is_symlink()
+
+    @pytest.mark.skipif(os.name == "nt", reason="symlink/fchmod semantics differ on non-POSIX")
+    def test_atomic_write_does_not_use_literal_predictable_tmp_name(self, tmp_path: Path) -> None:
```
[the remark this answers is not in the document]
jainal09: Done in 26da168. The temp-name test now takes `monkeypatch` and uses `monkeypatch.setattr` for both `os.replace` and the `tempfile.mkstemp` spy (see `test_atomic_write_uses_unguessable_tmp_name_and_cleans_up_on_failure`).

tests/unit/test_sync_operations.py
```diff
@@ -219,6 +219,98 @@ def test_atomic_write_overwrites_existing(self, tmp_path: Path) -> None:
...
+
+        def boom(src: str | os.PathLike[str], dst: str | os.PathLike[str]) -> None:
+            raise OSError("blocked replace for test")
+
+        os.replace = boom  # type: ignore[assignment]
+        try:
+            with pytest.raises(OSError, match="blocked replace"):
+                atomic_write(dest, "data")
+        finally:
+            os.replace = original_replace  # type: ignore[assignment]
```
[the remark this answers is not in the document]
jainal09: Adopted in 26da168. Replaced the bare `os.replace = boom` global mutation with `monkeypatch.setattr(os, "replace", boom)` so teardown is scoped to the test and safe under `pytest-xdist`.

src/envdrift/sync/operations.py
jainal09: Fixed in 26da168. The mkstemp fd is now handed to `os.fdopen` immediately, and the wrapper owns/closes it exactly once on `__exit__` (success or exception); `fchmod`/`fsync` operate via `tmp_file.fileno()` inside the `with`. The failure path no longer calls `os.close(fd)` at all -- it only unlinks the temp file -- so the original exception (e.g. ENOSPC) propagates instead of being replaced by `EBADF`, and the temp file is always cleaned up (no leak). I reproduced the exact double-close/leak with the prior code (`OSError: [Errno 9] Bad file descriptor` masking `No space left on device`, temp file left behind). New regression test `test_atomic_write_propagates_write_error_without_leaking_temp` injects a write failure and asserts (a) the original error propagates and (b) no `.envdrift-tmp` leftover remains -- it fails against the prior code.

tests/unit/test_sync_operations.py
jainal09: Fixed in 26da168. Removed the non-regressing `test_atomic_write_does_not_use_literal_predictable_tmp_name` (which, as you noted, passed against the old code because cleanup deleted the predictable file before the assertion ran) and replaced it with two load-bearing tests: `test_atomic_write_does_not_write_through_predictable_tmp_name` plants a symlink at `path.with_suffix('.tmp')` and asserts the victim is never written through (the symlink-follow vector), and `test_atomic_write_uses_unguessable_tmp_name_and_cleans_up_on_failure` spies on `tempfile.mkstemp` to capture the temp name *while it still exists* and asserts it is the unguessable mkstemp form, never the predictable sibling -- then asserts cleanup leaves nothing behind.

src/envdrift/sync/operations.py
jainal09: 🤖 **envdrift review-bot**


You're right — resolved in 26da168. The `fd = -1` sentinel and the `os.close(fd)` double-close are gone entirely: `os.fdopen` now owns the fd for the whole body and closes it exactly once on `__exit__`, and the `except BaseException:` only `tmp_path.unlink()`s (with `FileNotFoundError` suppressed) before re-raising. I reproduced the prior code in a scratch dir and confirmed the masking/leak (`OSError: [Errno 9] Bad file descriptor` replacing "No space left on device", `.secret.txt.*.envdrift-tmp` left behind); on 26da168 the original ENOSPC propagates and no temp leaks. Your `test_atomic_write_propagates_write_error_without_leaking_temp` is a real regression test — it fails on the prior code and passes on head. Thanks for the thorough fix.

tests/unit/test_sync_operations.py
jainal09: 🤖 **envdrift review-bot**


Resolved. You're right — the non-regressing test is gone in 26da168 and the two replacements are load-bearing. I extracted the old `atomic_write` from base `eca1a72` (`path.with_suffix(".tmp")` + `write_text`/`chmod`) and ran both new tests against it: `test_atomic_write_does_not_write_through_predictable_tmp_name` fails (old `write_text` follows the planted symlink and clobbers `victim`), and `test_atomic_write_uses_unguessable_tmp_name_and_cleans_up_on_failure` fails on `len(created) == 1` (old code never calls `tempfile.mkstemp`, so the spy captures nothing). Both pass against the head impl, which uses `mkstemp(prefix=f".{path.name}.", suffix=".envdrift-tmp")` — matching the `.secret.txt.…envdrift-tmp` assertion. Thanks for the fix.

PR #406, opened 2026-06-07
fix(scanner): report all secrets per line via finditer

jainal09:
Deep-review finding (#348): the native scanner's per-line pass used `search()` once per pattern, so multiple secrets matching the same pattern on one line only reported the first. Fix: `finditer` + a zero-width trailing boundary on aws-access-key-id so adjacent keys both match. Load-bearing tests added. Part of #348.

## Summary by CodeRabbit

* **New Features**
  * Multiple distinct secrets on the same line now generate separate findings with distinct column positions.
  * AWS Access Key ID detection now identifies adjacent keys previously missed on the same line.

* **Documentation**
  * Updated deduplication behavior clarification: default deduplication groups findings by file, line, rule, and secret value; `--skip-duplicate` deduplicates by secret value across all scanners and locations.

<h3>Greptile Summary</h3>

This PR fixes a silent miss in the native scanner where `re.search()` was used per-line per-pattern, meaning only the first secret matching a pattern on a given line was ever reported. The fix replaces `search()` with `finditer()`, changes the AWS access-key pattern's trailing boundary to a zero-width lookahead (so a single shared delimiter between two adjacent keys is not consumed by the first match), and extends the default dedup key in `ScanEngine._deduplicate` to include the secret hash — preventing the engine from re-collapsing two genuinely distinct same-line findings back into one.

- **`native.py`**: `search()` → `finditer()` in the per-line pattern loop; comment explains the lookahead / delimiter interaction.
- **`engine.py`**: Default dedup key now `(file, line, rule, secret_hash)`, extracted into `_dedup_key()`/`_should_replace()`/`_drop_hashless_duplicates()` helpers; `_drop_hashless_duplicates` post-pass keeps backward-compatible behaviour when one scanner provides a hash and another doesn't.
- **`patterns.py`**: AWS pattern trailing boundary `(?:[^A-Z0-9]|$)` → `(?=[^A-Z0-9]|$)` (lookahead).
- Tests: new `test_engine_dedup.py` module; five new scenarios in `test_native.py` including adjacent-key, exact-column, and end-to-end engine dedup assertions.

Safe to merge — the three-part fix is self-consistent and the new tests confirm all key scenarios including regressions.

The finditer change is a narrow, well-scoped mechanical substitution; the lookahead regex change preserves all single-key behaviour while fixing the adjacent-key case; and the dedup key extension is carefully guarded by _drop_hashless_duplicates so no previously-merged findings accidentally split. End-to-end tests drive the full engine path and a separate counterpart test confirms the same-secret-repeated case still collapses as expected.

| Filename | Overview |
|----------|----------|
| src/envdrift/scanner/native.py | Changed per-line pattern scan from search() to finditer(); new inline comment explains the lookahead/delimiter interaction for AWS keys. No issues found. |
| src/envdrift/scanner/patterns.py | AWS pattern trailing boundary changed from consuming group to zero-width lookahead; comment explains why. Change is correct and backward-compatible for single-key lines. |
| src/envdrift/scanner/engine.py | Default dedup key now includes secret_hash; refactored into _dedup_key/_should_replace/_drop_hashless_duplicates helpers. Logic is sound and the _drop_hashless_duplicates post-pass correctly preserves the one-scanner/no-hash backward-compat case without re-merging distinct same-line secrets. |
| tests/scanner/test_engine_dedup.py | New focused dedup test module replacing the bulkier inline class in test_engine.py; covers default path, skip_duplicate, hashless/hashed interaction, cross-scanner scenario, and deterministic tie-breaking. |
| tests/scanner/test_native.py | Five new tests: two-keys-on-line, adjacent-keys single-delimiter, two api_key assignments, single-key regression, and full end-to-end engine dedup. Column assertions are deliberate anchors for the leading-boundary behaviour. |
| docs/cli/guard.md | New prose paragraph accurately describes the updated default-vs-skip_duplicate dedup semantics; consistent with the code changes. |
| tests/scanner/test_engine.py | Removed the old TestDeduplication class (moved to test_engine_dedup.py); no other changes. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Line of file content] --> B[For each pattern: finditer loop]
    B --> C{match found?}
    C -- yes --> D[Extract secret\ngroup 1 or full match]
    D --> E{EC pubkey\nor entropy filter?}
    E -- skip --> B
    E -- keep --> F[Build ScanFinding\ncol = match.start + 1]
    F --> B
    C -- no more matches --> G[Next pattern]
    G --> H{More patterns?}
    H -- yes --> B
    H -- no --> I[All findings from scanner]
    I --> J[ScanEngine._deduplicate]
    J --> K[First pass: seen dict\nkey = file+line+rule+hash]
    K --> L{Same key exists?}
    L -- no --> M[Add to seen]
    L -- yes --> N[_should_replace?\nseverity > verified > hashed > tie-key]
    N -- yes --> M
    N -- no --> O[Discard]
    M --> P[survivors = seen.values]
    P --> Q{skip_duplicate?}
    Q -- no --> R[_drop_hashless_duplicates\nDrop hashless if co-located hashed\nfinding at same severity exists]
    Q -- yes --> S[No post-pass]
    R --> T[Sort by severity]
    S --> T
    T --> U[unique_findings]
```

<sub>Reviews (7): Last reviewed commit: ["test(scanner): merge duplicate dedup tes..."](https://github.com/jainal09/envdrift/commit/696feccd807fe27c4c778e17e4fd3eebddc93ed4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36051858)</sub>


tests/scanner/test_native.py
```diff
@@ -748,6 +748,93 @@ def test_finding_has_redacted_preview(self, scanner: NativeScanner, tmp_path: Pa
...
+    def test_two_aws_keys_on_one_line_yield_two_findings(
+        self, scanner: NativeScanner, tmp_path: Path
+    ):
+        """Two AWS access keys on a single line each produce a finding (#348).
+
+        The per-line pattern pass used ``search`` (first match only), so the
+        second key on the line was silently dropped. With ``finditer`` both keys
+        are reported, each with its own distinct column number.
+        """
+        config_file = tmp_path / "config.py"
```
jainal09, line 904: 🤖 **envdrift review-bot** · high · bug


PR #348 / this PR's stated goal is "report all secrets per line" so that two AWS keys (or two `api_key=` assignments) on one line are both reported. The `finditer` change makes `NativeScanner._scan_patterns` emit both findings, and the new tests assert that against `NativeScanner.scan(...)` directly. But the real CLI path (`envdrift guard` / `scan`) runs findings through `ScanEngine.scan` -> `_deduplicate`, and with the DEFAULT config (`GuardConfig.skip_duplicate = False`, the documented default per `docs/cli/guard.md` and `config.py:142`) the dedup key is `(file_path, line_number, rule_id)` (`engine.py:591`). Two DISTINCT secrets on the same line share that exact key, so the second is dropped again — the very bug #348 describes.

I verified this end-to-end with the repo's own (unchanged) engine code: feeding two distinct-hash AWS findings at (config.py, line 1, aws-access-key-id, cols 7 and 29) into `ScanEngine._deduplicate` with default config returns **1** finding (col 7, first hash only); only with `skip_duplicate=True` (which keys on `secret_hash`, `engine.py:585`) do both survive. So the fix is effective only at the scanner layer; the default user-facing behavior is unchanged, and counterintuitively the duplicate-SUPPRESSION flag (`--skip-duplicate`) is the only mode that actually preserves both distinct secrets.

This is also a test-fidelity gap: the four new tests call `scanner.scan([tmp_path])` (the `NativeScanner`), never the `ScanEngine`, so they pass while the bug the PR claims to fix persists for real invocations. There is no regression test that drives the engine (or the `envdrift` CLI as a subprocess) to assert both same-line secrets reach output under the default config.

Impact: a `.env`/config line with two different secrets of the same rule still reports only one finding via the CLI — the user-visible fix is incomplete. Suggested direction: include `column_number` (or `secret_hash`) in the default dedup key for same-file/same-line findings so distinct secrets aren't merged, and add a real engine-level (or subprocess CLI) regression test that fails without that change. Note the docs/cli/guard.md claim that the default `--no-skip-duplicate` "Show[s] all findings including duplicates" is contradicted by this collapse and should be reconciled in the same PR.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit b64e6b4


The `finditer` change fixes the scanner layer, but the default CLI dedup undoes it — two distinct same-line secrets still collapse to one for real `envdrift guard`/`scan` invocations, so the user-visible fix is incomplete.

**Findings (1):**
- [high] finditer fix is undone by the engine's default dedup; two same-line secrets still collapse to one through the real CLI — `tests/scanner/test_native.py:760` (inline)

tests/scanner/test_native.py
jainal09: Confirmed and fixed in 8271fb5 — the engine dedup *was* collapsing distinct same-line secrets exactly as described.

**Empirical evidence (unchanged engine on this PR, before the fix), one file `creds.env` with two distinct AWS keys on one line:**

- `NativeScanner.scan` → **2** findings (cols 5, 31; distinct `secret_hash`) — the `finditer` fix works at the scanner layer.
- `ScanEngine.scan` with **default** `GuardConfig` (`skip_duplicate=False`) → **1** finding (col 5 only). The key `(file_path, line_number, rule_id)` at `engine.py:592` merged them.
- `ScanEngine.scan` with `skip_duplicate=True` → **2** findings (keys on `secret_hash`).

So, exactly as you found: the default user-facing path dropped the second secret, and counterintuitively only `--skip-duplicate` preserved both.

**Fix:** the default dedup key now includes the secret hash — `(file_path, line_number, rule_id, secret_hash or None)`. Two distinct secrets on one line have different hashes → both survive; the same secret reported at one location by multiple scanners shares a hash → still collapses to one. A post-pass drops a *hashless* finding when a co-located *hashed* finding exists, preserving the existing "prefer the hashed finding" tie-break (so `test_deduplicate_prefers_secret_hash` and policy-finding dedup both stay green). Policy findings (no hash) keep the historical location-only key.

**Regression tests (real, end-to-end):**
- `test_two_aws_keys_one_line_survive_engine_dedup_default` — drives the real `ScanEngine` (not just the scanner) and asserts both distinct same-line AWS keys survive under the default config (distinct cols + distinct hashes).
- `test_same_secret_repeated_collapses_in_engine_dedup_default` — the same secret repeated on a line still dedups to one.
- `test_deduplicate_distinct_secrets_same_line_both_kept` and `test_deduplicate_hashless_collapses_into_hashed_same_location` — focused `_deduplicate` unit tests.

**Docs:** reconciled `docs/cli/guard.md` `--no-skip-duplicate` to describe the actual semantics (identical secret at one location collapses; distinct secrets on one line both report).

Note: scoped strictly to engine dedup — `native.py` is untouched to minimize the eventual conflict with #404.

tests/scanner/test_native.py
```diff
@@ -748,6 +748,93 @@ def test_finding_has_redacted_preview(self, scanner: NativeScanner, tmp_path: Pa
...
+    ):
+        """Adjacent AWS keys split by a single delimiter both match (#348).
+
+        The aws pattern consumes a trailing boundary char ``(?:[^A-Z0-9]|$)``;
+        that delimiter doubles as the leading boundary of the next key, so
+        ``finditer`` still finds both even with no spare separator between them.
+        """
+        config_file = tmp_path / "config.py"
+        config_file.write_text("KEYS=AKIAIOSFODNN7EXAMPLE,AKIAI44QH8DHBEXAMPLE\n")
+
```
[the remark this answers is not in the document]
jainal09: Fixed in 8271fb5. The docstring for `test_two_adjacent_aws_keys_single_delimiter_yield_two_findings` now describes the zero-width lookahead `(?=[^A-Z0-9]|$)` and explains that, because the lookahead does not consume the trailing delimiter, that same delimiter remains available as the leading boundary of the next key — with a note that a *consuming* trailing group would eat the single delimiter and leave the second key without a leading boundary (the bug this PR fixes).

tests/scanner/test_native.py
jainal09: 🤖 **envdrift review-bot**


You're right — verified against 8271fb5, this resolves the finding. The default dedup key at `engine.py:606-611` is now `(file_path, line_number, rule_id, secret_hash or None)` (vs. the location-only `(file_path, line_number, rule_id)` still on `origin/main`), so two distinct same-line secrets keep distinct keys and both survive. The hashless-collapse post-pass preserves the "prefer the hashed finding" tie-break without merging distinct hashed secrets. The regression coverage is real and decisive: `test_two_aws_keys_one_line_survive_engine_dedup_default` (test_native.py:843) drives the actual `ScanEngine` under default `GuardConfig` and asserts 2 findings with distinct columns and distinct hashes — it would fail on old code (collapses to 1) — and `test_same_secret_repeated_collapses_in_engine_dedup_default` locks in the same-secret non-regression. Docs reconciled at `docs/cli/guard.md:184-187`. Resolving.

jainal09:
Addressed review feedback in 9096769:

- **cubic (engine.py, severity):** extracted the hashless-duplicate prune into `_drop_hashless_duplicates`. A hashless finding is now kept when it carries *higher* severity than every co-located hashed finding, so pruning never lowers the severity reported at a location. Added a regression test (`test_deduplicate_keeps_higher_severity_hashless_over_hashed`).
- **cubic (test_native.py, gitleaks):** the engine-level dedup tests now build `ScanEngine(GuardConfig(use_native=True, use_gitleaks=False, auto_install=False))` so construction never probes/auto-installs optional external scanners before `engine.scanners` is overridden.
- **CodeScene (Complex Method):** the prune extraction lowers `_deduplicate` cyclomatic complexity.
- **CodeScene (functions-per-module on test_engine.py):** the count reflects one focused regression test per behavior; per CLAUDE.md every bug fix ships a real regression test, so the added function is intentional.

docs/cli/guard.md
```diff
@@ -181,6 +181,13 @@ envdrift guard --skip-duplicate
 envdrift guard --no-skip-duplicate
 ```
 
+By default (`--no-skip-duplicate`), findings are deduplicated by file, line, rule, and
+secret value: the same secret reported at one location by several scanners collapses to
```
jainal09: 🤖 **envdrift review-bot** · medium · doc-mismatch


This commit newly adds — in user-facing docs and in two engine.py comments — the claim that under the default `--no-skip-duplicate` the *same* secret reported at one location by several scanners collapses to a single finding. That is not true on the default path.

The default dedup key is `(finding.file_path, finding.line_number, finding.rule_id, finding.secret_hash or None)` (`engine.py:603-608`). But every external scanner prefixes `rule_id` with its own name and `native` uses bare ids, so the rule_id component never matches across two different scanners:
- gitleaks: `rule_id=f"gitleaks-{rule_id}"` (`gitleaks.py:616`)
- trufflehog: `rule_id=f"trufflehog-{detector_name...}"` (`trufflehog.py`)
- detect-secrets: `rule_id=f"detect-secrets-{detector_type...}"` (`detect_secrets.py:522`); talisman/trivy/infisical/kingfisher/git-secrets all similar
- native: bare `rule_id=pattern.id` e.g. `aws-access-key-id` (`native.py:848`)

There is no rule_id normalization before `_deduplicate` (the scan pipeline at `engine.py:511-539` only filters, never rewrites rule_id). So when gitleaks AND native both flag the *same* secret at the same (file,line), they have the same `secret_hash` but different rule_ids -> different default keys -> BOTH survive. The hashless-prune post-pass is also keyed on `(file_path, line_number, rule_id)` (`engine.py:640`), so it too is per-scanner and doesn't merge them.

Reproduced with the repo's exact key logic in a scratch script: two findings, same file/line, same `secret_hash`, rule_ids `aws-access-key-id` (native) and `gitleaks-aws-access-token` (gitleaks) -> 2 survivors, not 1.

Impact: this is a doc/feature mismatch the PR introduces. CLAUDE.md states a doc that misleads is a reportable bug. The behavior itself (no cross-scanner collapse on the default path) is pre-existing, but the PR adds three new statements asserting the opposite: `docs/cli/guard.md` ("the same secret reported at one location by several scanners collapses to a single finding"), the `_deduplicate` docstring (`engine.py:556-557`), and the inline comment (`engine.py:600-602`). Suggested direction: either correct the doc/comments to say dedup is per-(file,line,rule_id) and therefore effectively per-scanner — cross-scanner same-secret merging only happens under `--skip-duplicate` (which keys on `secret_hash` alone, `engine.py:590-591`) — or, if cross-scanner collapse is actually intended on the default path, normalize/strip the scanner prefix from rule_id (or drop rule_id from the default key) before dedup and add a real two-scanner regression test.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 8271fb5


The same-line distinct-secret fix is solid, but this commit's new docs/comments overclaim: under the default `--no-skip-duplicate` the *same* secret found by several scanners does NOT collapse to one finding, because each scanner prefixes `rule_id` with its own name and `rule_id` is part of the default dedup key.

**Findings (1):**
- [medium] Docs (and matching engine.py comments) newly claim cross-scanner same-secret collapse on the default path, but it never happens — rule_id is scanner-prefixed — `docs/cli/guard.md:185` (inline)

jainal09:
Fixed in c23a5c1. You were right — the default path does not collapse the same secret across scanners (rule_ids are scanner-namespaced). I corrected all three false statements (the `_deduplicate` docstring, the inline comment, and `docs/cli/guard.md`) to describe the real behavior, and added `test_deduplicate_same_secret_two_scanners_not_collapsed_on_default_path` asserting both findings survive by default and collapse only under `--skip-duplicate`, so the docs can't drift back.

jainal09:
Fixed in 87e51f3. Introduced a typed `_finding()` factory + `_dedup()` helper so the 12 dedup tests no longer repeat near-identical multi-line `ScanFinding(...)` blocks — each test now passes only the key-affecting fields its scenario needs. Behavior preserved exactly (every file/line/rule/hash/severity that defines a dedup outcome is still set explicitly; only cosmetic fields asserted nowhere move to defaults). All 16 tests pass.

jainal09:
Accepting this advisory with rationale. `_finding` is a typed test-data builder for `ScanFinding`. CodeScene's max-args=4 fits production functions but not a typed object builder: a dedup test can meaningfully vary 9 `ScanFinding` fields (scanner + file_path/line_number/rule_id/severity/column_number/secret_preview/secret_hash/verified), so the builder needs one explicit parameter each. I deliberately kept them explicit rather than `**kwargs`/`dataclasses.replace`, because the typed signature lets pyrefly validate every call site — the `**kwargs`/`replace` form produces 14 pyrefly errors against ScanFinding's strict `__replace__`. The two tools are in direct conflict here; type safety on the test data wins over the arg-count heuristic.

docs/cli/guard.md
jainal09: 🤖 **envdrift review-bot** — resolved ✅


Verified at `696fecc`: `docs/cli/guard.md` now correctly states that the rule is part of the default dedup key and each scanner namespaces its rule IDs, so the same secret flagged by multiple scanners is **not** collapsed on the default path (only `--skip-duplicate`, which keys on the secret value alone, merges them). The `_deduplicate` refactor preserves the distinct-same-line-secret and hashless-tie-break guarantees, and the test move to `test_engine_dedup.py` keeps the key scenarios load-bearing. Doc/behavior now match — thanks for the fix.

PR #407, opened 2026-06-07
fix(sync): skip lone mismatched .env.<env> in auto-detect

### LINKED ISSUE #395: sync: mapping configured for env X silently syncs a single .env.Y file (and Y's key) via legacy auto-detect
## Summary
When a sync mapping is configured for environment **X** (e.g. `production`) but the service folder has **no** `.env.X` and **no** custom-named match, yet contains **exactly one** other `.env.*` file (e.g. `.env.staging`), `resolve_mapping_env_file` falls through to the legacy `detect_env_file`, which returns that file's suffix as the environment. The sync engine then derives the private-key name from **that detected** environment — so a `production`-configured mapping silently operates on `.env.staging` using `DOTENV_PRIVATE_KEY_STAGING` instead of skipping.

## Location / evidence
- `src/envdrift/sync/engine.py:106-109` — `effective_environment = detection.environment` then `effective_key_name = f"DOTENV_PRIVATE_KEY_{effective_environment.upper()}"`. The key name follows the **detected** file's environment, with no check that it matches the mapping's configured `effective_environment`.
- `src/envdrift/env_files.py:179` — steps 1-3 (explicit `env_file`, exact `.env.<env>`, custom-named match) all keep `mapping.effective_environment` canonical; but when they all miss it falls back to `detect_env_file(folder_path, effective_environment)`.
- `src/envdrift/env_files.py:108-110` — `detect_env_file`'s single-`.env.*` branch: `environment = env_file.name[len('.env.'):]` → `EnvFileDetection(env_file, environment, 'found')`. The configured environment is **not** consulted in this branch (only the plain-`.env` branch uses `default_environment`).

## Repro
1. `mapping = {folder_path, environment='production'}` (no `env_file`).
2. Folder contains **only** `.env.staging` (no `.env.production`, no custom-named match).
3. `resolve_mapping_env_file(mapping)` → `detect_env_file` → `EnvFileDetection('.env.staging', 'staging', 'found')`.
4. `_sync_service` proceeds (status is 'found') and uses `DOTENV_PRIVATE_KEY_STAGING` on `.env.staging` — the `production` mapping never skips and never uses a production key.

## Impact
Low severity / edge case (requires exactly one non-matching `.env.*` and no configured-env file). But a `production` mapping silently syncing a `staging` file/key can pull or push the wrong environment's secret.

## Suggested direction (decision needed — possibly intended)
This may be deliberate single-file auto-detect convenience. If strict environment matching is preferred, `detect_env_file`'s single-`.env.*` branch should adopt the suffix environment only when it matches the requested `default_environment` (or when the mapping has no/explicit-default environment); otherwise return `not_found` so `_sync_service` SKIPS. Flagging for a deliberate call rather than a speculative change — happy to implement either way once the intended semantics are confirmed.

## Notes
Distinct from #325 (ephemeral-pull key_name fell back to the *folder name*; fixed). This is the *single-`.env.*`-file* auto-detect path returning a non-configured environment.

*Found via multi-agent deep review of `main` (sync-engine); cited file:line + repro per repo convention.*

jainal09:
Fixes #395: a mapping configured for env X with no `.env.X` but exactly one other `.env.Y` fell through to `detect_env_file`, which adopted Y's suffix → synced `.env.Y` under `DOTENV_PRIVATE_KEY_Y`. Fix: only adopt a lone `.env.<suffix>` when the suffix matches the requested environment; otherwise return not_found so sync/vault-push SKIPS. Load-bearing regression test added.

Closes #395

<h3>Greptile Summary</h3>

This PR fixes issue #395 where a sync mapping configured for environment X (e.g. `production`) with no `.env.X` but a single `.env.Y` (e.g. `.env.staging`) would fall through to `detect_env_file`, which adopted Y's suffix and synced the staging file under `DOTENV_PRIVATE_KEY_STAGING` for a production mapping. The fix adds a suffix-match guard (`_resolve_lone_env_file`) so a lone `.env.*` is only adopted when its suffix equals the requested environment; otherwise `"not_found"` is returned and the mapping is skipped.

- **`env_files.py`**: Introduces `_resolve_lone_env_file`, `_candidate_env_files`, `_one_or_ambiguous`, and `_resolve_explicit_env_file` as private helpers; refactors `detect_env_file` and `resolve_mapping_env_file` to use them, eliminating the duplicate exclusion-filter logic and the unsafe lone-file adoption.
- **Test suite**: The test previously validating the buggy behavior (`environment=\"production\"` + `.env.staging`) is corrected to use `environment=\"staging\"`, and a new load-bearing regression test is added across three test files (`test_env_files.py`, `test_cov_sync_engine.py`, `test_cli_vault_push.py`).

Safe to merge — the change narrows a single detection fallback path, all affected callers are covered by existing and new tests, and no pre-existing correct behavior is altered.

The fix is surgical and well-contained: only the lone-file adoption branch in detect_env_file changes, and every other resolution path (exact match, custom-named files, explicit env_file, plain .env) is untouched. Regression tests were added at the unit, integration, and end-to-end CLI levels, and the previously incorrect test that was validating the buggy behavior has been corrected. The refactoring helpers are straightforward extractions with no logic change.

| Filename | Overview |
|----------|----------|
| src/envdrift/env_files.py | Core fix: adds _resolve_lone_env_file with a suffix-equality guard, extracts _candidate_env_files/_one_or_ambiguous/_resolve_explicit_env_file as clean helpers. Logic is correct and behavior-equivalent for all pre-existing valid cases. |
| tests/unit/coverage/test_cov_sync_engine.py | Class docstring and inline comment updated per prior review thread; new test_lone_other_env_file_is_skipped_not_synced is the key end-to-end regression test for #395. |
| tests/unit/test_cli_vault_push.py | Extracts _setup_push_all_single_mapping helper; refactors the old buggy-behavior test to use environment='staging'; adds test_push_all_skips_lone_other_environment_file regression for #395. |
| tests/unit/test_env_files.py | Fixes test_resolve_mapping_env_file_preserves_legacy_single_env_detection (environment now 'sqa' to match the file); adds test_resolve_mapping_env_file_skips_lone_other_environment_file regression. |
| tests/unit/test_cli.py | Adds two direct unit tests for detect_env_file: mismatched-env returns not_found; matching-env returns found. Good coverage of the new guard in isolation. |
| docs/guides/env-file-sync.md | Documentation updated to accurately describe the new restriction: only a .env.<env> whose suffix matches the mapping's environment is auto-adopted. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[resolve_mapping_env_file] --> B{env_file configured?}
    B -- yes --> C[_resolve_explicit_env_file\nfolder_not_found / found / not_found]
    B -- no --> D{.env.effective_env\nexists?}
    D -- yes --> E[Return found\nexact match]
    D -- no --> F[_match_env_files_for_environment\ncustom-named files]
    F --> G{matches?}
    G -- 1 file --> H[Return found]
    G -- >1 files --> I[Return multiple_found]
    G -- none --> J[detect_env_file\nlegacy fallback]
    J --> K{plain .env\nexists?}
    K -- yes --> L[Return found\ndefault env]
    K -- no --> M{lone .env.* file?}
    M -- yes --> N{suffix ==\nrequested env?}
    N -- yes --> O[Return found]
    N -- no --> P[Return not_found\n⚠️ FIX for #395: skip]
    M -- ">1" --> Q[Return multiple_found]
    M -- none --> R[Return not_found]
```

<sub>Reviews (6): Last reviewed commit: ["test(vault-push): share --all scaffold t..."](https://github.com/jainal09/envdrift/commit/4087a383a87a86df7fadb91f66e170fd9cd10f53) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36051861)</sub>

### DIFF
--- a/src/envdrift/env_files.py
+++ b/src/envdrift/env_files.py
@@ -75,7 +75,12 @@ def detect_env_file(folder_path: Path, default_environment: str = "production") 
 
     Checks for:
     1. Plain .env file (returns default environment)
-    2. Single .env.* file (returns environment from suffix)
+    2. Single .env.* file, but only when its suffix matches ``default_environment``
+
+    A single ``.env.<suffix>`` whose suffix differs from ``default_environment``
+    is *not* adopted: returning it would let a ``production`` lookup sync, say,
+    ``.env.staging`` under ``DOTENV_PRIVATE_KEY_STAGING`` (see #395). Such a
+    mismatch reports "not_found" so the caller skips instead.
 
     Custom service-prefixed names (e.g. ``service.env.docker``) are intentionally
     not handled here; that requires the mapping's environment and lives in
@@ -85,7 +90,7 @@ def detect_env_file(folder_path: Path, default_environment: str = "production") 
     - "found": env file found
     - "folder_not_found": folder doesn't exist (or isn't a directory)
     - "multiple_found": multiple .env.* files exist (ambiguous)
-    - "not_found": no env files found
+    - "not_found": no env files found (or the only .env.* file is for another env)
     """
     if not folder_path.is_dir():
         return EnvFileDetection(None, None, "folder_not_found")
@@ -105,7 +110,14 @@ def detect_env_file(folder_path: Path, default_environment: str = "production") 
         env_file = env_files[0]
         # Extract environment from filename: .env.soak -> soak
         environment = env_file.name[len(".env.") :]
-        return EnvFileDetection(env_file, environment, "found")
+        # Only adopt the suffix's environment when it matches the requested one.
+        # A single ``.env.staging`` must NOT be claimed by a ``production`` lookup:
+        # doing so would sync that file under the wrong DOTENV_PRIVATE_KEY_<ENV>
+        # (see #395). Mismatches fall through to "not_found" so the caller SKIPS
+        # rather than silently operating on a different environment.
+        if environment == default_environment:
+            return EnvFileDetection(env_file, environment, "found")
+        return EnvFileDetection(None, None, "not_found")
 
     if len(env_files) > 1:
         return EnvFileDetection(None, None, "multiple_found")
@@ -144,7 +156,10 @@ def resolve_mapping_env_file(mapping: Any) -> EnvFileDetection:
     3. A custom-named file that belongs to the environment, such as
        ``service.env.<env>`` or ``service-<env>.env`` (and, for a default
        environment, a plain ``service.env``).
-    4. Legacy auto-detection for ``.env`` or a single ``.env.*`` file.
+    4. Legacy auto-detection for a plain ``.env`` (default env) or a single
+       ``.env.<effective_environment>`` file. A lone ``.env.*`` for a *different*
+       environment is not adopted, so the mapping is skipped rather than synced
+       under the wrong key (see #395).
 
     Custom filenames matched via steps 2-3 keep ``mapping.effective_environment``
     as the environment of record. This preserves canonical vault key names even
--- a/src/envdrift/env_files.py
+++ b/src/envdrift/env_files.py
@@ -45,11 +45,16 @@ def _name_encodes_environment(name: str, environment: str) -> bool:
     environment-specific lookup (e.g. "docker") from grabbing a plain file, and
     avoids relabeling a plain file under a non-default environment.
     """
-    if name.endswith(f".env.{environment}"):
-        return True
-    if name == f"{environment}.env":
-        return True
-    if any(name.endswith(f"{sep}{environment}.env") for sep in ("-", ".", "_")):
+    # Suffix conventions ".env.<env>"/"<prefix>.env.<env>" and infix forms
+    # "<prefix>-<env>.env" / ".<env>.env" / "_<env>.env". ``str.endswith`` accepts
+    # a tuple, so all the encoded forms collapse to one check.
+    encoded_suffixes = (
+        f".env.{environment}",
+        f"-{environment}.env",
+        f".{environment}.env",
+        f"_{environment}.env",
+    )
+    if name == f"{environment}.env" or name.endswith(encoded_suffixes):
         return True
     # Plain ".env" / "<prefix>.env" carries no encoded environment -> default only.
     return environment == _DEFAULT_ENVIRONMENT and name.endswith(".env")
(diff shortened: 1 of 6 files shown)


tests/unit/coverage/test_cov_sync_engine.py
```diff
@@ -104,6 +105,33 @@ def test_uses_detected_env_file_for_key_name(
         content = (service_dir / ".env.keys").read_text()
         assert "DOTENV_PRIVATE_KEY_STAGING=secret" in content
```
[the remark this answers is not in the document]
jainal09: Good catch — fixed in b6d5967. Updated the `TestAutoDetectEnvFile` class docstring and the inline comment so they describe the path this case actually exercises: the file suffix matches the mapping environment (`staging`), so `resolve_mapping_env_file` returns "found" via the exact-match path (step 2). The legacy auto-detect / lone-mismatch branch is still covered directly — by `test_lone_other_env_file_is_skipped_not_synced` here, plus the unit tests in `test_cli.py` (`detect_env_file` matching and not_found cases) and `test_env_files.py`.

src/envdrift/env_files.py
```diff
@@ -75,7 +75,12 @@ def detect_env_file(folder_path: Path, default_environment: str = "production")
 
     Checks for:
     1. Plain .env file (returns default environment)
-    2. Single .env.* file (returns environment from suffix)
+    2. Single .env.* file, but only when its suffix matches ``default_environment``
+
+    A single ``.env.<suffix>`` whose suffix differs from ``default_environment``
+    is *not* adopted: returning it would let a ``production`` lookup sync, say,
+    ``.env.staging`` under ``DOTENV_PRIVATE_KEY_STAGING`` (see #395). Such a
+    mismatch reports "not_found" so the caller skips instead.
```
jainal09, line 118: Addressed in b6d5967 with a clean, behavior-preserving extraction: the single-`.env.*` matching branch now lives in `_resolve_lone_env_file`, which pulls those branches out of `detect_env_file` and drops its cyclomatic complexity back below the threshold. No behavior change — `env_files.py` stays at 98% line coverage and the full non-integration suite passes (2258 passed).

tests/unit/test_cli_vault_push.py
```diff
@@ -374,14 +374,14 @@ def test_push_all_detects_env_file_and_environment(
         mock_resolve_backend,
         tmp_path,
     ):
-        """vault-push --all should auto-detect env files and update environment."""
+        """vault-push --all should auto-detect a lone env file matching the env."""
```
jainal09, line 424: This advisory is acknowledged but intentionally not refactored. `test_push_all_skips_lone_other_environment_file` is the negative (#395 regression) counterpart to `test_push_all_detects_env_file_and_environment`: same `vault-push --all` arrange/act scaffold, opposite assertions (the matching case pushes under `DOTENV_PRIVATE_KEY_STAGING`; the mismatched case asserts `encrypt_calls == []` / `set_secret.assert_not_called()`). The structural overlap CodeScene flags is the standard arrange/act/assert shape of two paired CLI tests. Hoisting it into a shared helper would couple the positive and negative cases and hide each test's setup, hurting readability for no real DRY win — these are independent scenarios, not copy-paste drift. Leaving as-is per the "real tests, readable in isolation" convention.

jainal09:
CodeScene flags module-mean cyclomatic complexity 5.57 vs a threshold of 4 — an aggregate advisory, not a correctness issue. The fix already extracts `_resolve_lone_env_file` to keep per-function complexity low; driving the module *mean* under 4 would require over-fragmenting small, cohesive helpers. Accepting the advisory.

Issue #413, opened 2026-06-07
Reliability audit (round 2): 41 adversarially-verified findings across vault, sync, encryption, scanner, CLI, agent, docs


jainal09:
## Context

Adversarially-verified deep reliability audit of `main` (`2bf7075`), run as a multi-agent workflow: 9 subsystem finders (vault backends, sync, encryption, scanner/guard, CLI surface, config/schema, Go agent, docs-vs-behavior, filesystem/cross-platform) read the real code + ran the CLI, then **every candidate was adversarially verified** by an independent agent that tried to refute it against the actual code.

**Result: 41 of 44 candidates confirmed** (1 critical, 12 high, 17 medium, 11 low). The report below deduplicates/merges related findings (e.g. the two `guard --config` paths share one call site) into a grouped, severity-ranked list. It deliberately excludes the 27 already-fixed #348 findings and the merged Playground/#283/#285 work — these are *remaining* gaps.

Each finding carries `file:line`, a concrete repro, and a suggested fix; verification corrections are folded in (look for *(Correction: …)*). This is a tracking umbrella in the style of #348 — pick findings off via small themed PRs.

> Methodology note: 54 agents, ~2M tokens. Confirmed findings only (confidence ≥ 6 after refutation). A handful of candidate findings were dropped as false positives during verification.

---

# envdrift Reliability Audit — Synthesized Report

## Executive Summary

**35 confirmed findings** (each adversarially verified) across vault backends, partial-encryption, the secret scanner/guard, the sync engine, CLI commands, docs, and the Go agent. The most serious themes are **silent secret leakage** (plaintext secrets committed or pushed because of over-eager "already encrypted" short-circuits and a missing `.gitignore` guard) and **machine-readable-output corruption** (`guard`/`diff` emit tracebacks or prose to stdout in `--json`/`--sarif` mode, plus exit-code gating that never fails CI on LOW findings). A second cluster is **inconsistent error handling** — transport/credential failures and config typos escape as raw tracebacks because only a subset of exception types are caught.

**Fix first (top 5):**
1. **Mixed-state `.secret` plaintext leak** (critical) — `partial_encryption.py`
2. **`pull --merge` writes decrypted combined file without `.gitignore`** (high) — `sync.py`
3. **Per-line encryption skip drops adjacent real secrets** (high, false-negative) — `native.py`
4. **`guard --ci --fail-on low` never fails CI** (high) — `guard.py`/`scanner/base.py`
[...]

PR #416, opened 2026-06-07
fix(partial-encryption): re-encrypt mixed-state .secret files to stop plaintext leak

jainal09:
Fixes findings from #413 (Cluster A — partial-encryption secret leakage).

### [CRITICAL] Mixed-state `.secret` file leaks newly-added plaintext secrets into committed files
`src/envdrift/core/partial_encryption.py:308-312` (`encrypt_secret_file`), `456-461` (`push_secrets_only`), `109-159` (`is_file_encrypted`)

`encrypt_secret_file()` / `push_secrets_only()` early-returned whenever
`is_file_encrypted()` was True, and that helper returns True on the **first**
ciphertext value. A file in a mixed state (some values already `encrypted:`, one
newly-added plaintext value) was therefore treated as fully encrypted: the new
plaintext secret was never encrypted and flowed verbatim into both the committed
`.secret` file and the generated combined file.

Fix: a new `has_plaintext_secret_value()` helper detects a leftover non-empty,
non-`DOTENV_PUBLIC_KEY` plaintext value. The push path now skips re-encryption
**only** when the file is fully encrypted (ciphertext present **and** no
plaintext secret value); a mixed file is re-encrypted via the idempotent
`dotenvx encrypt` (it re-encrypts only the plaintext values and leaves existing
ciphertext untouched). The fully-encrypted skip path is unchanged.

### [LOW] `combine_files` silently drops user comment lines beginning with `#/`
`src/envdrift/core/partial_encryption.py:219`

`combine_files` filtered every line whose stripped form started with `#/`,
silently removing a legitimate user comment such as `#/ note about this key`
from the generated combined file. It now strips only the precise 4-line dotenvx
public-key header block (mirroring `DotenvxWrapper.DOTENVX_HEADER_BLOCK_PATTERN`
in `integrations/dotenvx.py`), preserving user `#/` comments and the
`DOTENV_PUBLIC_KEY` assignment.

## Regression tests (all verified to fail on the unfixed source, pass with the fix)

Unit (`tests/unit/test_partial_encryption.py`):
- `test_has_plaintext_secret_value_*` — fully-encrypted / mixed / public-key+comment cases
- `test_encrypt_secret_file_reencrypts_mixed_state` — `dotenvx.encrypt` is invoked on a mixed file
- `test_push_secrets_only_reencrypts_mixed_state` — mixed file re-encrypted, fully-encrypted one skipped
- `test_combine_files_preserves_user_hash_slash_comment` — user `#/` comment survives into the combined file
- updated `test_push_secrets_only_skips_already_encrypted` to use a genuinely fully-encrypted fixture

Integration, real `dotenvx` binary (`tests/integration/test_partial_encryption_e2e.py`):
- `TestMixedStateReEncryption::test_combine_push_reencrypts_newly_added_plaintext_real_binary`
  — encrypt a `.secret`, append a new plaintext var, re-push; asserts both the
  `.secret` and the combined file end fully encrypted with no plaintext leak.

## Docs
- `docs/guides/partial-encryption.md` — clarified the idempotency note for the mixed-state case.

## Gates
`ruff check` / `ruff format --check` / `pyrefly` (src + touched test files) / `bandit` all green;
touched `partial_encryption.py` at 95% line coverage. Targeted unit + integration suites pass (82 tests).

<h3>Greptile Summary</h3>

This PR fixes a critical plaintext-leak bug in `partial_encryption.py` where mixed-state `.secret` files (one ciphertext value already present, one freshly-added plaintext value) were incorrectly treated as fully encrypted, causing the new secret to be committed verbatim. It also fixes `combine_files` silently dropping legitimate user `#/` comment lines.

- Introduces `has_plaintext_secret_value()` and `_is_fully_encrypted()` predicates, replacing bare `is_file_encrypted()` at all three call sites: `encrypt_secret_file`, `_encrypt_secrets_only_file`, and the `push --check` dry-run path.
- Replaces the overly-broad `#/`-line filter in `combine_files` with a regex that strips only the exact 4-line dotenvx public-key header block, preserving user comments.
- Adds precise SOPS metadata key exclusions (`_SOPS_METADATA_SCALAR_KEYS` / `_SOPS_METADATA_GROUP_KEY`) so a fully SOPS-encrypted `.secret` is not spuriously flagged as holding leftover plaintext secrets.

Safe to merge — the fix is logically complete, all three early-return call sites have been updated, and both the unit and integration test suites cover the mixed-state and SOPS-trailer scenarios.

All three call sites that decide 'already pushed' now use the shared _is_fully_encrypted() predicate, including the dry-run --check path that was flagged in the previous review round. The SOPS metadata exclusion is precise (exact key family, not a bare prefix), preventing both false positives (SOPS file spuriously re-encrypted) and false negatives (user var named sops_token silently skipped). The combine_files fix is narrowly scoped to the exact 4-line header block. Test coverage is thorough across unit, integration, and real-binary paths.

| Filename | Overview |
|----------|----------|
| src/envdrift/core/partial_encryption.py | Core fix: adds has_plaintext_secret_value(), _is_fully_encrypted(), and SOPS metadata exclusions; updates all three early-return call sites to use the fully-encrypted predicate; replaces overly-broad #/ line filter with a regex-anchored 4-line header strip. No bugs found. |
| tests/unit/test_partial_encryption.py | Adds regression tests for mixed-state re-encryption, #/ comment preservation, and --check dry-run out-of-sync detection; updates the skips-already-encrypted test to use a genuinely fully-encrypted fixture. |
| tests/unit/test_partial_encryption_sops.py | New module covering has_plaintext_secret_value() edge cases: fully encrypted, mixed state, SOPS metadata trailer, quoted empty values, and user vars starting with sops_ that must still be scanned. |
| tests/integration/test_partial_encryption_e2e.py | Adds TestMixedStateReEncryption with a real dotenvx binary test: encrypt, append plaintext, push again, assert both .secret and combined file hold only ciphertext. |
| tests/integration/test_partial_encryption_sops_e2e.py | New integration module for real sops binary: verifies a genuine SOPS-encrypted .secret is recognized as fully encrypted and push --check reports in_sync=True. Correctly skips when sops binary is absent. |
| docs/guides/partial-encryption.md | Updates --check description and idempotency bullet to reflect mixed-state handling and SOPS metadata trailer semantics. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[push / encrypt called] --> B{secret_file.exists?}
    B -- No --> Z[return / skip]
    B -- Yes --> C[_is_fully_encrypted]
    C --> D{is_file_encrypted?}
    D -- No --> G[run dotenvx.encrypt]
    D -- Yes --> E{has_plaintext_secret_value?}
    E -- Yes mixed state --> G
    E -- No fully encrypted --> F[skip — lift skip-worktree]
    G --> H[_git_unskip_worktree]
    H --> I[done — no plaintext leak]
```

1. `src/envdrift/core/partial_encryption.py`, line 461-462 ([link](https://github.com/jainal09/envdrift/blob/d574b0fdf86e5d845bd63d6424e9556e0ee74d37/src/envdrift/core/partial_encryption.py#L461-L462)) 

   <a href="#">P1</a> The `push_partial_encryption` dry-run path uses only `is_file_encrypted()` to decide whether to set `in_sync = False`. For a mixed-state file (`is_file_encrypted()` returns True because there is at least one ciphertext value, but `has_plaintext_secret_value()` is also True), this condition is never entered, so `in_sync` stays True — the same false-positive the main push path was just fixed to avoid. A `--check` run on a mixed-state file would therefore report the project as already pushed when a real push would still re-encrypt it.
</details>

<sub>Reviews (7): Last reviewed commit: ["fix(partial-encryption): match exact SOP..."](https://github.com/jainal09/envdrift/commit/23e094801ff26aa63b5712afdc45c27ad1b29d04) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36083401)</sub>


jainal09:
Addressed in 2c74e7af6ffc41f87ab10d0fa4ac1067d7b926e4:

- **cubic P2 (quoted-empty values)**: `_unquote_value()` now strips a single layer of matching quotes *and* re-strips, so `KEY=""` / `KEY=''` / `KEY=" "` normalize to `""` and are correctly skipped as carrying no secret. A fully-encrypted file with empty placeholders is no longer misclassified as holding plaintext. Added `test_has_plaintext_secret_value_quoted_empty_values_are_not_secrets` (and a quoted-plaintext counter-test).
- **CodeScene Complex Method / Complex Conditional on `has_plaintext_secret_value`**: extracted `_line_has_plaintext_secret()` (flat early-returns), so the function is now a single `any(...)` over lines. CC 9 -> 3; the complex conditional is gone.
- **CodeScene Complex Method on `push_secrets_only`**: extracted `_is_pushable_secret_file()` and `_encrypt_secrets_only_file()`; loop body is now flat. CC 10 -> 4.

Also added unit tests for the `push_secrets_only(check=True)` dry-run branches so codecov/patch covers the new lines.

src/envdrift/core/partial_encryption.py
```diff
@@ -8,6 +8,7 @@
 from __future__ import annotations
 
 import logging
+import re
```
jainal09, line 11: Refactored in 949713f: the 4-branch matched-quote test is extracted into a small `_is_quote_wrapped()` predicate, so `_unquote_value` now has a single straight-line conditional. No suppression.

src/envdrift/core/partial_encryption.py
```diff
@@ -93,19 +112,74 @@ def file_not_found(cls, path: Path) -> PartialEncryptionError:
...
+def _line_has_plaintext_secret(line: str) -> bool:
+    """Return True if a single ``.secret`` line is a PLAINTEXT secret assignment.
+
+    Comments, blank lines, non-assignments, dotenvx's ``DOTENV_PUBLIC_KEY_*``
+    artifact (a public key is not a secret) and empty assignments
+    (``KEY=`` / ``KEY=""`` / ``KEY=''``) all carry no secret to leak and return
+    False. A real assignment whose value is not ciphertext returns True.
+    """
+    stripped = line.strip()
+    if not stripped or stripped.startswith("#") or "=" not in stripped:
```
jainal09: Refactored in 949713f: the comment/blank/`=` guard now reuses the existing `_is_secret_var_line()` predicate (which already filters comments, non-assignments and the DOTENV_PUBLIC_KEY artifact), removing the complex conditional. Behaviour is unchanged. No suppression.

src/envdrift/core/partial_encryption.py
```diff
@@ -372,9 +475,11 @@ def push_partial_encryption(
...
         check: When True, run as a dry run — do not encrypt and do not write the
             combined file. ``in_sync`` is True only when BOTH the combined file
-            already matches the source files AND the secret file is encrypted. A
-            plaintext secret file always reports out of sync (a real push would
-            encrypt it), even if the combined file happens to match.
+            already matches the source files AND the secret file is FULLY
+            encrypted. A plaintext or mixed-state secret file (a freshly-added
+            plaintext value alongside existing ciphertext) always reports out of
+            sync — a real push would re-encrypt it — even if the combined file
+            happens to match.
```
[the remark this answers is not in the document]
jainal09: Fixed in cdf13e8. Extracted a shared `_is_fully_encrypted()` predicate (`is_file_encrypted() and not has_plaintext_secret_value()`) and used it in all three call sites that decide "already pushed": `encrypt_secret_file`, `_encrypt_secrets_only_file` and the `--check` dry-run path. The dry-run now sets `in_sync = False` when `not _is_fully_encrypted(secret_file)` — logically equivalent to the suggested `not is_file_encrypted() or has_plaintext_secret_value()` (De Morgan), but sharing one predicate so the three paths cannot drift. Added regression `test_push_partial_check_out_of_sync_when_secret_mixed_state` (mixed secret + byte-for-byte matching combined file → in_sync False) plus docstring/guide updates.

jainal09:
Bot review summary: reviewed latest head `cdf13e8` for PR #416 in an isolated checkout, including partial-encryption mixed-state logic, docs, unit/integration coverage, linked #413 context, visible comments, and CI status.

No new findings in this pass. I read the partial-encryption patch and tests, then ran `uv run --extra dev pytest tests/unit/test_partial_encryption.py -q` (70 passed), the new `tests/integration/test_partial_encryption_e2e.py::TestMixedStateReEncryption` test (1 passed), and `git diff --check`. Visible checks are green, with Cloud Provider Tests skipped.

src/envdrift/core/partial_encryption.py
```diff
@@ -93,19 +112,79 @@ def file_not_found(cls, path: Path) -> PartialEncryptionError:
...
+    return bool(_unquote_value(value)) and not _value_is_ciphertext(value)
+
+
+def has_plaintext_secret_value(file_path: Path) -> bool:
+    """Return True if the file still holds at least one PLAINTEXT secret value.
+
+    A "plaintext secret value" is a real variable assignment whose value is not
+    ciphertext — excluding comments, blank lines and dotenvx's
+    ``DOTENV_PUBLIC_KEY_*`` artifact (a public key is not a secret).
+
```
jainal09, line 234: 🤖 **envdrift review-bot** · high · regression


`_is_fully_encrypted = is_file_encrypted() and not has_plaintext_secret_value()` regresses the handling of SOPS-encrypted `.secret` files.

A genuine SOPS-encrypted dotenv file carries a flat metadata trailer of *plaintext* `KEY=value` lines — `sops_version=3.13.1`, `sops_lastmodified=2021-...`, `sops_unencrypted_suffix=_unencrypted` (only `sops_mac=ENC[...]` is ciphertext). This module itself documents that format: `is_file_encrypted`'s SOPS belt-and-suspenders branch keys off `"sops_version" in content` (partial_encryption.py:238), and the dedicated test `test_is_file_encrypted_with_sops_value` plus the #352 work intentionally treat a SOPS `.secret` as "already encrypted" so the push path SKIPS it.

`has_plaintext_secret_value()` walks every line through `_line_has_plaintext_secret` → `_is_secret_var_line`, which only excludes comments, lines without `=`, and the `DOTENV_PUBLIC_KEY` artifact. The SOPS trailer lines `sops_version=...` / `sops_lastmodified=...` / `sops_unencrypted_suffix=...` are non-comment, contain `=`, are non-empty, and are not ciphertext — so each is reported as a "plaintext secret value". `has_plaintext_secret_value()` therefore returns True for a *fully* SOPS-encrypted file, and `_is_fully_encrypted()` returns False.

I confirmed this by extracting the head helpers verbatim into /tmp and running them against a realistic SOPS dotenv file: `is_file_encrypted=True`, `has_plaintext_secret_value=True`, `_is_fully_encrypted=False`, with the three trailer lines flagged.

Impact at all three call sites that now use `_is_fully_encrypted`:
- `encrypt_secret_file` (line ~410): runs `dotenvx.encrypt()` on a SOPS-encrypted file on EVERY push (pre-PR it was skipped because `is_file_encrypted` returned True). dotenvx encrypting a SOPS file either raises `DotenvxError` → `PartialEncryptionError` (a previously-working push now fails) or rewrites the file (corrupting/double-wrapping the SOPS structure).
- `_encrypt_secrets_only_file` (line ~327): same — counts a SOPS file as needing encryption every push.
- `push_partial_encryption(check=True)` (line ~482): reports a fully-SOPS-encrypted `.secret` as perpetually out-of-sync, a false positive that would break a `push --check` CI gate.

Test gap: the only SOPS coverage is `test_is_file_encrypted_with_sops_value`, a single `ENC[...]` line with NO metadata trailer — exactly the case that does NOT trip the new helper — so the regression is uncovered. A real fix would exclude SOPS metadata keys (e.g. reuse the canonical `sops_version=`/`sops_mac=`/`sops_*` recognition the project already centralizes) from the plaintext-secret scan, and add a test using a full SOPS metadata trailer.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit cdf13e8


The new `_is_fully_encrypted` predicate fixes the mixed-state dotenvx case but regresses fully-SOPS-encrypted `.secret` files: their plaintext `sops_*` metadata trailer is misread as a plaintext secret, so a previously-skipped SOPS file is now re-encrypted (corrupting it) on every push and reported out-of-sync under `--check`.

**Findings (1):**
- [high] has_plaintext_secret_value() flags SOPS metadata trailer as plaintext → fully-SOPS-encrypted .secret is never treated as encrypted — `src/envdrift/core/partial_encryption.py:167` (inline)

jainal09:
Fixed in e9540bb. Verified the regression against the **real sops binary**: `sops --encrypt --input-type dotenv --output-type dotenv --in-place` writes a flat PLAINTEXT metadata trailer (`sops_age__list_0__map_recipient=`, `sops_lastmodified=`, `sops_unencrypted_suffix=`, `sops_version=`; only `sops_mac=` is ciphertext). Pre-fix `has_plaintext_secret_value()` flagged all 5 plaintext `sops_*` lines as leftover secrets, so `_is_fully_encrypted()` returned False for an already-SOPS-encrypted file.

Root cause + fix: `_is_secret_var_line()` now excludes the canonical `sops_` metadata family (the same prefix `SOPSEncryptionBackend.SOPS_METADATA_PATTERNS` and the scanners key off), so the plaintext scan and the `secret_var` count both ignore SOPS bookkeeping. A freshly-added **non-`sops_`** plaintext value in a mixed SOPS file is still caught (re-encryption still fires).

Regression tests (each confirmed to FAIL on the reverted source hunk, PASS with the fix):
- unit (`tests/unit/test_partial_encryption.py`): full SOPS trailer fixture -> `has_plaintext_secret_value` False, `_is_fully_encrypted` True, `push --check` in sync; plus a guard that a new non-`sops_` plaintext value still leaks.
- integration (`tests/integration/test_partial_encryption_e2e.py`, skip-gated on the `sops` binary): the **real sops binary** encrypts a dotenv `.secret` in place; the helpers and `push --check` recognize it as fully encrypted.

Touched-file coverage: `partial_encryption.py` at 97%. Gates green (ruff/format/pyrefly/bandit). Docs updated in `docs/guides/partial-encryption.md`.

jainal09:
Bot review summary for latest head `e9540bbc`:

I re-reviewed the partial-encryption mixed-state fix after the new SOPS metadata commit. The SOPS metadata regression is addressed: a realistic SOPS dotenv trailer now gives `has_plaintext_secret_value=False`, `is_file_encrypted=True`, and `_is_fully_encrypted=True`, while adding a fresh non-`sops_` plaintext value flips `has_plaintext_secret_value=True` as expected.

No new findings in this pass. Current visible checks are not failing.

tests/integration/test_partial_encryption_e2e.py
```diff
@@ -635,6 +635,102 @@ def test_plaintext_value_literally_containing_encrypted_prefix_is_not_encrypted(
         assert is_file_encrypted(secret) is False
 
 
+# ---------------------------------------------------------------------------
```
jainal09: Addressed in 9063f40: extracted the cohesive real-sops e2e class (`TestSopsEncryptedSecretIsFullyEncryptedRealBinary`) into a focused sibling module `tests/integration/test_partial_encryption_sops_e2e.py` (gated on the sops binary). The original `test_partial_encryption_e2e.py` is now well under the 600-LoC threshold. Pure relocation, no assertions changed; all 19 partial-encryption integration tests still pass.

tests/unit/test_partial_encryption.py
```diff
@@ -14,6 +14,7 @@
 from envdrift.core.partial_encryption import (
     PartialEncryptionError,
     combine_files,
+    has_plaintext_secret_value,
```
jainal09, line 17: Addressed in 9063f40: moved the cohesive #416 value-scan / fully-SOPS-encrypted-`.secret` family (the `has_plaintext_secret_value` tests and sops metadata-trailer tests, plus their `_sops_ciphertext` helper and `_SOPS_ENCRYPTED_DOTENV` fixture) into a focused sibling module `tests/unit/test_partial_encryption_sops.py`. The original module drops from 79 to 69 functions, under the 75 threshold. Pure relocation, no assertions changed; all 74 partial-encryption unit tests still pass.

src/envdrift/core/partial_encryption.py
```diff
@@ -65,12 +99,16 @@ def _build_warning_header(clear_file: str, secret_file: str) -> str:
...
-    reported secret-var count reflects only the actual secrets.
+    Excludes comments, dotenvx's ``DOTENV_PUBLIC_KEY_*`` artifact and SOPS's
+    ``sops_*`` metadata trailer (version/timestamp/recipient/mac bookkeeping, not
+    real secrets — see #416) so the reported secret-var count and the plaintext
+    scan reflect only the actual secrets.
     """
     stripped = line.strip()
     if "=" not in stripped or stripped.startswith("#"):
         return False
+    if stripped.startswith(_SOPS_METADATA_KEY_PREFIX):
```
jainal09: 🤖 **envdrift review-bot** · medium · bug


The SOPS-trailer fix excludes **any** assignment whose stripped key starts with the literal prefix `sops_` from the secret scan:

```python
if stripped.startswith(_SOPS_METADATA_KEY_PREFIX):  # _SOPS_METADATA_KEY_PREFIX = "sops_"
    return False
```

This is correct for genuine SOPS bookkeeping, but it is a *prefix* match, not a match against the known SOPS metadata key family. A user variable that happens to be named with a lowercase `sops_` prefix (e.g. `sops_token`, `sops_api_key`, `sops_config`) is now treated as non-secret bookkeeping.

Impact (same leak class this PR set out to fix): in a MIXED-STATE dotenvx `.secret` (existing `encrypted:` value + a freshly-added plaintext `sops_token=...`), the new plaintext value is excluded from `has_plaintext_secret_value()`, so `_is_fully_encrypted()` returns True and the push path early-returns without re-encrypting — shipping that plaintext secret verbatim into the committed `.secret` and the generated combined file. I reproduced this with the extracted head helpers: a file containing `API_KEY=encrypted:...` plus `sops_token=my-brand-new-plaintext-secret` gives `is_file_encrypted=True`, `has_plaintext_secret_value=False`, `_is_fully_encrypted=True` → push SKIPS → leak.

This is a NEW regression: pre-PR `_is_secret_var_line` did not exclude `sops_` lines, so such a value was correctly flagged as plaintext and re-encrypted. Probability is low (env var names are conventionally uppercase, and a real SOPS file does not mix dotenvx ciphertext with arbitrary user `sops_` keys), but the impact is a silent plaintext-secret leak into a committed file — exactly the failure mode #413/#416 target.

Suggested direction (do not implement here): match against the canonical SOPS metadata key set rather than a bare prefix — e.g. restrict the exclusion to the known `sops_version`/`sops_mac`/`sops_lastmodified`/`sops_unencrypted_suffix`/`sops_age_*`/`sops_kms_*`/`sops_pgp_*`/`sops_unencrypted_*` family (the same family `SOPSEncryptionBackend.SOPS_METADATA_PATTERNS` recognizes), or only apply the exclusion when the file is actually SOPS-shaped (contains `sops_mac=`/`sops_version=` + `ENC[AES256_GCM,`). Add a mixed-state test where the new plaintext value is named `sops_<x>` and assert it is still flagged.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit e9540bb


The SOPS-trailer fix is correct for genuine SOPS bookkeeping, but it now excludes **any** key starting with the bare literal prefix `sops_` from the plaintext-secret scan. A user secret named `sops_*` in a mixed-state dotenvx `.secret` is silently treated as bookkeeping and ships unencrypted — the same leak class #413/#416 set out to fix. Notably, `scanner/native.py` already rejected this exact bare-prefix shape (#348) in favor of the canonical `sops_version=`/`sops_mac=` markers; this module reintroduces it.

**Findings (1):**
- [medium] Broad `sops_` prefix exclusion can silently leak a user secret whose name starts with `sops_` in a mixed-state file — `src/envdrift/core/partial_encryption.py:110` (inline)

jainal09:
Bot re-review update for latest head `9063f400`:

The new commit only relocates the SOPS regression tests into focused unit/integration modules. I re-ran the helper predicate on this head and it still behaves correctly: SOPS metadata trailer => `has_plaintext_secret_value=False`, `is_file_encrypted=True`, `_is_fully_encrypted=True`; adding a fresh non-`sops_` plaintext value makes `has_plaintext_secret_value=True`.

Focused verification: `uv run --extra dev pytest tests/unit/test_partial_encryption_sops.py -q` passed (`9 passed`). No new findings on this head.

src/envdrift/core/partial_encryption.py
jainal09: 🤖 **envdrift review-bot** — resolved (with follow-up) ✅

Verified at `e9540bb`: SOPS metadata trailers are no longer mis-flagged as plaintext — this finding is resolved. ⚠️ However the fix's exclusion is a bare `stripped.startswith("sops_")`, which is too broad: a real plaintext secret var named `sops_*` would now be skipped (plaintext-leak false-negative). Filed as a follow-up finding on this PR — recommend anchoring to the known sops metadata keys instead of a bare prefix.

src/envdrift/core/partial_encryption.py
jainal09: Fixed in 23e0948. Replaced the bare `startswith("sops_")` prefix match with `_is_sops_metadata_key`, which matches only the canonical SOPS metadata family:

- fixed scalar bookkeeping keys: `sops_version`, `sops_mac`, `sops_lastmodified`, `sops_unencrypted_suffix`/`sops_encrypted_suffix`, `sops_unencrypted_regex`/`sops_encrypted_regex` (+ comment variants), `sops_mac_only_encrypted`, `sops_shamir_threshold`;
- nested key-group provider entries anchored on the provider token: `sops_(age|pgp|kms|gcp_kms|azure_kv|hc_vault)` followed by `_`/`__`/end, so `sops_age__list_0__map_recipient` matches but `sops_agent`/`sops_token` do not.

A user var like `sops_token=AKIA…` / `sops_api_key=…` now stays in the plaintext-secret scan, so the mixed-state leak path you described no longer drops it. Regression tests: `test_has_plaintext_secret_value_user_var_with_sops_prefix_is_scanned` (asserts `sops_token`/`sops_api_key`/`sops_config`/`sops_agent`/`sops_keys` are still flagged — fails on the prefix-match code), `test_has_plaintext_secret_value_genuine_sops_metadata_keys_excluded`, and `test_is_sops_metadata_key_exact_family`. Verified the first fails without the fix and passes with it.

jainal09: 🤖 **envdrift review-bot** — resolved ✅

Verified at `23e0948`: `_is_sops_metadata_key` now matches only the canonical SOPS metadata key family (sops_version/sops_mac/sops_lastmodified/sops_unencrypted_suffix/…), so a plaintext secret var named `sops_*` is correctly detected as plaintext (no more silent skip / leak FN), while real SOPS trailers stay excluded. New test covers both directions. Resolved — thanks.

Issue #362, opened 2026-06-05
Go agent: watcher events channel never closed + unconditional blocking send → goroutine leak on project removal and wedged run()


jainal09:
## Summary
Two related defects in the per-project watcher make long-running guardians leak goroutines and can wedge the watcher loop:
1. `Watcher.Stop()` never closes the `events` channel, so the dedicated `forwardEvents` goroutine blocked on `<-pw.Events()` never sees a closed channel when a project is removed at runtime.
2. `handleEvent` sends on `events` unconditionally (no `select` on `done`), so a full buffered channel (slow/stopped consumer) blocks `run()` and defeats `Stop()`.

## Location
`envdrift-agent/internal/watcher/watcher.go:86` (`Stop`) and `:141` (blocking send)
Consumer: `envdrift-agent/internal/guardian/guardian.go` (`forwardEvents`, `pw.Stop()` on project removal)

## Details
```go
func (w *Watcher) Stop() {
    close(w.done)                 // events channel is NOT closed
    if err := w.fsWatcher.Close(); err != nil { ... }
}
```
```go
// handleEvent — unconditional send, no select on w.done
w.events <- FileEvent{Path: path, ModTime: info.ModTime(), Operation: event.Op.String()}
```
When `onRegistryChange` removes a project (`pw.Stop()`), the matching `forwardEvents` goroutine blocking on `<-pw.Events()` keeps `ok == true` and leaks until full guardian shutdown. Separately, if the buffered `events` channel (cap 100) fills, `run()` blocks on the send and can no longer observe `w.done`.

## Impact
In a daemon with churning project registrations, goroutines accumulate unboundedly; a stalled consumer can wedge the watcher so `Stop()` never returns.

## Suggested fix
- Close (or signal) the `events` channel in `Stop()` so `forwardEvents` exits.
- Make the send `select { case w.events <- ev: case <-w.done: return }`.

---
*Read-only deep review of `main` (Go agent). Distinct from the KNOWN watcher "new subdirectories not watched" item in #348.*

jainal09:
✅ Verified VALID on `main@211999e`. **Distinct from #348's `watcher.go:112`** (startup-only dir watching): this is a **goroutine leak** — `Stop()` never `close(w.events)` (`watcher.go:87-90`) so `forwardEvents` blocks forever on project removal, and `handleEvent` sends unconditionally with no `select` on `done` (`:141`) wedging `run()`. Same file, separate concurrency bug. Tracked in the Go-agent batch.

PR #424, opened 2026-06-07
fix(agent): harden watcher, lockcheck, registry, and stop reliability (#413 cluster J)

jainal09:
## Summary

Fixes the five adversarially-verified **Go agent** reliability findings (cluster J) from #413. Small, focused diff confined to `envdrift-agent/` plus one doc line.

## Findings fixed

1. **[HIGH] Recursive watch skips a project whose root dir name starts with `.`** — `envdrift-agent/internal/watcher/watcher.go:64-77`
   `filepath.Walk` visits the root first, so an unconditional hidden-dir `SkipDir` silently skipped the entire subtree of a registered dotted root (e.g. `~/.dotfiles`), watching nothing with no log line. Fix: apply the hidden-dir skip only when `path != filepath.Clean(dir)`, so the registered root is always watched and only *nested* hidden dirs are skipped.

2. **[MEDIUM] `stop` never stops the agent but reports success (exit 0)** — `envdrift-agent/internal/cmd/root.go:197-207`, `envdrift-agent/internal/daemon/daemon.go`
   `runStop` only printed status and returned nil. Added `daemon.Stop()` (`launchctl unload` / `systemctl --user stop` / `schtasks /end`, leaving the install unit in place) and wired `runStop` to call it, returning nonzero on failure. Help text and `docs/guides/agent-setup.md` now match behavior.

3. **[MEDIUM] Agent encrypts `.env` files open in an editor when `lsof` is absent** — `envdrift-agent/internal/lockcheck/lockcheck.go:28-42`
   `isFileOpenUnix` collapsed every `lsof` error into "not open", disabling the open-file guard on hosts without `lsof`. Fix: `errors.As` distinguishes `*exec.Error` (missing binary) and non-clean exits from a clean exit-1; missing/ambiguous now returns `true` (open/unknown) and warns once, so the guardian conservatively skips encrypting an in-use file.

4. **[MEDIUM] `RegistryWatcher.reload()/onChange` can fire after `Stop()` returns** — `envdrift-agent/internal/registry/registry.go:124-186`, `envdrift-agent/internal/guardian/guardian.go:280-334`
   The `time.AfterFunc(100ms, reload)` debounce wasn't stopped by `Stop()`, so a late reload re-added project watchers past guardian shutdown (leaking fsnotify watchers/FDs/goroutines). Fix: store the timer on the struct under `timerMu`, cancel it in `Stop()`, guard `reload()` behind a `stopped` check, and have `Guardian.onRegistryChange` early-return when the guardian ctx is Done.

5. **[LOW] `RegistryWatcher.Stop()` panics on a second/concurrent call** — `envdrift-agent/internal/registry/registry.go:124-127`
   Bare `close(rw.done)` panicked on a double call. Fix: wrap `Stop()` in `sync.Once`, mirroring `Watcher.stopOnce` from #362.

## Regression tests

Each was verified to **fail on the unfixed code and pass with the fix** (temporarily reverting each hunk):

- `internal/watcher/watcher_test.go`: `TestAddDirectoryDottedRootIsWatched` (dotted root is watched), `TestAddDirectorySkipsNestedHidden` (nested hidden dirs still skipped)
- `internal/lockcheck/lockcheck_test.go`: `TestIsFileOpenUnixMissingLsof` (missing-lsof path returns true) — drives a real cleared `PATH`
- `internal/registry/registry_test.go`: `TestRegistryWatcher_StopIsIdempotent` (double Stop, no panic), `TestRegistryWatcher_ReloadAfterStopDoesNotFireOnChange` (already-fired reload skips onChange post-Stop), `TestRegistryWatcher_NoOnChangeAfterStop` (end-to-end fsnotify path)
- `internal/daemon/daemon_test.go`: `TestStopDispatchesPerPlatform` (Stop runs real service-control, never the unsupported-platform no-op on a supported OS); `internal/cmd/root_test.go`: `TestRunStopMatchesBehavior` + `TestStopCmdHelpMatchesBehavior`

**Note on finding 2:** the stop *success* path on a genuinely running agent can't be reproduced in CI without installing a live launchd/systemd service. The tests cover the not-running no-op, the per-platform dispatch, and help/behavior alignment; the wiring to `daemon.Stop` is exercised deterministically.

## Gates

- `go build ./...`, `go vet ./...`, `go test ./...` — green (also `-race` on touched packages)
- `gofmt -l` clean; `golangci-lint` v2.12.2 — 0 issues
- `markdownlint-cli2` on `docs/guides/agent-setup.md` — 0 errors

Refs #413. Does not merge / no auto-merge.

<h3>Greptile Summary</h3>

This PR hardens five reliability bugs in the Go agent across watcher, lockcheck, registry, and stop-command subsystems. Each fix is narrow, documented against the original finding, and backed by regression tests that fail on the unfixed code.

- **Watcher**: `addRecursive` now exempts only the registered root from the hidden-dir skip, so a dotted root (e.g. `~/.dotfiles`) is watched correctly; `shouldWatchNewDir` independently blocks runtime-created hidden dirs from becoming their own exempt roots, closing the follow-up TOCTOU.
- **Registry**: `Stop()` is wrapped in `sync.Once` to prevent double-close panics; the debounce timer is stored on the struct under `timerMu` so `Stop()` can cancel a pending reload, and `reload()` checks the `stopped` flag before invoking `onChange`.
- **lockcheck/daemon/cmd**: `isFileOpenUnix` distinguishes a missing `lsof` binary from a clean exit-1 and returns conservatively; `runStop` is wired to `daemon.Stop()` with platform-specific `stopMacOS`/`stopLinux`/`stopWindows` implementations; `dispatch`/`dispatchBool` helpers deduplicate the `runtime.GOOS` switch.

Safe to merge; all five documented findings are addressed with targeted changes and regression tests that fail on the unfixed code.

Each fix is narrowly scoped to the documented failure mode. The watcher dotted-root skip, the double-Stop panic, the lsof missing-binary path, the post-shutdown debounce leak, and the runStop no-op are all correctly fixed and independently tested. The guardian's double-check pattern (fast RLock path + re-check inside Lock) correctly closes the TOCTOU for watcher creation after shutdown. No new correctness issues were introduced in the changed code.

No files require special attention; the most layered logic is in registry.go (debounce timer lifecycle) and guardian.go (double-check shutdown pattern), both of which have solid test coverage in this PR.

| Filename | Overview |
|----------|----------|
| envdrift-agent/internal/watcher/watcher.go | Fixes dotted-root SkipDir by introducing addRecursive/isHiddenName; adds shouldWatchNewDir to filter hidden dirs on runtime creates, closing the TOCTOU path where a newly-created hidden dir would be its own exempt root. |
| envdrift-agent/internal/registry/registry.go | Wraps Stop() in sync.Once; stores debounce timer on struct under timerMu so Stop() can cancel it; reload() checks stopped flag before invoking onChange. A narrow lock-release window remains between the stopped check in reload() and the onChange call, mitigated by the guardian's secondary ctxDone check. |
| envdrift-agent/internal/lockcheck/lockcheck.go | Distinguishes *exec.Error (missing lsof binary) from clean exit-1, returning true (conservative) for missing binary with a sync.Once-guarded warning; logic is correct and conservative for all error branches. |
| envdrift-agent/internal/daemon/daemon.go | Introduces dispatch/dispatchBool helpers to deduplicate GOOS switches; adds Stop() routing to stopMacOS/stopLinux/stopWindows. macOS and Linux paths are idempotent; Windows error-disambiguation logic is noted by a prior review thread. |
| envdrift-agent/internal/guardian/guardian.go | onRegistryChange gains a fast-path isShuttingDown() check and a guarded re-check inside g.mu.Lock(); the double-check pattern correctly closes the TOCTOU for watcher creation post-shutdown. |
| envdrift-agent/internal/cmd/root.go | runStop now actually stops the service via daemon.Stop(), gates the no-op only on IsInstalled() (not IsRunning()), and returns a non-nil error on failure. |
| envdrift-agent/internal/watcher/watcher_test.go | Adds TestAddDirectoryDottedRootIsWatched, TestAddDirectorySkipsNestedHidden, TestRuntimeCreatedHiddenDirIsNotWatched, and a shared drainEvents helper covering all three watcher cases. |
| envdrift-agent/internal/registry/registry_test.go | Adds StopIsIdempotent, ReloadAfterStopDoesNotFireOnChange, and NoOnChangeAfterStop tests; waitForArmedTimer eliminates fixed-sleep flakiness. |
| envdrift-agent/internal/lockcheck/lockcheck_test.go | TestIsFileOpenUnixMissingLsof clears PATH and verifies the missing-binary path returns true conservatively. |
| envdrift-agent/internal/daemon/daemon_test.go | Adds routing tests using stub handlers — no destructive service-control commands invoked on CI. |
| envdrift-agent/internal/cmd/root_test.go | TestRunStopNotInstalledIsNoOp and TestStopCmdHelpMatchesBehavior guard the stop regression. |
| docs/guides/agent-setup.md | One-line doc update aligning stop command description with new behavior. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant CLI as envdrift-agent stop
    participant Root as cmd/root.go runStop
    participant Daemon as daemon.Stop()
    participant Platform as stopMacOS/stopLinux/stopWindows

    CLI->>Root: runStop()
    Root->>Daemon: IsInstalled()?
    alt not installed
        Root-->>CLI: exit 0 "not installed"
    else installed
        Root->>Daemon: Stop()
        Daemon->>Platform: dispatch(stopMacOS, stopLinux, stopWindows)
        Platform-->>Daemon: nil or error
        alt success
            Daemon-->>Root: nil
            Root-->>CLI: exit 0 "Agent stopped"
        else failure
            Daemon-->>Root: error
            Root-->>CLI: exit 1 "failed to stop agent"
        end
    end

    participant RW as RegistryWatcher
    participant Timer as debounceTimer
    participant Guardian as Guardian.onRegistryChange

    RW->>RW: Stop() via sync.Once
    RW->>Timer: "timerMu.Lock - stopped=true - timer.Stop()"
    RW->>RW: close(done), fsWatcher.Close()
    Timer-->>RW: cancelled - reload() skips onChange
    Guardian->>Guardian: isShuttingDown() ctxDone() check
    Guardian->>Guardian: g.mu.Lock re-check ctxDone()
    Note over Guardian: early-return if ctx cancelled
```

1. `envdrift-agent/internal/watcher/watcher.go`, line 147-153 ([link](https://github.com/jainal09/envdrift/blob/4c1160e4233f83a9277ee0da1cb0e0f00e237e57/envdrift-agent/internal/watcher/watcher.go#L147-L153)) 

   <a href="#">P1</a> **Dynamic hidden subdir incorrectly watched after fix**

   `handleEvent` calls `AddDirectory(path)` when a new directory is created at runtime. With the PR's fix, `AddDirectory` sets `root = filepath.Clean(dir)`, and the Walk's very first entry has `path == root`, so the `path != root` guard does **not** skip the newly-created directory — even when its name is dotted (e.g. `/project/.git`, `/home/user/.cache`). Every sub-directory of that hidden dir that doesn't itself have a dotted name then gets added to the fsnotify watcher too.

   Before this fix, the first Walk entry for a dotted new subdir was caught by the unconditional `strings.HasPrefix(info.Name(), ".")` check. After the fix the same entry passes straight through to `w.fsWatcher.Add(path)`. The regression is not covered by `TestAddDirectorySkipsNestedHidden` because that test creates the `.git` dir *before* `AddDirectory(root)` is called — the static path through `Walk` — rather than exercising the dynamic-creation path through `handleEvent`.

2. `envdrift-agent/internal/registry/registry.go`, line 200-223 ([link](https://github.com/jainal09/envdrift/blob/4c1160e4233f83a9277ee0da1cb0e0f00e237e57/envdrift-agent/internal/registry/registry.go#L200-L223)) 

   <a href="#">P2</a> **Narrow TOCTOU between stopped check and onChange call**

   `reload()` checks `rw.stopped` under `timerMu`, then releases the lock before calling `rw.onChange(reg)`. If `Stop()` runs in that window — marking `stopped = true` — `onChange` still fires. The guardian's secondary `ctx.Done()` check in `onRegistryChange` provides defence-in-depth and closes the realistic risk in production, but the race within `reload()` itself is not fully eliminated: any caller that passes a different `onChange` without a context-cancellation guard would leak past shutdown.
</details>

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/go-agent-re..."](https://github.com/jainal09/envdrift/commit/29d5f4068713764b9a9e53250f1561ce5a620941) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36087226)</sub>


jainal09:
Bot review findings on the latest PR head (`4c1160e`):

1. `envdrift-agent/internal/watcher/watcher.go:76` still lets dynamically-created hidden directories become recursive watch roots. `handleEvent` calls `AddDirectory(path)` for directories created at runtime; with the new `path != root` exception, a created `.git` or `.cache` becomes the root of that walk and is not skipped. I verified this with a focused test that starts a recursive watcher, creates `.git`, then writes `.git/.env.local`; the hidden file event is emitted. Please keep the dotted-root exception only for explicitly registered project roots, or skip dotted directories before calling `AddDirectory` from `handleEvent`, and add a dynamic-creation regression test.

2. `envdrift-agent/internal/guardian/guardian.go:289` only checks shutdown before `LoadAllProjectConfigs`. If shutdown starts after that check but before `g.mu.Lock()`, `stopAllProjects()` can clear the project map first, and then `onRegistryChange` resumes and starts new project watchers after shutdown. I reproduced the interleaving by blocking config loading, canceling the context, calling `stopAllProjects`, then unblocking the load. Please re-check shutdown after config loading and while holding `g.mu` before starting any new watchers, or use a guarded stopping flag.

3. `envdrift-agent/internal/daemon/daemon_test.go:80` calls `Stop()` directly, which dispatches to the real host service manager (`launchctl unload`, `systemctl --user stop`, or `schtasks /end`). Running `go test ./internal/daemon` on a machine with an installed EnvDrift agent can stop the live service, and accepting `err == nil` means a silent no-op can still pass. Please inject or fake the command runner, or split command construction from execution so the test asserts the intended command without mutating host service state.

jainal09:
Bot review summary: reviewed latest head `4c1160e` for PR #424, including the Go agent watcher/guardian/daemon changes, linked #413 context, visible review comments, and CI status.

Posted 3 findings in this review pass: dynamically-created hidden directories can still become watched roots, shutdown can still race with registry reload and add watchers after `stopAllProjects()`, and a daemon unit test can invoke the real host service stop command. Visible checks are green, with Cloud Provider Tests skipped.

envdrift-agent/internal/daemon/daemon.go
```diff
@@ -42,6 +42,23 @@ func Uninstall() error {
 	}
 }
 
+// Stop stops the running agent service without removing its install unit, so a
```
jainal09, line 60: Resolved: Install/Uninstall/Stop and IsInstalled/IsRunning now route through shared dispatch/dispatchBool helpers, so Stop no longer duplicates the GOOS switch.

envdrift-agent/internal/guardian/guardian.go
```diff
@@ -279,6 +279,21 @@ func (g *Guardian) loadProjects(reg *registry.Registry) {
...
+	ctx := g.ctx
+	g.mu.RUnlock()
+	if ctx != nil {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
+	}
+
```
jainal09, line 289: Resolved: extracted loadEnabledConfigs / stopRemovedProjects / startNewProjects (and isShuttingDown/ctxDone), bringing onRegistryChange back below the complexity threshold.

envdrift-agent/internal/watcher/watcher.go
```diff
@@ -62,13 +62,19 @@ func (w *Watcher) AddDirectory(dir string) error {
...
+		// below the root.
+		root := filepath.Clean(dir)
 		return filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
 			if err != nil {
 				return nil // Skip inaccessible directories
 			}
 			if info.IsDir() {
-				if strings.HasPrefix(info.Name(), ".") && info.Name() != "." {
-					return filepath.SkipDir // Skip hidden directories
+				if path != root && strings.HasPrefix(info.Name(), ".") && info.Name() != "." {
```
jainal09: Resolved: the hidden-dir test is now an isNestedHiddenDir helper, flattening AddDirectory's conditional.

envdrift-agent/internal/watcher/watcher_test.go
```diff
@@ -215,6 +215,108 @@ func TestNewSubdirectoryIsWatched(t *testing.T) {
...
+				t.Fatalf("nested hidden dir %q should not be watched", hidden)
+			}
+			if event.Path == rootEnv {
+				gotRoot = true
+			}
+		case <-deadline:
+			t.Fatalf("no event for root .env; root should be watched")
+		}
+	}
+}
```
jainal09, line 315: Resolved: the select/for event loop is now in waitForEvent/assertNoEvent helpers, dropping this test below the complexity threshold.

envdrift-agent/internal/watcher/watcher_test.go
```diff
@@ -215,6 +215,108 @@ func TestNewSubdirectoryIsWatched(t *testing.T) {
...
+		select {
+		case event := <-w.Events():
+			if event.Path == envPath {
+				return // success: the dotted root is watched
+			}
+		case <-deadline:
+			t.Fatalf("no event for .env in dotted root %q (#413): root was SkipDir'd", root)
+		}
+	}
+}
```
jainal09, line 270: Resolved: same waitForEvent extraction simplifies this test below the threshold.

envdrift-agent/internal/watcher/watcher_test.go
```diff
@@ -215,6 +215,108 @@ func TestNewSubdirectoryIsWatched(t *testing.T) {
 	}
 }
 
+// TestAddDirectoryDottedRootIsWatched is the #413 regression: a recursive
```
jainal09, line 233: Resolved: extracting the shared waitForEvent/assertNoEvent helpers lowers the module's mean cyclomatic complexity back under threshold.

envdrift-agent/internal/daemon/daemon_test.go
```diff
@@ -57,6 +57,40 @@ func TestIsRunning(t *testing.T) {
...
+// and that any error is the wrapped "failed to stop agent" form — proving Stop
+// runs a real service-control command instead of silently succeeding like the
+// old runStop did. (The success path on a *running* agent can't be reproduced in
+// CI without installing a live service; see the PR note.)
+func TestStopDispatchesPerPlatform(t *testing.T) {
+	switch runtime.GOOS {
+	case "darwin", "linux", "windows":
+		// supported
+	default:
+		err := Stop()
```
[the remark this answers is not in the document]
jainal09: Fixed: the test no longer calls the real Stop() against a live service. TestDispatchRoutesPerPlatform/TestDispatchPropagatesHandlerError exercise the dispatch() helper with stub handlers, proving per-platform routing + unsupported-platform + error propagation without ever running launchctl/systemctl/schtasks.

envdrift-agent/internal/daemon/daemon_test.go
```diff
@@ -57,6 +57,40 @@ func TestIsRunning(t *testing.T) {
...
+		if err == nil || !strings.Contains(err.Error(), "unsupported platform") {
+			t.Errorf("Stop() on %s should report unsupported platform, got %v", runtime.GOOS, err)
+		}
+		return
+	}
+
+	err := Stop()
+	if err == nil {
```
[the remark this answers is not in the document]
jainal09: Fixed: replaced with stub-based dispatch tests that assert the routed handler is actually invoked (called == runtime.GOOS) and that a handler error propagates, so a silent no-op can no longer pass.

envdrift-agent/internal/watcher/watcher.go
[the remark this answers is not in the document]
jainal09: Fixed: AddDirectory now compares filepath.Clean(path) against the cleaned root (via isNestedHiddenDir), so a dotted root passed with a trailing slash or otherwise non-clean form is no longer SkipDir'd.

envdrift-agent/internal/watcher/watcher_test.go
```diff
@@ -215,6 +215,108 @@ func TestNewSubdirectoryIsWatched(t *testing.T) {
...
+	}
+	// ...while a .env in the nested hidden dir should NOT.
+	hiddenEnv := filepath.Join(hidden, ".env.local")
+	if err := os.WriteFile(hiddenEnv, []byte("Y=2\n"), 0o644); err != nil {
+		t.Fatalf("write hidden .env: %v", err)
+	}
+
+	gotRoot := false
+	deadline := time.After(2 * time.Second)
+	for !gotRoot {
```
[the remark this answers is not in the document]
jainal09: Fixed: after waitForEvent(rootEnv, forbidden=hiddenEnv) the test now drains for a 300ms grace window via assertNoEvent(hiddenEnv), so a lagging hidden-dir event can no longer slip past undetected.

envdrift-agent/internal/registry/registry_test.go
```diff
@@ -100,3 +102,127 @@ func TestRegistry_HasProject(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("NewRegistryWatcher: %v", err)
+	}
+	if err := rw.Start(); err != nil {
+		t.Fatalf("Start: %v", err)
+	}
+
+	// Arm the 100ms debounce timer, then Stop() inside the window.
+	writeRegistry(t, "/tmp/some/project")
+	time.Sleep(40 * time.Millisecond)
```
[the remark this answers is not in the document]
jainal09: Fixed: replaced the fixed 40ms sleep with waitForArmedTimer, which polls rw.debounceTimer under rw.timerMu until the timer is actually armed, so the Stop-inside-window path is always exercised.

envdrift-agent/internal/cmd/root.go
```diff
@@ -190,19 +190,25 @@ func runStart(cmd *cobra.Command, args []string) error {
...
+// next boot/login. It returns a non-nil error (non-zero exit) if stopping fails.
 func runStop(cmd *cobra.Command, args []string) error {
 	fmt.Println("Stopping envdrift-agent...")
-	// For daemon mode, we'd need to signal the process
-	// For now, just print status
-	if daemon.IsRunning() {
-		fmt.Println("Agent is running - use 'envdrift-agent uninstall' to stop")
-	} else {
+
+	if !daemon.IsRunning() {
```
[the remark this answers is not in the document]
jainal09: Fixed: runStop now keys the no-op on IsInstalled() rather than IsRunning(). A failed status probe can no longer be read as 'not running' to skip the stop and falsely report success; Stop is idempotent so attempting it when installed is safe.

envdrift-agent/internal/guardian/guardian.go
```diff
@@ -279,6 +279,21 @@ func (g *Guardian) loadProjects(reg *registry.Registry) {
...
 // onRegistryChange handles changes to the projects registry.
 func (g *Guardian) onRegistryChange(reg *registry.Registry) {
+	// If the guardian is already shutting down, do nothing: a late registry
+	// reload (e.g. a debounce timer that fired during Stop) must not re-create
+	// and start project watchers after stopAllProjects() cleared g.projects, or
+	// those fsnotify watchers/goroutines leak past shutdown (#413).
+	g.mu.RLock()
+	ctx := g.ctx
+	g.mu.RUnlock()
+	if ctx != nil {
```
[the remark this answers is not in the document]
jainal09: Fixed: the early ctx.Done() check is now just a fast path; onRegistryChange re-checks shutdown via ctxDone() *inside* g.mu.Lock (the same lock stopAllProjects holds) before re-adding watchers, so a Stop landing between the two can no longer be followed by watcher recreation.

jainal09:
Bot re-review update for latest head `8b3e916f`: the shutdown TOCTOU finding and destructive `Stop()` unit-test finding look addressed by this commit. I also ran the targeted Go packages locally: `go test ./internal/watcher ./internal/guardian ./internal/daemon ./internal/cmd` passed.

The dynamic hidden-directory watcher finding remains open. `handleEvent()` still calls `w.AddDirectory(path)` for any created directory at `envdrift-agent/internal/watcher/watcher.go:157-160`. For a runtime-created `.git`, that `.git` path becomes the `root` inside `AddDirectory()`, so `isNestedHiddenDir(filepath.Clean(path), root, info.Name())` returns false for the root and recursively adds watchers under it. The dotted-root exception still applies to dynamically-created hidden dirs, not just explicitly registered project roots.

Please skip dotted directories before calling `AddDirectory` from `handleEvent`, or pass enough context to `AddDirectory` to distinguish explicitly registered roots from discovered subdirectories.

envdrift-agent/internal/daemon/daemon_test.go
```diff
@@ -57,6 +57,88 @@ func TestIsRunning(t *testing.T) {
...
+		}
+	default:
+		if err == nil || !strings.Contains(err.Error(), "unsupported platform") {
+			t.Errorf("dispatch on %s should report unsupported platform, got %v", runtime.GOOS, err)
+		}
+		if called != "" {
+			t.Errorf("dispatch invoked %q handler on unsupported OS %s; want none", called, runtime.GOOS)
+		}
+	}
+}
```
jainal09, line 111: Resolved in e4c6156: extracted supportedGOOS + assertRoutedToCurrentOS so the dispatch and dispatchBool routing tests share the per-OS routing assertion instead of duplicating the GOOS switch.

envdrift-agent/internal/watcher/watcher.go
```diff
@@ -62,13 +62,22 @@ func (w *Watcher) AddDirectory(dir string) error {
 	dir = expandPath(dir)
 
 	if w.recursive {
+		// filepath.Walk visits the root first, so the hidden-dir skip must
```
jainal09: Resolved in e4c6156: replaced the 3-string-arg isNestedHiddenDir helper with a single isHiddenName(name) predicate and moved the recursive walk into an addRecursive method, cutting the string-argument ratio.

envdrift-agent/internal/watcher/watcher_test.go
```diff
@@ -215,6 +215,135 @@ func TestNewSubdirectoryIsWatched(t *testing.T) {
 	}
 }
 
+// waitForEvent drains w.Events until an event for wantPath arrives (returning
```
jainal09: Resolved in e4c6156: folded the per-test select/for event loops into one drainEvents helper, lowering the module's mean cyclomatic complexity below threshold.

envdrift-agent/internal/watcher/watcher_test.go
```diff
@@ -215,6 +215,135 @@ func TestNewSubdirectoryIsWatched(t *testing.T) {
...
+				}
+			}
+			if event.Path == wantPath {
+				return
+			}
+		case <-deadline:
+			t.Fatalf("no event for %q within %s", wantPath, timeout)
+		}
+	}
+}
```
jainal09: Resolved in e4c6156: waitForEvent/assertNoEvent are replaced by a single drainEvents(t, w, eventWait{...}) driven by a struct, so no function exceeds the argument limit.

jainal09:
Bot re-review update for latest head `e4c61568`: the new commit is a CodeScene-oriented refactor and the remaining watcher finding is still open.

`handleEvent()` still does:

```go
if w.recursive && event.Op&fsnotify.Create != 0 {
    if info, err := os.Stat(path); err == nil && info.IsDir() {
        if err := w.AddDirectory(path); err != nil { ... }
    }
}
```

So a runtime-created `.git`/`.cache` directory is still passed as the root of `AddDirectory()`, preserving the dotted-root exception for that discovered hidden directory. `go test ./internal/watcher` passed locally, but there is still no dynamic hidden-directory regression covering this case.

envdrift-agent/internal/watcher/watcher.go
```diff
@@ -62,13 +62,22 @@ func (w *Watcher) AddDirectory(dir string) error {
...
 			if err != nil {
 				return nil // Skip inaccessible directories
 			}
 			if info.IsDir() {
-				if strings.HasPrefix(info.Name(), ".") && info.Name() != "." {
-					return filepath.SkipDir // Skip hidden directories
+				// Compare the cleaned path against the cleaned root so a dotted
+				// root passed with a trailing slash or otherwise non-clean form
+				// (e.g. "~/.dotfiles/") still matches and isn't SkipDir'd.
+				if isNestedHiddenDir(filepath.Clean(path), root, info.Name()) {
```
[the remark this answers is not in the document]
jainal09: Valid bug, fixed in ea0f707: handleEvent now skips re-adding a newly-created directory when its base name is hidden (isHiddenName), so a runtime-created .cache/.git/etc. is no longer watched. The explicit-AddDirectory root exemption stays so a registered ~/.dotfiles is still watched. Added TestRuntimeCreatedHiddenDirIsNotWatched.

envdrift-agent/internal/cmd/root.go
```diff
@@ -190,19 +190,31 @@ func runStart(cmd *cobra.Command, args []string) error {
...
 func runStop(cmd *cobra.Command, args []string) error {
 	fmt.Println("Stopping envdrift-agent...")
-	// For daemon mode, we'd need to signal the process
-	// For now, just print status
-	if daemon.IsRunning() {
-		fmt.Println("Agent is running - use 'envdrift-agent uninstall' to stop")
-	} else {
-		fmt.Println("Agent is not running")
+
+	if !daemon.IsInstalled() {
```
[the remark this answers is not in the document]
jainal09: Valid, fixed in ea0f707: stopWindows now treats a not-running task as a successful no-op (returns nil when isRunningWindows() is false) instead of letting 'schtasks /end' fail on an installed-but-idle task. This matches launchctl unload / systemctl --user stop of an inactive unit, so runStop's IsInstalled() gate is safe across all three platforms.

jainal09:
Bot re-review update for latest head `ea0f707f`: the remaining dynamic hidden-directory watcher finding looks addressed.

`handleEvent()` now guards runtime-created directories with `!isHiddenName(filepath.Base(path))` before calling `AddDirectory(path)`, so a created `.git`/`.cache` no longer becomes its own root-exempt recursive watch root. The PR also adds `TestRuntimeCreatedHiddenDirIsNotWatched` for this case.

Local verification on the latest head: `go test ./internal/watcher ./internal/daemon ./internal/cmd` passed.

envdrift-agent/internal/watcher/watcher.go
```diff
@@ -138,7 +155,12 @@ func (w *Watcher) handleEvent(event fsnotify.Event) {
 	// If a new directory was created, start watching it (recursively) so that
 	// .env files created beneath it later are not missed (#348 G2). Do this
 	// before the pattern filter, since a directory name won't match .env*.
-	if w.recursive && event.Op&fsnotify.Create != 0 {
+	// Skip hidden directories: a runtime-created .git/.venv/etc. must NOT be
+	// watched. AddDirectory exempts its own (possibly dotted) root so an
+	// explicitly-registered ~/.dotfiles is watched, but that exemption would
+	// also make a runtime-created hidden dir its own root and watch it, so we
+	// filter it out here before re-entering AddDirectory.
+	if w.recursive && event.Op&fsnotify.Create != 0 && !isHiddenName(filepath.Base(path)) {
```
jainal09: Resolved in 6a585ba: extracted the recursive+create+non-hidden test into a shouldWatchNewDir method, so handleEvent's branch is now a single predicate (under the complex-conditional threshold).

jainal09:
Bot CI/status re-check for latest head `6a585ba3`: the watcher behavior finding remains addressed by the hidden-name guard and regression test, but `CodeScene Code Health Review (main)` is still failing. Several integration and Windows checks were still pending when I polled.

jainal09:
Accepting this advisory with rationale. `watcher.go` is a filesystem watcher — its arguments are `dir`/`path`/`name` strings, which is exactly what Go's `filepath`/`os`/`fsnotify` stdlib takes (`filepath.Walk(dir string, ...)`, `os.Stat(path string)`, etc.). Introducing newtypes (e.g. `type DirPath string`) solely to drop below CodeScene's 39% string-argument threshold would force a cast at every stdlib boundary and add indirection with no type-safety or correctness benefit for path handling. The substantive reliability fixes in this PR (dotted-root watch, lsof guard, registry-watcher shutdown leak + double-Stop panic, agent stop) are all covered by Go tests. Leaving the path-string arguments idiomatic.

jainal09:
Bot re-review update for latest head `224fc0cc`: the watcher include/exclude matcher refactor appears behavior-preserving from this review pass. The dynamic hidden-directory fix remains in place, and `go test ./internal/watcher` passed locally on this head.

GitHub checks for this new head were still pending when I polled.

jainal09:
Addressed: (1) **greptile (stopWindows)** — fixed; `/end` now runs unconditionally and the `isRunningWindows()` probe is only consulted *after* an error to tell already-stopped from a real failure, so a transient probe failure can't falsely report success (matches the runStop contract + stopMacOS/stopLinux). (2) **CodeScene watcher_test duplication** — extracted `assertBaseMatchCases` shared by both table tests. (3) **CodeScene String Heavy Function Arguments (watcher.go)** — accepting with rationale: this is a filesystem watcher whose args are `dir`/`path`/`name` strings, exactly what Go's `filepath`/`os`/`fsnotify` stdlib takes; newtypes solely to clear the 39% threshold would add casts at every stdlib boundary with no safety gain.

jainal09:
Bot review summary for latest head 591b2e39:

I re-reviewed the latest stopWindows/test refactor. It keeps `/end` as the primary stop path and only probes after `/end` errors, which addresses the earlier stop reliability concern. I do not have a new code finding on this latest commit.

Current CI state from `gh pr checks`: all visible GitHub Actions, CodeQL, Codecov, commitlint, and semantic-pr checks pass; CodeScene Code Health Review (main) is still failing. Review state: previous code findings appear addressed; PR still needs the failing CodeScene check resolved or explicitly triaged.

jainal09:
Bot review summary for latest head `591b2e39`:

I re-reviewed the Windows stop refactor, watcher/registry/lockcheck changes, and current checks. One Windows-only stop correctness issue remains.

Finding:
- Medium: `stopWindows()` can still report failure when the scheduled task is merely idle. After `schtasks /end` fails, it calls `isRunningWindows()`, but that probe checks `tasklist` for the current executable name. During `envdrift-agent stop`, the stop command itself has the same image name as the daemon, so the probe can return true even when the scheduled task has no running instance. The result is a false nonzero stop for an installed-but-idle task. The stop path should distinguish scheduled-task running state from the current CLI process, or parse/query task state directly.

The previous watcher/shutdown/destructive-test findings look addressed. Current visible checks are not failing.

Issue #356, opened 2026-06-05
Sync never converges when vault value is quoted (read_key strips quotes, _fetch_vault_secret does not)


jainal09:
## Summary
Sync never converges (idempotency is broken) when a vault secret value is stored quoted. `EnvKeysFile.read_key` strips surrounding quotes from the locally-stored value, but `_fetch_vault_secret` returns the vault value verbatim (quotes intact), so the fetched value and the read-back local value never compare equal.

## Location
`src/envdrift/sync/engine.py:251-264` (`_fetch_vault_secret`)
`src/envdrift/sync/operations.py:44-49` (`EnvKeysFile.read_key`)

## Details
```python
# engine.py _fetch_vault_secret — strips only the DOTENV_PRIVATE_KEY_*= prefix, keeps quotes
match = re.match(r"^DOTENV_PRIVATE_KEY_[A-Za-z0-9_]+=(.+)$", value)
if match:
    value = match.group(1)
return value
```
```python
# operations.py read_key — strips surrounding quotes from the LOCAL value only
if (value.startswith('"') and value.endswith('"')) or (value.startswith("'") and value.endswith("'")):
    value = value[1:-1]
```
If the vault stores `"abc123"` (or the full line `DOTENV_PRIVATE_KEY_PROD="abc123"`), the engine writes the quoted form locally, then on the next run reads it back as `abc123`, which never equals the fetched `"abc123"`.

## Impact
- `--verify` reports a false **"Value mismatch"** on every run.
- `--force` rewrites the file and creates a new timestamped `.backup.<ts>` **every run** → unbounded backup accumulation.
- Interactive mode re-prompts forever.

## Suggested fix
Normalize quotes consistently on both sides — strip surrounding quotes in `_fetch_vault_secret` the same way `read_key` does (or stop stripping in `read_key` and compare raw).

---
*Read-only deep review of `main`.*

PR #433, opened 2026-06-08
fix(sync): unify lock --verify-vault parser; activate already-decrypted profiles

jainal09:
Closes the final two #413 reliability findings (14, 23), both in `cli_commands/sync.py`.

## Finding 14 — `lock --verify-vault` false KEY MISMATCH

The verify path parsed the vault value with a bespoke inline parser:
```python
if "=" in vault_value and vault_value.startswith("DOTENV_PRIVATE_KEY"):
    vault_key = vault_value.split("=", 1)[1]
else:
    vault_key = vault_value
```
No whitespace strip, no quote strip, prefix-split only on a literal `DOTENV_PRIVATE_KEY` prefix. But `local_key` comes from `read_key`, which **does** normalize (strip whitespace + quotes). So a vault value stored quoted (`"abc"`) or prefixed always reported a **false KEY MISMATCH**.

**Fix:** extract the engine's canonical normalization into a shared `normalize_vault_key_value()` helper and use it in **both** `_fetch_vault_secret` and `lock --verify-vault`, so they parse identically. Verify also flags a genuine environment-suffix mismatch (`DOTENV_PRIVATE_KEY_STAGING` vs a production mapping).

## Finding 23 — `pull --profile` activation not idempotent

`activate_to` (copy the decrypted file to the active path) ran **only** after a fresh decrypt. A profile file that was already decrypted — committed decrypted, or from a second `pull` — hit the `skipped (not encrypted)` branch and was never activated, contradicting the docstring/`pull.md` promise.

**Fix:** extract `_maybe_activate_profile()` and call it from **both** the post-decrypt path and the already-decrypted skip path.

## Tests

- `TestNormalizeVaultKeyValue` — whitespace/quotes/prefix/suffix, quote-before-prefix order
- `lock --verify-vault` with a quoted vault value matching the local key → reports a match, not MISMATCH
- `pull --profile` on an already-decrypted file → still activated (idempotent), never decrypted

`engine.py` 94%, `cli_commands/sync.py` 88% line coverage; existing engine/pull/lock suites green (refactor preserves behavior).

## Summary by CodeRabbit

* **New Features**
  * Enhanced `pull --profile` to activate plaintext environment files without requiring decryption.

* **Bug Fixes**
  * Improved vault secret normalization to correctly match quoted vault values with local keys in `lock --verify-vault`.
  * Separated and clarified error reporting for activation failures versus decryption failures.
  * Added dedicated tracking for activation errors in the command summary.

<h3>Greptile Summary</h3>

This PR resolves two reliability findings from #413: it unifies the vault-value parser between `lock --verify-vault` and the sync engine via a shared `normalize_vault_key_value` helper (fixing false KEY MISMATCH on quoted/prefixed vault values), and it makes `pull --profile` idempotent by calling the new `_maybe_activate_profile` helper from both the post-decrypt path and the already-plaintext skip path.

- **Finding 14 (verify-vault parser):** `normalize_vault_key_value` strips whitespace → outer quotes → `DOTENV_PRIVATE_KEY_<SUFFIX>=` prefix → inner whitespace/quotes, converging with `EnvKeysFile.read_key`. Both verify-vault and the engine now use it, eliminating the bespoke inline parser.
- **Finding 23 (pull --profile idempotency):** `_maybe_activate_profile` is extracted and called from the \"skipped (not encrypted)\" branch in addition to the post-decrypt branch, so a previously-decrypted profile file is still copied to its `activate_to` destination.
- **Error accounting:** `activation_error_count` is tracked separately from `error_count`, so activation failures on plaintext files never trigger the misleading \"could not be decrypted\" warning.

Safe to merge — both fixes are narrowly scoped and well-tested; the only rough edge is a non-blocking wording issue in one error message.

Both changes are surgical: `normalize_vault_key_value` is a straightforward extraction of existing logic with an added inner-quote strip that matches `read_key`'s behaviour, and `_maybe_activate_profile` is called in one new location with correct guard conditions. The directory-traversal check, idempotency semantics, and error accounting are all handled correctly. The one observation — that the verify-vault MISMATCH message conflates a key-content mismatch with an environment-label mismatch, making the --sync-keys suggestion misleading in the suffix-only case — is a UX nit that does not affect correctness or data integrity.

No files require special attention; the new `suffix_ok` branch in `sync.py` is the one area worth a second read for the error-message wording.

| Filename | Overview |
|----------|----------|
| src/envdrift/sync/engine.py | Extracts `normalize_vault_key_value` from the inline normalization in `_fetch_vault_secret`; behavior-preserving refactor plus a new inner-quote/whitespace strip for the post-prefix value (matching `read_key`). Regex and helper are correct; exported function makes the engine's logic reusable. |
| src/envdrift/cli_commands/sync.py | Adds `_maybe_activate_profile` helper (idempotent activation for both post-decrypt and already-plaintext paths) and `activation_error_count` tracking; the verify-vault parser is unified via `normalize_vault_key_value`. One misleading error message when the vault suffix doesn't match the target environment. |
| tests/unit/test_sync_engine.py | Adds `TestNormalizeVaultKeyValue` covering whitespace, single/double quotes, prefix+suffix extraction, quote-before-prefix ordering, inner-quote strip, and a convergence test against `EnvKeysFile.read_key`. Thorough for the new helper. |
| tests/unit/coverage/test_cov_cli_commands_sync.py | Adds two new tests: idempotent activation of an already-decrypted profile, and activation-failure message accuracy (ensures "could not be activated" not "could not be decrypted"). Both cover the idempotency fix cleanly. |
| tests/unit/test_cli.py | Adds integration test for `lock --verify-vault` with a quoted vault value, verifying no false KEY MISMATCH. Updates existing activation-failure assertion to use whitespace-normalized output and checks the new "could not be activated" warning. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant CLI as pull / lock CLI
    participant MAA as _maybe_activate_profile
    participant NVK as normalize_vault_key_value
    participant ENG as SyncEngine._fetch_vault_secret

    Note over CLI: pull --profile path
    CLI->>CLI: detect env_file encrypted?
    alt already plaintext (skip path) [NEW]
        CLI->>MAA: _maybe_activate_profile(mapping, env_file, profile)
        MAA-->>CLI: activated / error / noop
    else freshly decrypted
        CLI->>MAA: _maybe_activate_profile(mapping, env_file, profile)
        MAA-->>CLI: activated / error / noop
    end

    Note over CLI: lock --verify-vault path
    CLI->>NVK: normalize_vault_key_value(vault_secret.value)
    NVK-->>CLI: (vault_key, vault_suffix)
    CLI->>CLI: suffix_ok check
    CLI->>CLI: "if suffix_ok and local_key == vault_key → MATCH else MISMATCH"

    Note over ENG: sync engine path
    ENG->>NVK: normalize_vault_key_value(secret.value)
    NVK-->>ENG: (value, suffix)
    ENG->>ENG: "if suffix != effective_env → raise VaultError"
```

<sub>Reviews (5): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/sync-cli-co..."](https://github.com/jainal09/envdrift/commit/c0414439a4fb1c633097980e3faacacdd0fa3ba6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36300338)</sub>


src/envdrift/cli_commands/sync.py
```diff
@@ -818,6 +852,14 @@ def prompt_callback(msg: str) -> bool:
...
             console.print(f"  [dim]=[/dim] {env_file} [dim]- skipped (not encrypted)[/dim]")
             skipped_count += 1
+            # An already-decrypted profile file must still be activated so
+            # `pull --profile` is idempotent (#413): a file committed decrypted
+            # or decrypted by an earlier run isn't silently left un-activated.
+            outcome = _maybe_activate_profile(mapping, env_file, profile)
+            if outcome == "activated":
+                activated_count += 1
+            elif outcome == "error":
+                error_count += 1
```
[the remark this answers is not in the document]
jainal09: Fixed in cbb6179 — activation failures now use a dedicated activation_error_count and print 'Some profile files could not be activated' instead of the misleading 'could not be decrypted'. Either error still exits 1. Tests cover both the post-decrypt and already-decrypted activation-failure paths.

src/envdrift/sync/engine.py
```diff
@@ -27,6 +27,37 @@
...
+    match the target environment (the engine raises; verify reports a mismatch).
+    """
+    value = raw.strip()
+    if len(value) >= 2 and (
+        (value[0] == '"' and value[-1] == '"') or (value[0] == "'" and value[-1] == "'")
+    ):
+        value = value[1:-1]
+    match = _DOTENV_PRIVATE_KEY_RE.match(value)
+    if match:
+        return match.group(2), match.group(1)
```
jainal09: 🤖 **envdrift review-bot** · medium · bug


The new helper's docstring claims it "mirrors how `EnvKeysFile.read_key` normalizes the locally-stored value, so a vault value and the local value converge," and the PR body asserts the unified parser eliminates false KEY MISMATCH for quoted/prefixed vault values. That is only true for the OUTER-quote case. The two parsers diverge when a `DOTENV_PRIVATE_KEY_<SUFFIX>=` prefix is present AND the value after the `=` is itself quoted or whitespace-padded.

`normalize_vault_key_value` does: strip whole string → dequote one OUTER layer → regex `^DOTENV_PRIVATE_KEY_([A-Za-z0-9_]+)=(.+)$` → return `group(2)` **verbatim** (no further strip/dequote). `EnvKeysFile.read_key` (operations.py:52-60) does the opposite order on the post-`=` part: it captures `(.+)` after the key, then `value = match.group(1).strip()`, then dequotes one layer.

Consequence (reproduced in a /tmp scratch script):
- Vault stores `DOTENV_PRIVATE_KEY_PROD="abc123"` → normalize returns `'"abc123"'` (quotes intact); read_key on the same line returns `'abc123'`. verify-vault → false KEY MISMATCH.
- Vault stores `DOTENV_PRIVATE_KEY_PROD=  abc123` → normalize returns `'  abc123'`; read_key returns `'abc123'`. Mismatch again.

This is also a real correctness problem in the sync engine (not just verify): `_fetch_vault_secret` returns this un-stripped value, and `sync_all` both compares it to `read_key`'s normalized `local_value` (perpetual UPDATE churn / never-converges) and writes it back via `write_key(effective_key_name, vault_value, ...)` (engine.py:193,240) — installing a key with literal quotes/leading whitespace, i.e. corrupted key material.

This specific inner-value gap is pre-existing (the old inline _fetch_vault_secret also returned `match.group(2)` raw), so it's not a regression introduced here — but per CLAUDE.md "there are no pre-existing issues," and the PR explicitly raises the claim that the helper now matches read_key. To actually deliver that convergence (and let verify-vault stop false-mismatching the documented #356 scenario), the post-prefix `group(2)` should be `.strip()`-ed and dequoted exactly as read_key does. Suggested direction: apply the same strip+single-quote-strip to the returned key material after prefix removal (and add a TestNormalizeVaultKeyValue case for an inner-quoted prefixed value to lock it in).

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit cbb6179


One medium-severity correctness gap remains: the new unified `normalize_vault_key_value` parser does not fully converge with `EnvKeysFile.read_key` for prefixed values whose inner value is quoted or whitespace-padded, so the #356 false-mismatch / write-churn class isn't fully eliminated for that subcase.

**Findings (1):**
- [medium] normalize_vault_key_value does NOT normalize identically to read_key for prefixed values (inner quotes/whitespace not stripped) — residual false KEY MISMATCH / key corruption — `src/envdrift/sync/engine.py:58` (inline)

src/envdrift/sync/engine.py
jainal09: Sharp catch — fixed in 8ecd373. normalize_vault_key_value now strips+dequotes the inner value after the DOTENV_PRIVATE_KEY_<SUFFIX>= prefix, exactly as read_key treats the post-= part. Added tests for inner double/single-quote and whitespace cases, plus a direct assertion that the helper and EnvKeysFile.read_key agree on DOTENV_PRIVATE_KEY_PRODUCTION="abc123". This closes the engine write-back of quoted/whitespace key material too.

tests/unit/test_cli.py
```diff
@@ -3618,6 +3624,59 @@ def test_lock_hook_check_errors_exit(self, monkeypatch, tmp_path: Path):
...
+        config_file.write_text('[vault]\nprovider = "aws"\n')
+
+        result = runner.invoke(app, ["lock", "--verify-vault", "-c", str(config_file), "-p", "aws"])
+
+        # Collapse whitespace so a Rich soft-wrap of the long tmp path (CI runs
+        # at a narrow width) doesn't split "keys match vault" across lines.
+        normalized = " ".join(result.output.split())
+        # The quoted vault value normalizes to the same bare key as local.
+        assert "keys match vault" in normalized, result.output
+        assert "KEY MISMATCH" not in normalized
```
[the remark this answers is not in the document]
jainal09: Done — added assert result.exit_code == 0 before the output checks (the keys-match path proceeds to a clean lock exit).

jainal09:
🤖 **envdrift review-bot** — automated re-review of latest head `fec9c99a`


Reviewed the PR description and cited #413/#356 context, the previous bot comment on `cbb6179`, the current diff, and the latest parser fix commit `8ecd373`. Focused tests passed locally:

`uv run --extra dev pytest tests/unit/test_sync_engine.py::TestNormalizeVaultKeyValue tests/unit/coverage/test_cov_cli_commands_sync.py::TestPullCommand::test_pull_activates_already_decrypted_profile_file tests/unit/coverage/test_cov_cli_commands_sync.py::TestPullCommand::test_pull_already_decrypted_activation_error_is_not_decrypt_error tests/unit/test_cli.py::TestLockCommand::test_verify_vault_quoted_value_matches_no_false_mismatch -q` → `14 passed`

The earlier quoted/whitespace prefixed-value finding is fixed for cases like `DOTENV_PRIVATE_KEY_PROD="abc123"`, `DOTENV_PRIVATE_KEY_PROD='abc123'`, and `DOTENV_PRIVATE_KEY_PROD=  abc123  `; those now normalize to `abc123` and match `EnvKeysFile.read_key()`.

**Findings (2):**

- [high] Hyphenated environment suffixes are still not parsed as prefixed key values, which can write nested key lines.

`src/envdrift/sync/engine.py:32` still uses `^DOTENV_PRIVATE_KEY_([A-Za-z0-9_]+)=(.+)$`, which rejects suffixes containing `-`. The rest of the sync path can derive key names directly from environment labels, so `.env.qa-blue` / `environment="qa-blue"` leads to `DOTENV_PRIVATE_KEY_QA-BLUE`. At head `fec9c99a`, `normalize_vault_key_value('DOTENV_PRIVATE_KEY_QA-BLUE=abc123')` returns `('DOTENV_PRIVATE_KEY_QA-BLUE=abc123', None)`, while `EnvKeysFile.read_key('DOTENV_PRIVATE_KEY_QA-BLUE')` returns `abc123`.

I also reproduced the write impact: syncing a vault value `DOTENV_PRIVATE_KEY_QA-BLUE=abc123` for environment `qa-blue` writes `.env.keys` as `DOTENV_PRIVATE_KEY_QA-BLUE=DOTENV_PRIVATE_KEY_QA-BLUE=abc123`. In verify mode, the same local/vault key material false-fails with `KEY MISMATCH`. Please allow the suffix charset already produced by environment labels, e.g. `[A-Za-z0-9_-]+`, and add a regression test for the hyphenated case.

- [medium] Suffix-mismatch remediation still tells users to run a command that rejects the same mismatch.

In `src/envdrift/cli_commands/sync.py:1355`, `lock --verify-vault` uses the generic `run 'envdrift lock --sync-keys' to fix` error for both content mismatches and environment-label mismatches. I reproduced a local `PRODUCTION` key with a vault value `DOTENV_PRIVATE_KEY_STAGING=abc123`: `lock --verify-vault` exits 1 and suggests `--sync-keys`, then `lock --sync-keys --force` exits 1 with `vault key labeled for environment STAGING cannot be installed as environment PRODUCTION`. Please split the `suffix_ok is false` branch and tell the user to fix or repush the vault secret under the correct environment label instead of suggesting `--sync-keys`.

Current live checks: integration matrix and CI coverage were still in progress at the last poll.

src/envdrift/sync/engine.py
jainal09: 🤖 **envdrift review-bot** — resolved ✅

Verified at `8ecd373`: `normalize_vault_key_value` now strips/dequotes the inner value after the `DOTENV_PRIVATE_KEY_<SUFFIX>` prefix, so its output matches `read_key` across quoting variants — the spurious vault-vs-local mismatch is gone, and the new equivalence test is load-bearing (fails without the fix). Resolved, thanks.

PR #442, opened 2026-06-09
ci: run the non-integration suite on macOS and Windows

jainal09:
## What

Adds a **Cross-Platform Tests** workflow that runs `pytest -m "not integration"`
on a 3-OS × 2-Python matrix:

| | py3.11 | py3.14 |
|---|---|---|
| ubuntu-latest | ✅ | ✅ |
| macos-latest | ✅ | ✅ |
| windows-latest | ✅ | ✅ |

`fail-fast: false` so one red OS doesn't cancel the others — we want to see
*every* platform's result in one run.

## Why

Actions runners are free for public repos, and `envdrift` is path- and
filesystem-heavy (`.env.keys`, symlinks, `chmod`, tempfiles, default text
encodings). Those are exactly the side effects that stay invisible until a
macOS or Windows user hits them. Validating the pure-Python suite only on
Linux leaves that whole class of bug uncaught.

## Scope boundary (deliberate)

- **Container-backed integration** (`aws` / `vault` / `azure` markers, served by
  docker-compose in `integration-tests.yml`) stays **Linux-only** — macOS
  runners have no Docker daemon and Windows runners only run Windows
  containers, so a Linux service stack can't start there.
- **Binary-backed integration** (the real dotenvx/sops seam) on macOS/Windows
  is a **planned follow-up**. The binaries already ship `windows_amd64` /
  `darwin_*` builds in `src/envdrift/constants.json`, so the install step just
  needs a per-OS branch once this matrix is green — and the upcoming
  adversarial harness will carry POSIX-only skip-guards (`/dev/null`, `chmod`,
  symlinks) so it runs cleanly on Windows.

## Rollout

Intentionally **not a required status check yet** — it's expected to surface
pre-existing cross-platform bugs first (those become tracked findings + real
regression tests). Promote to required in branch protection once the matrix is
reliably green.

Path-filtered to Python changes (`src/**`, `tests/**`, `pyproject.toml`,
`uv.lock`, the workflow itself) so docs-only and vscode/agent-only PRs don't
spend matrix minutes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed backslash escaping issues in environment file paths on Windows to prevent generated modules from failing to compile.
  * Improved UTF-8 encoding handling across tests to ensure consistent behavior on Windows.

* **Tests**
  * Enhanced cross-platform test reliability with platform-specific test skips for Windows edge cases.
  * Added GitHub Actions workflow for automated testing across Windows, macOS, and Ubuntu with Python 3.11 and 3.14.

<h3>Greptile Summary</h3>

This PR adds a 3-OS × 2-Python (3.11/3.14) cross-platform CI matrix for the non-integration test suite, and pairs it with the Windows compatibility fixes that the new matrix immediately requires.

- **New workflow** (`cross-platform.yml`): runs `pytest -m \"not integration\"` on ubuntu/macos/windows with `fail-fast: false`, uv cache keyed per-OS/Python, and `shell: bash` for uniform marker quoting; deliberately not a required check yet.
- **Source fixes**: `init_cmd.py` doubles backslashes in the generated docstring to prevent `\\U`/`\\x` compile errors from Windows paths; `config.py` adds `encoding=\"utf-8\"` to `create_example_config` so the em-dash in `EXAMPLE_CONFIG` survives on Windows (default cp1252).
- **Test fixes**: `write_text` calls with non-ASCII content gain `encoding=\"utf-8\"`; TOML path values switch to single-quoted literal strings so the TOML parser ignores backslashes; POSIX-only tests get `win32` skip guards; `Path.home()` is patched directly instead of via the `HOME` env var.

Safe to merge — the workflow is opt-in (not a required check), and the source/test fixes are narrowly scoped Windows encoding and escaping corrections.

The production code changes are minimal and address real compile-time bugs (generated docstring backslash escaping, UTF-8 config write). The test changes are all additive guards or encoding fixes with no logic removal. The CI workflow is not required, so a flaky run cannot block main.

| Filename | Overview |
|----------|----------|
| .github/workflows/cross-platform.yml | New 3-OS × 2-Python matrix workflow for non-integration tests; correctly sets fail-fast: false, adds uv cache keyed per-OS/Python, and uses shell: bash for uniform marker quoting. |
| src/envdrift/cli_commands/init_cmd.py | Adds env_file_doc with backslash doubling so Windows paths inside the generated docstring don't produce invalid escape sequences at compile time. |
| src/envdrift/config.py | Adds encoding="utf-8" to create_example_config's write_text call, fixing a UnicodeDecodeError on Windows where the default cp1252 encoding would mangle the em-dash in EXAMPLE_CONFIG. |
| tests/unit/coverage/test_cov_cli_commands_sync.py | Switches TOML path values from double-quoted basic strings to single-quoted literal strings so Windows backslashes aren't treated as escapes by the TOML parser. |
| tests/unit/test_config_deferred_validation.py | Adds encoding="utf-8" to TOML config writes and a new regression test verifying create_example_config produces a file load_config can read end-to-end. |

</details>

<sub>Reviews (4): Last reviewed commit: ["fix: make the non-integration test suite..."](https://github.com/jainal09/envdrift/commit/7b22315e677022d193ffd796d767bd4117443d14) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36363998)</sub>

### DIFF
--- a/.github/workflows/cross-platform.yml
+++ b/.github/workflows/cross-platform.yml
@@ -69,6 +69,11 @@ jobs:
 
       - name: Install uv
         uses: astral-sh/setup-uv@v8.2.0
+        with:
+          # Persist uv's package cache across runs; keyed per-OS/Python below so
+          # the 6-job matrix doesn't re-download wheels every time.
+          enable-cache: true
+          cache-suffix: ${{ matrix.os }}-py${{ matrix.python-version }}
 
       - name: Install dependencies
         run: uv sync --all-extras
--- a/.github/workflows/cross-platform.yml
+++ b/.github/workflows/cross-platform.yml
@@ -1,0 +1,81 @@
+name: Cross-Platform Tests
+
+# Runs the NON-integration test suite (pure-Python: path handling, parsing,
+# encoding, CLI rendering) on Linux, macOS and Windows so OS-specific side
+# effects — path separators, line endings, tempfile/permission semantics,
+# default encodings — surface in CI instead of on a user's machine.
+#
+# Scope boundary: container-backed integration tests (the `aws` / `vault` /
+# `azure` markers, served by docker-compose in integration-tests.yml) stay
+# Linux-only — GitHub's macOS runners have no Docker daemon and Windows runners
+# only run Windows containers, so a Linux service stack cannot start there.
+# Binary-backed integration (real dotenvx/sops seam) on macOS/Windows is a
+# planned follow-up once this matrix is green; it needs the binaries installed
+# per-OS from src/envdrift/constants.json.
+#
+# This workflow is intentionally NOT a required status check yet: it is expected
+# to surface pre-existing cross-platform bugs first. Promote it to required in
+# branch protection once the matrix is reliably green.
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - "src/**"
+      - "tests/**"
+      - "pyproject.toml"
+      - "uv.lock"
+      - ".github/workflows/cross-platform.yml"
+  push:
+    branches: [main]
+    paths:
+      - "src/**"
+      - "tests/**"
+      - "pyproject.toml"
+      - "uv.lock"
+      - ".github/workflows/cross-platform.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: cross-platform-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  tests:
+    name: ${{ matrix.os }} · py${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        # Linux is also covered by ci.yml; including it here keeps the matrix
+        # apples-to-apples so a "passes on Linux, fails on Windows" diff is
+        # obvious from a single workflow.
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        # Endpoints of the supported range (3.11 oldest, 3.14 newest). The full
+        # 3.11–3.14 sweep stays on Linux in publish.yml; cross-OS is the new axis.
+        python-version: ["3.11", "3.14"]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.2.0
+
+      - name: Install dependencies
+        run: uv sync --all-extras
+
+      # bash on every OS (Git Bash on Windows) so the marker expression quotes
+      # identically across runners. Tests themselves run under the native Python
+      # for each OS, so OS-specific behaviour is still exercised.
+      - name: Run non-integration tests
+        shell: bash
+        run: uv run pytest -m "not integration"
--- a/src/envdrift/cli_commands/init_cmd.py
+++ b/src/envdrift/cli_commands/init_cmd.py
@@ -178,6 +178,11 @@ def _module_header(class_name: str, env_file: Path) -> list[str]:
     would otherwise be mangled by string-escape interpretation.
     """
     env_file_literal = repr(str(env_file))
+    # The docstring embeds the path inside a triple-quoted string, so its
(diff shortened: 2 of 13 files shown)


jainal09:
Local preview on **macOS (ARM64)**: `pytest -m "not integration"` → **2466 passed, 4 skipped, 313 deselected** in 45s. So the macOS legs should be green; **Windows** is the leg to watch for path-separator / permission / default-encoding side effects. This PR edits the workflow file (in the path filter), so the matrix self-triggers here — the Windows result on this PR is the first real signal.

jainal09:
First matrix run: **ubuntu ✅✅ · macOS ✅✅ · windows ❌❌** (22 failed / 2421 passed, identical on py3.11 + py3.14). Exactly the cross-platform side effects the Linux-only CI never caught. Triaged + filed as #445 — splits into real `init` code bugs (encoding / unsafe path-literal, cousins of #443) and POSIX-assumption test-portability bugs. These are the backlog to clear before promoting this workflow to a required check.

jainal09:
**envdrift review-bot** (posted via `gh` using Jainal's account) — re-review of latest head `6e364fe3`


Reviewed the workflow, PR description, current checks, and Windows job logs.

**Findings**

1. **[medium] Merging this as-is adds a known-red workflow on `main`.** The workflow itself is valid and the Windows failures are real portability signals, but both Windows cells are currently failing while Ubuntu/macOS pass:
   - `windows-latest · py3.11`: `22 failed, 2421 passed, 27 skipped, 313 deselected`
   - `windows-latest · py3.14`: same failure shape

   The failed classes include Windows path escaping in generated Python/TOML, default-encoding assumptions, executable-bit/chmod assumptions, invalid Windows filename behavior, and POSIX path expectations. Because `.github/workflows/cross-platform.yml` also runs on `push` to `main`, merging before those are fixed/skipped/xfail'd means `main` will carry a red cross-platform workflow when touched paths are pushed. If this is discovery-only for now, make the Windows leg non-blocking/experimental or restrict this workflow to PR/manual until the matrix is green.

2. **[low] The future required-check plan conflicts with the PR path filter.** Lines 16-18 say to promote the workflow to required once green, but the `pull_request.paths` filter at lines 20-28 means unrelated PRs can skip the workflow. GitHub required checks that are skipped by path filtering can remain pending and block merges. Before making this required, remove the PR path filter or add a tiny always-running gate job that succeeds when the matrix is intentionally skipped.

Notes: the Greptile `setup-uv` cache suggestion is performance-only. The current red Windows jobs are not caused by dependency setup or workflow syntax; they are exposing product/test portability issues.

jainal09:
Addressed in 686116f: enabled `setup-uv`'s built-in cache (`enable-cache: true`) with a per-OS/Python `cache-suffix`, so the 6-job matrix reuses downloaded wheels instead of re-fetching each run.

jainal09:
**envdrift review-bot** (posted via `gh` using Jainal's account) — current-head re-review of `686116f1`


The new commit correctly addresses the Greptile performance item: `astral-sh/setup-uv@v8.2.0` now has `enable-cache: true` and a per-OS/Python `cache-suffix`.

**Remaining findings**

1. **[medium] The workflow is still red on Windows and still runs on `push` to `main`.** Current run `27210664610` on this head:
   - `windows-latest · py3.11`: failed, `22 failed, 2421 passed, 27 skipped, 313 deselected`
   - `windows-latest · py3.14`: failed, `22 failed, 2421 passed, 27 skipped, 313 deselected`
   - Ubuntu/macOS cross-platform cells passed.

   Current failures still include Windows path escaping in generated settings modules (`SyntaxError: unicodeescape`), Windows/default encoding errors (`UnicodeDecodeError`), TOML parsing failures from Windows path escapes, executable-bit/chmod assumptions, invalid Windows filename behavior, and POSIX absolute-path expectations. Because this workflow also runs on `push` to `main`, merging it while Windows is knowingly red will put a red cross-platform workflow on `main` for touched paths. If this is still discovery-only, make Windows non-blocking/experimental or keep the workflow PR/manual-only until the Windows backlog is fixed or explicitly skipped/xfail'd.

2. **[low] The future required-check plan still conflicts with the PR path filter.** Lines 16-18 still say to promote this workflow to required once green, while the PR trigger is path-filtered at lines 20-28. A required workflow skipped by path filters can leave checks pending on unrelated PRs. Before making this required, remove the PR path filter or add an always-running gate job that exits successfully when the matrix is intentionally skipped.

All other current checks are green/skipped except those two Windows matrix jobs.

Issue #443, opened 2026-06-09
Adversarial sweep (phase 4): 36 verified correctness/UX failures → real test-bench backlog


jainal09:
Multi-agent **adversarial** dogfood sweep of the real `envdrift` CLI (v10.15.0, dotenvx 1.66.0 / sops 3.13.1) — 16 cells throwing garbage/edge inputs, each finding **re-read from disk and independently reproduced by a skeptic agent** before logging. **37 candidates → 36 confirmed, 1 refuted.**

Backlog for the **real (non-mock) test bench** (sibling to #441). Each item becomes an integration test that drives the real binary/CLI and asserts the true on-disk outcome; `xfail(strict=False, reason="… (see #THIS)")` until its fix PR flips it green. Rule: *a clean error + non-zero exit is NOT a finding*; a traceback / wrong exit / plaintext-after-`[OK]` IS. Full repros live in the workflow transcript; condensed here.

## 🔴 Empty / content-free input → dotenvx fabricates secrets (data fabrication + loss)
- **HIGH** `envdrift` [wrong-state] — `encrypt` on an EMPTY .env silently fabricates 14 placeholder secrets (dotenvx scaffold) and reports success
  - obs: Reproduced exactly and deterministically (ran 3 fresh mktemp dirs, identical result). Starting state: .env is 0 bytes, `[ -s .env ]` confirms empty. After `envdrift encrypt .env`: console prints only `[OK] Encrypted .env using dotenvx`, exit code 0, NO warning
- **HIGH** `envdrift` [wrong-state] — encrypt on a 0-byte (or blank-line-only) .env injects a fake-secrets TEMPLATE, encrypts it, and reports clean success — original empty file destroyed, fabricated secrets written
  - obs: CONFIRMED independently. Three input variants all reproduce, exit code 0 each time, no warning of any kind:  1) 0-byte empty.env: BEFORE wc -c = 0. After `envdrift encrypt empty.env`: EXIT=0, stdout exactly "[OK] Encrypted empty.env using dotenvx", stderr empt
- **MED** `envdrift` [wrong-state] — encrypt on a comment-only .env prints '[OK] Encrypted' and exits 0 but is a complete NO-OP (file byte-identical, nothing encrypted) — and still litters an orphan .env.keys keypair
[...]

PR #444, opened 2026-06-09
fix(encrypt): verify on-disk outcome and refuse content-free files

jainal09:
## What

The dotenvx `encrypt` backend trusted the subprocess **exit code**. The
adversarial sweep (#443) showed dotenvx exits `0` in two dangerous cases, so
`envdrift` reported success while doing harm. This fixes the shared backend seam
and adds a **real-outcome** integration bench.

## The bugs (all verified, all HIGH)

| # | Input | dotenvx behaviour | old envdrift result |
|---|---|---|---|
| #1/#8 | empty / blank-line-only `.env` | scaffolds a 14-var **fake-secrets template** (HELLO, AWS keys, Stripe…) and exits 0 | `[OK] Encrypted` — original destroyed, fabricated secrets written |
| #23 | comment-only `.env` | no-op, exits 0 | `[OK] Encrypted` — silent no-op + orphan `.env.keys` |
| #4/#7 | `.env.keys` is a **directory** | can't write key, leaves plaintext, exits 0 | `[OK] Encrypted` — secrets in cleartext |
| #5 | **garbage** `.env.keys` | encrypts nothing, exits 0 | `[OK] Encrypted` |
| #6 | re-encrypt with **corrupted** key + new plaintext secret | leaves new secret in cleartext, exits 0 | `[OK] Encrypted` — plaintext leak |

## The fix — outcome verification at the seam

Both checks live in the **shared backend** (`encryption/dotenvx.py`), so
`encrypt`, `lock`, `sync` and `vault` are all covered (true single-source), and
both reuse existing primitives (no parallel logic):

1. **Before** delegating: refuse a file with zero parseable variables
   (`EnvParser`) → dotenvx can never scaffold a fake template.
2. **After** delegating: re-read the file and fail if any plaintext value
   survived (`has_plaintext_secret_value`) → never claim success while secrets
   sit in cleartext, regardless of dotenvx's exit code.

## The bench — real, not mocked

`tests/integration/test_encrypt_outcome_bench.py` drives the **real dotenvx
binary** and asserts the **true on-disk state** for every case (re-reads the
file; never trusts the `success` flag or `[OK]`). Proven real:

```
fix stashed  → 6 failed, 1 passed   (the bug reproduces; control passes)
fix restored → 7 passed
```

The one pre-existing unit test that mocked the wrapper to a no-op and asserted
`success is True` is updated to model the seam's real effect (write ciphertext)
so it stays honest under the new outcome check — that mock-of-the-behaviour was
exactly why the bug shipped.

## Scope

dotenvx backend only. SOPS backend, the `decrypt` seam, and the traceback /
machine-output / init-validate themes from #443 are follow-up themed PRs. Full
non-integration suite green (2468 passed); local gates (ruff/format/pyrefly/
bandit/markdownlint) clean.

Refs #443

## Summary by CodeRabbit

* **Documentation**
  * Added a "Safety guarantees" section for the encrypt command describing its validation behavior.

* **Improvements**
  * Encrypt now refuses to run on empty or content-free files.
  * Encrypt re-checks files after running to detect and report silent encryption failures.
  * Removed encrypt's --verify-vault; vault verification moved to decrypt --verify-vault.

* **Tests**
  * Added integration and unit tests covering outcome validation and edge cases.

<h3>Greptile Summary</h3>

This PR fixes a class of silent encryption failures in the dotenvx backend where dotenvx exits `0` without actually encrypting — fabricating a secrets template for empty files, or leaving plaintext values on disk when the key is missing or malformed. The fix adds two outcome-verification checks at the seam in `DotenvxEncryptionBackend.encrypt()` so that `lock`, `sync`, and `vault` all benefit.

- **Pre-guard** (`file_has_assignment`): refuses files with no assignment lines before dotenvx is invoked, preventing the fake-template scaffolding on empty/comment-only input; the helper uses `errors="replace"` to tolerate non-UTF-8 bytes and counts non-identifier keys (`X-API-KEY`, `1PASSWORD`) that `EnvParser` would skip.
- **Post-check** (`has_plaintext_secret_value`): re-reads the on-disk file after the subprocess returns and fails loudly if any plaintext value survived, catching the key-missing / key-corrupted / dir-instead-of-file failure modes that dotenvx silently ignores.
- **Tests**: integration bench drives the real dotenvx binary for 7 cases (6 adversarial + 1 happy path) and asserts true on-disk state; unit test updated so the mock models the actual ciphertext side-effect the post-check now requires.

Safe to merge — all changed paths are guarded by the new pre/post checks and covered by both unit and real-binary integration tests.

The two new checks (file_has_assignment before delegation, has_plaintext_secret_value after) are logically correct, mutually consistent in what they treat as content, and robust to the known edge cases (non-identifier keys, non-UTF-8 bytes, empty values). The unit test update honestly models the on-disk side-effect the post-check now depends on. The integration bench verifies all five adversarial scenarios against the real binary. No net-new failure modes are introduced.

| Filename | Overview |
|----------|----------|
| src/envdrift/encryption/dotenvx.py | Adds pre-guard (file_has_assignment) and post-check (has_plaintext_secret_value) to encrypt(); DotenvxError catch correctly moved outside the result path so the post-check always runs; deferred import inside the method is intentional and correct. |
| src/envdrift/core/partial_encryption.py | New file_has_assignment helper is deliberately looser than EnvParser: counts any key=value line, tolerates non-UTF-8 via errors="replace", does not exclude DOTENV_PUBLIC_KEY lines (intentional — they are real content). Logic is correct and well-tested. |
| tests/integration/test_encrypt_outcome_bench.py | Comprehensive real-binary bench covering 7 cases; correctly skips when dotenvx is absent; all assertions verify on-disk state rather than the returned flag. The non-identifier-key regression test (#444 review case) is appropriately guarded by dotenvx_on_path. |
| tests/unit/test_encryption_backends.py | Unit test for encrypt() updated to model the real seam effect (mock writes ciphertext to disk) so it stays honest under the new post-check; no issues. |
| tests/unit/test_partial_encryption.py | TestFileHasAssignment covers identifier keys, non-identifier keys, non-UTF-8 bytes, empty/comment/blank/keyless inputs; all meaningful edge cases for the new helper are exercised. |
| docs/cli/encrypt.md | Documentation accurately describes the two new guarantees; no issues. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["encrypt(env_file)"] --> B{file exists?}
    B -- No --> C["EncryptionResult(success=False, 'File not found')"]
    B -- Yes --> D{file_has_assignment?}
    D -- No --> E["EncryptionResult(success=False, 'Nothing to encrypt')"]
    D -- Yes --> F{dotenvx installed?}
    F -- No --> G["raise EncryptionNotFoundError"]
    F -- Yes --> H["wrapper.encrypt()"]
    H -- DotenvxError --> I["raise EncryptionBackendError"]
    H -- exits 0 --> J{has_plaintext_secret_value?}
    J -- True plaintext survives --> K["EncryptionResult(success=False, 'did not take effect')"]
    J -- False all encrypted --> L["EncryptionResult(success=True)"]
```

<sub>Reviews (2): Last reviewed commit: ["fix(encrypt): make the empty-guard robus..."](https://github.com/jainal09/envdrift/commit/3e0fb3bfb92eb3a4cdab6e7e94dfb7a388d88b33) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36380599)</sub>

### DIFF
--- a/src/envdrift/encryption/dotenvx.py
+++ b/src/envdrift/encryption/dotenvx.py
@@ -111,15 +111,20 @@ class DotenvxEncryptionBackend(EncryptionBackend):
                 file_path=env_file,
             )
 
-        # Refuse to "encrypt" a file with no variables. Handed an empty or
+        # Refuse to "encrypt" a file with no assignments. Handed an empty or
         # comment-only file, dotenvx scaffolds a placeholder-secrets template
         # (HELLO, AWS_ACCESS_KEY_ID, ...) into it and still exits 0, so a blind
         # delegation fabricates secrets the user never wrote and destroys the
-        # original content. Detect zero parseable variables up front and decline
-        # rather than letting that happen (envdrift #443).
-        from envdrift.core.parser import EnvParser
+        # original content. ``file_has_assignment`` counts ANY assignment line
+        # (including non-identifier keys like ``X-API-KEY`` the strict parser
+        # rejects) and tolerates non-UTF-8 bytes, so it neither false-refuses a
+        # file of non-identifier secrets nor crashes on a non-UTF-8 file (#443).
+        from envdrift.core.partial_encryption import (
+            file_has_assignment,
+            has_plaintext_secret_value,
+        )
 
-        if not EnvParser().parse(env_file).variables:
+        if not file_has_assignment(env_file):
             return EncryptionResult(
                 success=False,
                 message=(
@@ -135,7 +140,6 @@ class DotenvxEncryptionBackend(EncryptionBackend):
                 f"dotenvx is not installed.\n{self.install_instructions()}"
             )
 
-        from envdrift.core.partial_encryption import has_plaintext_secret_value
         from envdrift.integrations.dotenvx import DotenvxError
 
         try:
--- a/src/envdrift/encryption/dotenvx.py
+++ b/src/envdrift/encryption/dotenvx.py
@@ -111,11 +111,31 @@ class DotenvxEncryptionBackend(EncryptionBackend):
                 file_path=env_file,
             )
 
+        # Refuse to "encrypt" a file with no variables. Handed an empty or
+        # comment-only file, dotenvx scaffolds a placeholder-secrets template
+        # (HELLO, AWS_ACCESS_KEY_ID, ...) into it and still exits 0, so a blind
+        # delegation fabricates secrets the user never wrote and destroys the
+        # original content. Detect zero parseable variables up front and decline
+        # rather than letting that happen (envdrift #443).
+        from envdrift.core.parser import EnvParser
+
+        if not EnvParser().parse(env_file).variables:
+            return EncryptionResult(
+                success=False,
+                message=(
+                    f"Nothing to encrypt: {env_file} has no variables. Refusing "
+                    "to run the encryptor, which would otherwise scaffold "
+                    "placeholder secrets into the file."
+                ),
+                file_path=env_file,
+            )
+
         if not self.is_installed():
             raise EncryptionNotFoundError(
                 f"dotenvx is not installed.\n{self.install_instructions()}"
             )
 
+        from envdrift.core.partial_encryption import has_plaintext_secret_value
         from envdrift.integrations.dotenvx import DotenvxError
 
         try:
@@ -126,13 +146,30 @@ class DotenvxEncryptionBackend(EncryptionBackend):
                 env=kwargs.get("env"),
                 cwd=kwargs.get("cwd"),
             )
-            return EncryptionResult(
-                success=True,
-                message=f"Encrypted {env_file}",
-                file_path=env_file,
-            )
         except DotenvxError as e:
             raise EncryptionBackendError(f"dotenvx encryption failed: {e}") from e
+
+        # Verify the encryption actually took effect rather than trusting the
+        # exit code. dotenvx can exit 0 *without* encrypting when the private key
+        # is missing or malformed (e.g. .env.keys is a directory, garbage, or a
+        # mismatched key); it only prints a warning to stderr. Re-read the file
+        # and confirm no plaintext value survived (envdrift #443).
+        if has_plaintext_secret_value(env_file):
+            return EncryptionResult(
+                success=False,
+                message=(
+                    f"Encryption did not take effect: {env_file} still contains "
+                    "plaintext values. This usually means the encryption key was "
+                    "missing or invalid."
+                ),
+                file_path=env_file,
+            )
+
+        return EncryptionResult(
+            success=True,
+            message=f"Encrypted {env_file}",
+            file_path=env_file,
+        )
 
     def decrypt(
         self,
--- a/docs/cli/encrypt.md
+++ b/docs/cli/encrypt.md
@@ -21,6 +21,19 @@ The `encrypt` command works with [dotenvx](https://dotenvx.com/) or
 
 If `--backend` is omitted, envdrift uses the configured backend (envdrift.toml/pyproject.toml) or defaults to dotenvx.
 
+### Safety guarantees
+
+The `encrypt` command verifies its own outcome rather than trusting the backend's
+exit code (dotenvx can exit `0` without encrypting):
+
+- **Refuses content-free files.** An empty, blank-line-only, or comment-only file
+  has no variables to encrypt, so the command declines with a non-zero exit
+  instead of letting dotenvx scaffold a placeholder-secrets template into it.
+- **Reports silent encryption failures.** When the key is missing or malformed
+  (a `.env.keys` that is a directory, garbage, or a mismatched key), the file is
+  re-read after the call; if any plaintext value survives, the command fails
+  loudly instead of printing `[OK]`.
+
 ## Arguments
 
 | Argument   | Description           | Default |
--- a/src/envdrift/core/partial_encryption.py
+++ b/src/envdrift/core/partial_encryption.py
@@ -252,6 +252,31 @@ def has_plaintext_secret_value(file_path: Path) -> bool:
     return any(_line_has_plaintext_secret(raw_line) for raw_line in content.splitlines())
 
 
+def file_has_assignment(file_path: Path) -> bool:
+    """Return True if the file contains at least one ``KEY=value`` assignment.
+
+    Deliberately looser than ``EnvParser`` (and robust to non-UTF-8 bytes via
+    ``errors="replace"``): it counts ANY assignment line, including keys the
+    strict parser rejects (a leading digit like ``1PW``, a dash like
+    ``X-API-KEY``, or a non-ASCII identifier). Used to decide whether a file has
+    anything to encrypt — a parser-variable count would miss non-identifier-keyed
+    secrets and wrongly report the file as empty, and ``EnvParser.parse`` would
+    raise ``UnicodeDecodeError`` on a non-UTF-8 file.
+    """
+    if not file_path.exists():
+        return False
+
+    content = file_path.read_text(encoding="utf-8", errors="replace")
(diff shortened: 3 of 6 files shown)


src/envdrift/encryption/dotenvx.py
```diff
@@ -111,11 +111,31 @@ def encrypt(
...
+        if not EnvParser().parse(env_file).variables:
+            return EncryptionResult(
+                success=False,
+                message=(
+                    f"Nothing to encrypt: {env_file} has no variables. Refusing "
+                    "to run the encryptor, which would otherwise scaffold "
+                    "placeholder secrets into the file."
+                ),
+                file_path=env_file,
+            )
```
jainal09, line 136: 🤖 **envdrift review-bot** · high · regression


The pre-guard decides whether to encrypt by counting `EnvParser().parse(env_file).variables`. But `EnvParser.LINE_PATTERN` only accepts keys matching `[A-Za-z_][A-Za-z0-9_]*`, so it silently drops non-identifier keys (`X-API-KEY`, `1PASSWORD_TOKEN`, `my.api.key`, unicode `CAFÉ`) — the exact parser hole #443 documents under init. A `.env` whose keys are ALL non-identifier therefore parses to **zero** variables and is refused with `Nothing to encrypt: <file> has no variables` — even though the file is full of real secrets that dotenvx encrypts fine.

I verified both halves in a scratch dir:
- `EnvParser().parse` on `X-API-KEY=supersecret123` / `1PASSWORD_TOKEN=...` / `my.api.key=...` / `CAFÉ=...` each yields `variables=[]` → REFUSED.
- Real `dotenvx 1.66.0 encrypt -f .env` on `X-API-KEY=supersecret123` succeeds, rewriting it as `X-API-KEY=encrypted:BMvuGK...` with a `DOTENV_PUBLIC_KEY` header.

Impact: this is a regression. Before this PR, `encrypt` delegated to dotenvx (which has its own, more permissive parser) and these files encrypted successfully; after this PR they are blocked, and the message ("has no variables") is actively wrong — the file plainly has assignments. The post-encryption check correctly handles these keys (`_is_secret_var_line` treats `X-API-KEY=encrypted:...` as ciphertext), so only the pre-guard is the problem.

Suggested direction: gate refusal on "no encryptable content" using a check that matches dotenvx's notion of an assignment (e.g. a presence-of-`KEY=value` scan independent of the identifier regex), not on envdrift's stricter `EnvParser`; and/or word the refusal so it cannot fire on a file that visibly contains assignments. A regression test should cover a `.env` containing only a non-identifier key.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit fd56a25


The new empty-file guard is correct in spirit but too strict in implementation: it relies on `EnvParser`, whose identifier-only key regex silently drops valid non-identifier keys, regressing a previously-working encrypt flow with a misleading message.

**Findings (1):**
- [high] New empty-file guard false-refuses valid files whose keys aren't envdrift identifiers (dash/digit/dot/unicode), with a misleading "has no variables" message — `src/envdrift/encryption/dotenvx.py:131` (inline)

jainal09:
Addressed both review findings in 3e0fb3b — both were the empty-guard's use of `EnvParser().parse()`:

- **non-identifier keys** (greptile/review-bot): `LINE_PATTERN` rejects `X-API-KEY`/`1PASSWORD`, so a file of only such secrets parsed to 0 vars and would've been wrongly refused. Now uses a new `file_has_assignment()` that counts **any** `KEY=value` line.
- **non-UTF-8** (greptile P1): `file_has_assignment()` reads with `errors="replace"`, so the guard no longer `UnicodeDecodeError`s on a non-UTF-8 `.env`.

Added a `file_has_assignment` unit suite (identifier / non-identifier / non-UTF-8 bytes / empty-comment-keyless) + an integration regression asserting a file of only non-identifier secrets is **encrypted, not refused**. Full non-integration suite green (2472).

@coderabbitai review

src/envdrift/encryption/dotenvx.py
jainal09: 🤖 **envdrift review-bot** — resolved ✅

Verified at `3e0fb3b`: the content-free/empty guard now recognizes valid non-identifier dotenv keys (X-API-KEY, dotted, leading-digit, unicode) so files with real secrets under such keys are no longer misclassified as content-free and refused — and it's robust to non-UTF-8 input, while still refusing genuinely empty files. New tests cover both. Resolved, thanks.

jainal09:
🤖 **envdrift review-bot** — automated review of latest head `3e0fb3bf`


Reviewed the PR description, #443 context, existing inline comments, current diff, docs, and the dotenvx encryption seam. Focused local checks passed:

- `uv run --extra dev pytest tests/unit/test_partial_encryption.py::TestFileHasAssignment tests/unit/test_encryption_backends.py -q` → `83 passed`
- `uv run --extra dev pytest tests/integration/test_encrypt_outcome_bench.py -q` with real `/Users/jainal09/.local/bin/dotenvx` → `8 passed`

**Findings (3):**

- [high] The post-check can still report success while a real plaintext variable survives if the key starts with `DOTENV_PUBLIC_KEY`.

`src/envdrift/core/partial_encryption.py:155` excludes every line whose stripped text starts with `DOTENV_PUBLIC_KEY`, and `src/envdrift/encryption/dotenvx.py:161` trusts `has_plaintext_secret_value()` as the outcome check. That is broader than dotenvx's actual public-key artifact and hides real variables such as `DOTENV_PUBLIC_KEYSTORE_TOKEN` / `DOTENV_PUBLIC_KEYRING`.

Repro at `3e0fb3bf` with real dotenvx: create `.env.keys` as a directory so dotenvx exits without encrypting, then encrypt a file containing `DOTENV_PUBLIC_KEYSTORE_TOKEN=ghp_secretplaintext1234567890`. The backend returns `success=True` / `Encrypted ...`, the file is byte-for-byte unchanged, and `has_plaintext_secret_value()` returns `False`. The same setup with `API_KEY=...` correctly returns `success=False`. Please narrow the public-key exclusion to actual dotenvx metadata, ideally by exact key-name pattern and/or EC public-key value shape, instead of a raw prefix match.

- [high] Non-UTF-8 `.env` files still traceback despite the new guard claiming that path tolerates non-UTF-8 bytes.

`file_has_assignment()` reads with `errors="replace"`, but a file containing bytes `API_KEY=caf\xe9\n` still raises `UnicodeDecodeError`. Backend repro: `DotenvxEncryptionBackend().encrypt()` raises `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9...`; the traceback reaches `src/envdrift/integrations/dotenvx.py:810`, where `_clean_mismatched_headers()` does `read_text(encoding="utf-8")` and only catches `OSError`. CLI repro also exits 1 with a Rich traceback before clean reporting because `encrypt_cmd` parses via `EnvParser` (`src/envdrift/core/parser.py:167`). Please convert this to a clean failure path or make all reads in this flow consistently tolerate undecodable bytes.

- [medium] A file containing only a dotenvx public-key artifact reports `Encrypted ...` even though nothing was encrypted.

For a file containing only `DOTENV_PUBLIC_KEY="03...64 hex..."`, `file_has_assignment()` returns `True`, `has_plaintext_secret_value()` returns `False`, `is_file_encrypted()` returns `False`, `.env.keys` is not created, and `DotenvxEncryptionBackend().encrypt()` returns `success=True` while leaving the file unchanged. That contradicts the documented safety guarantee in `docs/cli/encrypt.md:26-35` that the command verifies its outcome and fails when encryption did not happen. Please either decline public-key-only files as having no encryptable variables or explicitly detect/report the no-op instead of success.

Current live checks were green at review start; please re-run after fixing because all three gaps sit outside the new passing bench cases.

Issue #445, opened 2026-06-09
Windows CI: 22 cross-platform failures surfaced by the new matrix (#442) — init encoding/path-literal bugs + test-portability


jainal09:
The new **Cross-Platform Tests** matrix (#442) runs `pytest -m "not integration"`
on ubuntu/macOS/Windows. **macOS and Linux are fully green; Windows fails 22 /
2443** (py3.11 and py3.14 identically). These are pre-existing cross-platform
side effects the Linux-only CI never surfaced. Triage splits them into real code
bugs vs. test-portability bugs.

## A. Likely real code bugs (fix the code)

Cross-platform cousins of the #443 encoding/`init` findings — these tests assert
correct behaviour and pass on Linux/macOS, so Windows is revealing a real defect:

- **`init` crashes on non-ASCII / Unicode-digit values** — `UnicodeDecodeError:
  'utf-8' codec can't decode byte 0xb2` (a file is read/written without an
  explicit `encoding="utf-8"`; on Windows the default codec differs).
  - `tests/unit/test_cli.py::TestInitCommand::test_init_unicode_digit_value_does_not_crash`
  - `tests/unit/test_api.py::TestInitAPI::test_init_unicode_digit_inferred_as_str`
- **`init` writes an env-file path as an UNSAFE string literal** — a Windows path
  with backslashes becomes `\U…` → `SyntaxError` in the generated module (must use
  `repr()` / a safe literal).
  - `tests/unit/test_cli.py::TestInitCommand::test_init_env_file_path_with_escapes_is_safe_literal`
  - `…::test_init_keyword_var_name_produces_importable_module`
  - `…::test_init_keyword_collision_keeps_both_env_bindings`
  - `…::test_init_pydantic_reserved_field_names_produce_importable_module`
  - `…::test_init_aliased_field_keeps_typed_default`
  - `tests/unit/test_api.py::TestInitAPI::test_init_keyword_key_produces_importable_module`

## B. Likely test-portability bugs (fix the test, POSIX assumptions)

- **`HOME` vs `USERPROFILE` + hardcoded `.local/bin`** — tests monkeypatch `HOME`,
  but `Path.home()` reads `USERPROFILE` on Windows, so the user-bin fallback
  resolves to the real home.
  - `tests/scanner/test_gitleaks.py::TestGetVenvBinDir::test_fallbacks_to_user_bin`
[...]

PR #446, opened 2026-06-09
fix(init): alias non-identifier keys instead of dropping them

jainal09:
## What

The adversarial sweep (#443) found `init` silently **drops** every `.env` key the
strict `EnvParser` rejects — so a secret under a dash/digit/dot/emoji key never
reaches the schema, and its drift is invisible to `validate` too (round-trip
hole). Worse, it even discarded **valid** non-ASCII identifiers (`CAFÉ`, `ΑΛΦΑ`).

## Fix

The sanitize/alias machinery already existed but only ran on parser-accepted
keys. Feed the parser-rejected keys through the same path:

| `.env` key | before | after |
|---|---|---|
| `1PASSWORD_TOKEN` | dropped + `[WARN]` | `field_1PASSWORD_TOKEN: str = Field(alias='1PASSWORD_TOKEN')` |
| `X-API-KEY` | dropped | `X_API_KEY: str = Field(alias='X-API-KEY')` |
| `KEY🔑` | dropped | sanitized + aliased |
| `CAFÉ`, `ΑΛΦΑ` | dropped as "non-identifier" | bare field, no alias |

Values for recovered keys are unquoted with the parser's own helpers
(single-source). The old drop-and-warn path — including the public
`api.init` `UserWarning` — is removed; there's nothing left to drop.

Also: write the generated module with `encoding="utf-8"` (`init_cmd` + `api`) so a
non-ASCII field name round-trips on platforms whose default encoding isn't UTF-8
(the Windows leg of the matrix, #445).

## Tests / docs

Updated the old drop/warn tests (`test_cli`, `test_api`) to the new alias
behaviour, added a case asserting Unicode identifiers stay bare while an emoji
key aliases, and verified end-to-end on macOS (all keys land in the schema, the
module imports, aliases bind back). Full non-integration suite green (2467);
ruff/format/pyrefly/bandit/markdownlint clean. Docs synced (`docs/cli/init.md`).

Refs #443, #445

## Release Notes

* **New Features**
  * Environment variables with non-identifier names (e.g., `2FA_ENABLED`, `MY-DASH-VAR`) are now included in generated settings with proper field aliases.
  * Enhanced support for Unicode identifiers in environment variable names.

* **Bug Fixes**
  * Output files now use UTF-8 encoding explicitly.
  * Settings schema generation preserves all environment variables.

* **Documentation**
  * Updated guides for variable name handling and sanitization practices.

<h3>Greptile Summary</h3>

This PR fixes a data-loss bug where `init` silently dropped every `.env` key the strict `EnvParser` rejected (leading digits, dashes, dots, emoji, and valid non-ASCII identifiers like `CAFÉ`). Rejected keys are now recovered via a second raw-assignment scan, run through the existing sanitize/alias pipeline, and emitted as aliased fields — so the generated module is both complete and importable.

- `EnvParser.value_from_raw` is promoted to a public method so the recovery path (`_unparsed_assignments`) reuses canonical unquoting logic without calling private helpers.
- The `SettingsGeneration.unparsed_keys` field and all drop/warn paths (`UserWarning` in `api.init`, `[WARN]` in CLI) are removed since no key is dropped any more.
- File writes in both `init_cmd` and `api` now pass `encoding=\"utf-8\"` explicitly, ensuring non-ASCII field names round-trip correctly on platforms whose default encoding isn't UTF-8.

Safe to merge — the change recovers previously-dropped keys and aliases them into the schema without altering any existing key paths.

Every changed code path has a clear, well-reasoned implementation: deduplication in `_unparsed_assignments` prevents double-emission, `value_from_raw` is the single source of unquoting for both parsed and recovered keys, and the `_sanitize_identifier` / `_resolve_field_name` logic that already handled keywords is unchanged. End-to-end importability tests cover the non-identifier and Unicode edge cases.

| Filename | Overview |
|----------|----------|
| src/envdrift/cli_commands/init_cmd.py | Core of the fix: `_unparsed_assignments` now returns a `name -> value` dict for parser-rejected keys, `generate_settings_module` merges it with parsed keys before rendering, and drop/warn paths are removed. |
| src/envdrift/core/parser.py | Introduces `value_from_raw` as a public method on `EnvParser`, consolidating inline-comment-strip + unquote into a single canonical path. |
| src/envdrift/api.py | Removes the `warnings.warn` path and adds `encoding='utf-8'` to `output.write_text`. |
| tests/unit/test_api.py | Replaces warn/drop tests with alias-content checks and an importability proof; adds a no-spurious-warnings guard. |
| tests/unit/test_cli.py | Mirrors API test updates for the CLI path; new test verifies Greek/emoji edge cases end-to-end by importing the generated module. |
| docs/cli/init.md | Updates the non-identifier key section from warn-and-drop to alias, syncing docs with the new behaviour. |

</details>

<sub>Reviews (4): Last reviewed commit: ["docs(init): make all alias examples sing..."](https://github.com/jainal09/envdrift/commit/11c893d0a834bce335574221a1d5c54d109d63d5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36441419)</sub>

### DIFF
--- a/docs/cli/init.md
+++ b/docs/cli/init.md
@@ -150,7 +150,7 @@ The generated module is always valid, importable Python:
   name plus a Pydantic `alias` so the field still binds to the original variable:
 
   ```python
-  class_: str = Field(alias="class")
+  class_: str = Field(alias='class')
   ```
 
 - **Pydantic-reserved names get an alias.** A key that is a valid identifier but
@@ -160,7 +160,7 @@ The generated module is always valid, importable Python:
   `alias`, so it cannot raise at import or shadow model internals:
 
   ```python
-  field_model_dump: str = Field(alias="model_dump")
+  field_model_dump: str = Field(alias='model_dump')
   ```
 
 - **Non-identifier keys are aliased, not dropped.** Keys that are not
--- a/docs/cli/init.md
+++ b/docs/cli/init.md
@@ -163,10 +163,19 @@ The generated module is always valid, importable Python:
   field_model_dump: str = Field(alias="model_dump")
   ```
 
-- **Non-identifier keys are reported.** Keys the parser cannot read because they
-  are not identifier-style (a leading digit like `2FA_ENABLED`, or a dash like
-  `MY-DASH-VAR`) are skipped and listed in a `[WARN]` line, never silently
-  dropped.
+- **Non-identifier keys are aliased, not dropped.** Keys that are not
+  identifier-style (a leading digit like `2FA_ENABLED`, a dash like
+  `MY-DASH-VAR`, a dot, or an emoji) are emitted with a sanitized attribute name
+  plus a Pydantic `alias`, so every key ends up in the schema and still binds to
+  the original environment variable:
+
+  ```python
+  field_2FA_ENABLED: str = Field(alias="2FA_ENABLED")
+  MY_DASH_VAR: str = Field(alias="MY-DASH-VAR")
+  ```
+
+- **Valid non-ASCII identifiers are kept verbatim.** A key that passes
+  `str.isidentifier()` (e.g. `CAFÉ`, `ΑΛΦΑ`) becomes a bare field with no alias.
 
 - **Invalid class names fail fast.** A `--class-name` that is not a valid Python
   identifier (e.g. `123Bad`) or is a keyword (e.g. `class`) errors with a
--- a/docs/cli/init.md
+++ b/docs/cli/init.md
@@ -170,8 +170,8 @@ The generated module is always valid, importable Python:
   the original environment variable:
 
   ```python
-  field_2FA_ENABLED: str = Field(alias="2FA_ENABLED")
-  MY_DASH_VAR: str = Field(alias="MY-DASH-VAR")
+  field_2FA_ENABLED: str = Field(alias='2FA_ENABLED')
+  MY_DASH_VAR: str = Field(alias='MY-DASH-VAR')
   ```
 
 - **Valid non-ASCII identifiers are kept verbatim.** A key that passes
--- a/src/envdrift/api.py
+++ b/src/envdrift/api.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import warnings
 from pathlib import Path
 
 from envdrift.core.diff import DiffEngine, DiffResult
@@ -113,14 +112,11 @@ def init(
         class_name (str): Name to use for the generated Settings class.
         detect_sensitive (bool): If True, attempt to detect sensitive variables and mark them in the generated fields.
 
-    Non-identifier .env keys (leading digits, dashes) are skipped by the strict
-    parser; keys that are valid identifiers but Python keywords (`class`,
-    `import`) are emitted with a sanitized attribute name plus a Pydantic
-    ``alias`` so the generated module always imports cleanly.
-
-    Any keys the strict parser cannot read (e.g. ``2FA_ENABLED``, ``MY-DASH-VAR``)
-    are dropped from the generated module; the API surfaces them via a
-    ``UserWarning`` so callers are not left silently with an incomplete file.
+    Every .env key becomes a field. Keys the strict parser rejects (leading
+    digits, dashes, dots) and Python keywords (`class`, `import`) are emitted
+    with a sanitized attribute name plus a Pydantic ``alias`` so the generated
+    module always imports cleanly while still binding to the original variable;
+    a valid non-ASCII identifier (``CAFÉ``) is emitted unchanged.
 
     Returns:
         Path: The path to the written settings file.
@@ -136,16 +132,7 @@ def init(
 
     result = generate_settings_module(env_file, class_name, detect_sensitive)
 
-    # Warn *before* writing: a caller running under
-    # ``warnings.filterwarnings("error")`` promotes this to an exception, and we
-    # must not leave a half-written settings.py on disk when that happens.
-    if result.unparsed_keys:
-        warnings.warn(
-            "Skipped .env variable(s) the parser cannot read "
-            f"(non-identifier keys): {', '.join(result.unparsed_keys)}",
-            UserWarning,
-            stacklevel=2,
-        )
-
-    output.write_text(result.source)
+    # encoding="utf-8" so a non-ASCII field name/value round-trips on platforms
+    # whose default text encoding is not UTF-8.
+    output.write_text(result.source, encoding="utf-8")
     return output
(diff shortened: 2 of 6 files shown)


jainal09:
Addressed in e3a5547:
- **Private-parser coupling**: added a public `EnvParser.value_from_raw()` (canonical strip-comment/whitespace/quotes); `_unparsed_assignments` and `parse_string` both use it — decoupled and single-source.
- **Doc quote-style**: the generator uses `repr()` → single quotes; updated init.md examples to `alias='2FA_ENABLED'`.

Full non-integration suite green (2469).

jainal09:
**envdrift review-bot** (posted via `gh` using Jainal's account) — re-review of latest head `e3a55473`


Reviewed the new commit that addressed the Greptile comments, then reran focused tests and repros in an isolated worktree.

Focused tests: `uv run --extra dev pytest tests/unit/test_api.py tests/unit/test_cli.py -q` -> `188 passed`.

**Findings**

1. **[high] Aliased fields still do not round-trip through `envdrift validate`.** The PR fixes generated Pydantic importability for keys like `X-API-KEY`, but validation still compares the strict parser's env keys against field names, not aliases. Repro on this head:

   ```sh
   tmp=$(mktemp -d)
   printf 'DATABASE_URL=postgres://x\nX-API-KEY=abc123\n' > "$tmp/.env"
   uv run envdrift init "$tmp/.env" --output "$tmp/settings.py" --class-name Cfg
   uv run envdrift validate "$tmp/.env" --schema settings:Cfg --service-dir "$tmp" --ci --no-check-encryption
   ```

   The generated module contains `X_API_KEY: str = Field(alias='X-API-KEY')`, but validate fails with `MISSING REQUIRED VARIABLES: X_API_KEY`. The inverse also happens for defaulted aliases: generate from `2FA_ENABLED=true`, remove `2FA_ENABLED` from `.env`, and validate still passes because `field_2FA_ENABLED` has `default=True`. That means the issue cited in #443 is not fully fixed: aliased keys can still be invisible or misreported in the init -> validate round trip.

2. **[high] Generated modules are still non-importable on Windows-style `.env` paths.** `_module_header()` correctly emits `model_config.env_file` with `repr(str(env_file))`, but the class docstring still embeds the raw path:

   `src/envdrift/cli_commands/init_cmd.py:214`

   Repro:

   ```py
   from pathlib import PureWindowsPath
   from envdrift.cli_commands.init_cmd import _module_header
   src = "\n".join(_module_header("Cfg", PureWindowsPath(r"C:\Users\runneradmin\tmp\.env")) + ["    API_KEY: str"])
   compile(src, "<generated>", "exec")
   ```

   This raises `SyntaxError: (unicode error) 'unicodeescape' ... truncated \UXXXXXXXX escape` because the generated docstring contains `C:\Users\...` as an interpreted escape sequence.

3. **[medium] `_sanitize_identifier()` can emit invalid Python identifiers for some Unicode keys.** `re.sub(r"\W", "_", name)` leaves Unicode “word” characters that Python does not allow in identifiers. Repro:

   ```sh
   tmp=$(mktemp -d)
   printf 'KEY²=abc\n' > "$tmp/.env"
   uv run envdrift init "$tmp/.env" --output "$tmp/settings.py" --class-name Cfg
   uv run python -m py_compile "$tmp/settings.py"
   ```

   `init` exits 0 and writes `KEY²: str = Field(alias='KEY²')`, then compilation fails with `SyntaxError: invalid character '²'`.

4. **[low] API reference docs still describe the removed skip-and-warning behavior.** `docs/reference/api.md:111-117` still says non-identifier keys are skipped and `envdrift.api.init()` emits `UserWarning`, but this PR removes that behavior and now aliases those keys.

docs/cli/init.md
```diff
@@ -163,10 +163,19 @@ The generated module is always valid, importable Python:
   field_model_dump: str = Field(alias="model_dump")
```
jainal09: 🤖 **envdrift review-bot** · low · doc-mismatch


The commit message for e3a5547 states its purpose is to "fix doc alias quote style" because the generator renders aliases with `repr()` (single quotes). The diff only updated the two non-identifier examples (lines 173-174) to `alias='2FA_ENABLED'` / `alias='MY-DASH-VAR'`, but the two earlier examples in the *same* `## Variable Names and Identifiers` section were left double-quoted:

- `docs/cli/init.md:153` — `class_: str = Field(alias="class")`
- `docs/cli/init.md:163` — `field_model_dump: str = Field(alias="model_dump")`

These are produced by the identical code path (`_field_call_args` -> `f"alias={alias!r}"`), so the generator actually emits **single** quotes. I confirmed by generating from `class=x` / `model_dump=y` / `schema=z` against the PR-head source:

```
class_: str = Field(alias='class')
field_model_dump: str = Field(alias='model_dump')
field_schema: str = Field(alias='schema')
```

The project's own test suite agrees and contradicts the docs: `tests/unit/test_api.py:230` asserts `"alias='class'" in content` and `tests/unit/test_cli.py:1610` asserts `"alias='model_dump'" in content`. So lines 153 and 163 now mislead users who grep the generated file for the exact string — the same confusion greptile flagged for line 175, just at two locations the fix missed. This is a distinct location from greptile's already-resolved inline comment on line 175.

Impact: low (cosmetic doc inaccuracy), but it directly undercuts the stated goal of this commit and leaves the section internally inconsistent (two double-quoted, two single-quoted). Suggested direction: switch lines 153 and 163 to single quotes to match the generator/tests.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit e3a5547


The commit's stated goal ("fix doc alias quote style") is incomplete: two sibling `Field(alias=...)` examples in the same section still use double quotes the generator never emits.

**Findings (1):**
- [low] Doc "alias quote style" fix is incomplete: two sibling `Field(alias=...)` examples still use double quotes the generator never emits — `docs/cli/init.md:153` and `:163` (inline anchored at line 163, the in-diff location; line 153 is below the diff hunk so could not be anchored there)

jainal09:
Fixed in 99e0f41 — the two remaining `alias="…"` examples (`class`, `model_dump`) in the same section are now single-quoted to match the `repr()` output and the test assertions. Grepped all of docs/ — no double-quoted alias examples remain.

docs/cli/init.md
jainal09: 🤖 **envdrift review-bot** — resolved ✅

Verified at `11c893d`: all `Field(alias=...)` examples in docs/cli/init.md now use a consistent single-quote style matching what the generator emits. Resolved, thanks.

jainal09:
Superseding this with a single PR that fixes the init→validate round-trip atomically. As CodeRabbit correctly flagged, aliasing non-identifier/Unicode keys into the schema (this PR) without teaching `validate` to recover + alias-match those keys makes the documented init→validate workflow fail (false 'MISSING REQUIRED'). Reopening the work as a combined init+validate PR so main never sees the broken round-trip.

PR #447, opened 2026-06-09
fix(decrypt): honest no-op on non-encrypted files, never corrupt them

jainal09:
## What

The adversarial sweep (#443) found the dotenvx `decrypt` backend reported
`[OK] Decrypted` **unconditionally**, trusting the exit code:

- a **plaintext** `.env` was reported as decrypted, and dotenvx's line-ending
  normalisation silently rewrote bytes (stripping `CR` on CRLF files);
- a **binary blob** handed to `decrypt` was **corrupted outright** while exiting 0.

## Fix

A pre-check at the shared backend seam: if the file holds no ciphertext value
(`is_file_encrypted`, which reads `errors="replace"` so binary input can't crash
it), report an **honest no-op** and never invoke dotenvx. A new
`EncryptionResult.changed` flag lets the CLI print `Nothing to decrypt: …`
instead of a misleading `Decrypted`. After a real decrypt, verify no ciphertext
survived (bad-key backstop) rather than trusting the exit code.

## The bench (real dotenvx)

`tests/integration/test_decrypt_outcome_bench.py` asserts true on-disk state:
plaintext no-op leaves bytes identical · CRLF bytes preserved · a **binary blob
is byte-for-byte intact** (no corruption, no `UnicodeDecodeError`) · a real
round-trip restores plaintext · a corrupted key errors cleanly without
corrupting the ciphertext. Proven **red-without-fix → green-with-fix**.

The mocked unit decrypt test is updated to model the seam's real effect (write
plaintext) so it stays honest under the new outcome check. Full non-integration
suite green (2466); all gates clean; `docs/cli/decrypt.md` synced.

## Scope

dotenvx backend only. The binary→`UnicodeDecodeError` traceback on the
**auto-detect** and **SOPS** paths (#443 #16/#17) is the separate traceback /
sops theme.

Refs #443

<h3>Greptile Summary</h3>

This PR fixes two file-corruption bugs in the dotenvx `decrypt` backend: it was calling dotenvx unconditionally, which caused plaintext files to be silently rewritten (stripping CRLF bytes) and binary blobs to be corrupted outright, while still reporting success. The fix adds a pre-check using `is_file_encrypted` (reading with `errors="replace"` to tolerate binary input) and returns an honest no-op result (`changed=False`) without ever invoking dotenvx when no ciphertext is present. A post-decrypt backstop re-reads the file after a real decrypt to confirm ciphertext was removed rather than trusting the exit code.

- **`EncryptionResult.changed`** flag added to `base.py` so the CLI can distinguish a real decryption from a no-op and print an accurate message.
- **`_report_decrypt_result`** extracted in `encryption.py` to route the three outcomes (success, no-op, failure) to the correct console output and exit code.
- **Integration bench** (`test_decrypt_outcome_bench.py`) asserts true on-disk byte state for plaintext, CRLF, binary, round-trip, and corrupted-key scenarios; the first three run without the dotenvx binary at all.

Safe to merge. The change is additive and defensive: it only prevents dotenvx from being invoked when there is nothing to decrypt, and adds a post-decrypt check to catch silent key failures.

The pre-check, post-decrypt backstop, and the `changed` flag are all covered by unit tests (including the exact backstop scenario where exit 0 leaves ciphertext) and integration tests that assert byte-level file identity. The `is_file_encrypted` function used for both checks already existed and is tested separately. No logic paths were removed; the change only adds guards before and after the existing dotenvx invocation.

| Filename | Overview |
|----------|----------|
| src/envdrift/encryption/dotenvx.py | Core fix: pre-check with `is_file_encrypted` before invoking dotenvx, and post-decrypt backstop that re-reads the file to confirm ciphertext was removed rather than trusting exit code 0. Both checks use `errors="replace"` so binary files can't crash the guard. Logic is correct for all documented edge cases. |
| src/envdrift/encryption/base.py | Adds `changed: bool = True` field to `EncryptionResult` dataclass. Default of `True` is correct for backwards compatibility; existing callers that don't set it (encrypt path, file-not-found) continue to signal "changed" as before, and the field is only consumed in the `success=True` branch of `_report_decrypt_result`. |
| src/envdrift/cli_commands/encryption.py | Extracts `_report_decrypt_result` helper to distinguish real decryption (`changed=True` → success message) from honest no-op (`changed=False` → warning message) and failure (exit 1). The surrounding `decrypt_cmd` `is_installed()` guard at line 689 still fires before the backend's no-op path for users without dotenvx installed, but that is pre-existing behaviour unchanged by this PR. |
| tests/integration/test_decrypt_outcome_bench.py | New integration bench covering: plaintext no-op, CRLF byte-identity, binary blob non-corruption (all three run without the dotenvx binary), plus binary-requiring round-trip restore and corrupted-key error-without-corruption. Tests correctly skip when the binary is absent via the `dotenvx_on_path` fixture. |
| tests/unit/test_encryption_backends.py | Updated `test_decrypt_success` to write plaintext in the mock (so the post-decrypt `is_file_encrypted` check passes), and added two new unit tests: honest no-op for plaintext files and the backstop path where ciphertext survives exit 0. All three key scenarios are now covered at the unit level. |
| tests/unit/coverage/test_cov_cli_commands_encryption.py | Adds `test_decrypt_noop_reports_nothing_to_decrypt` which verifies the CLI prints a "nothing to decrypt" warning (exit 0) when the backend returns `changed=False`. Uses a dummy backend so it doesn't touch dotenvx at all. |
| docs/cli/decrypt.md | Documentation-only addition describing the honest no-op behaviour for the dotenvx backend. Accurately reflects the new logic. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant CLI as decrypt_cmd (CLI)
    participant B as DotenvxEncryptionBackend
    participant IFE as is_file_encrypted()
    participant DX as dotenvx binary

    CLI->>CLI: is_installed() guard (pre-existing)
    CLI->>B: decrypt(env_file)
    B->>IFE: "is_file_encrypted(env_file) [errors=replace]"
    alt No ciphertext in file (plaintext / binary blob)
        IFE-->>B: False
        B-->>CLI: "EncryptionResult(success=True, changed=False)"
        CLI->>CLI: print_warning(Nothing to decrypt)
    else File has ciphertext values
        IFE-->>B: True
        B->>B: is_installed() check
        B->>DX: wrapper.decrypt(env_file)
        alt dotenvx exits non-zero
            DX-->>B: DotenvxError
            B-->>CLI: raises EncryptionBackendError
            CLI->>CLI: print_error + exit 1
        else dotenvx exits 0
            DX-->>B: ok
            B->>IFE: is_file_encrypted(env_file) [post-decrypt backstop]
            alt Ciphertext survived (bad/missing key)
                IFE-->>B: True
                B-->>CLI: "EncryptionResult(success=False, changed=False)"
                CLI->>CLI: print_error + exit 1
            else All ciphertext decrypted
                IFE-->>B: False
                B-->>CLI: "EncryptionResult(success=True, changed=True)"
                CLI->>CLI: print_success(Decrypted)
            end
        end
    end
```

1. `tests/unit/test_encryption_backends.py`, line 275-299 ([link](https://github.com/jainal09/envdrift/blob/8df3d5ba018508649bc43547b4bb5e0b0f4ac00c/tests/unit/test_encryption_backends.py#L275-L299)) 

   <a href="#">P2</a> **Post-decrypt backstop path has no unit test**

   The PR's key new invariant is "if dotenvx exits 0 but ciphertext survives, report `success=False`." The updated `test_decrypt_success` exercises the happy path (mock writes plaintext → `is_file_encrypted` returns False → success). The `DotenvxError` path is covered by `test_decrypt_wraps_dotenvx_error`. But there is no unit test where the mock leaves the file encrypted (i.e., does not write plaintext), which is the exact scenario the backstop at lines 200–211 of `dotenvx.py` is designed to catch. Without it, a regression that removes or short-circuits the post-decrypt check would go undetected by the unit suite.

</details>

<sub>Reviews (3): Last reviewed commit: ["fix(decrypt): address review — coverage,..."](https://github.com/jainal09/envdrift/commit/c735d76ce4f6cb8c2d1435e2eb70426620458571) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36441478)</sub>

### DIFF
--- a/docs/cli/decrypt.md
+++ b/docs/cli/decrypt.md
@@ -20,6 +20,14 @@ It can also **verify** that a key stored in your vault can decrypt the file with
 - Viewing encrypted values
 - Migrating to a different encryption system
 
+### Honest no-op on non-encrypted files
+
+If the target file has no encrypted values, `decrypt` reports an honest no-op
+(`Nothing to decrypt: <file> has no encrypted values.`) instead of claiming a
+decryption happened. The file is left byte-for-byte untouched — the backend is
+not invoked at all — so a plaintext file (including one with CRLF line endings)
+or a non-`.env` binary file is never silently rewritten or corrupted.
+
 ## Arguments
 
 | Argument   | Description                     | Default |
--- a/docs/cli/decrypt.md
+++ b/docs/cli/decrypt.md
@@ -20,13 +20,14 @@ It can also **verify** that a key stored in your vault can decrypt the file with
 - Viewing encrypted values
 - Migrating to a different encryption system
 
-### Honest no-op on non-encrypted files
+### Honest no-op on non-encrypted files (dotenvx backend)
 
-If the target file has no encrypted values, `decrypt` reports an honest no-op
-(`Nothing to decrypt: <file> has no encrypted values.`) instead of claiming a
-decryption happened. The file is left byte-for-byte untouched — the backend is
-not invoked at all — so a plaintext file (including one with CRLF line endings)
-or a non-`.env` binary file is never silently rewritten or corrupted.
+With the **dotenvx** backend, if the target file has no encrypted values
+`decrypt` reports an honest no-op (`Nothing to decrypt: <file> has no encrypted
+values.`) instead of claiming a decryption happened. The file is left
+byte-for-byte untouched — dotenvx is not invoked at all — so a plaintext file
+(including one with CRLF line endings) or a non-`.env` binary file is never
+silently rewritten or corrupted.
 
 ## Arguments
 
--- a/src/envdrift/encryption/dotenvx.py
+++ b/src/envdrift/encryption/dotenvx.py
@@ -203,6 +203,23 @@ class DotenvxEncryptionBackend(EncryptionBackend):
                 file_path=env_file,
             )
 
+        from envdrift.core.partial_encryption import is_file_encrypted
+
+        # Nothing to decrypt: the file holds no ciphertext value. dotenvx would
+        # still rewrite it (line-ending normalization, header cleanup) and exit 0
+        # — and handed a binary blob it corrupts the file outright — so report an
+        # honest no-op instead of a misleading "[OK] Decrypted" and never invoke
+        # the backend on a file that isn't encrypted. ``is_file_encrypted`` reads
+        # with errors="replace", so a binary/non-UTF-8 file cannot crash here
+        # (envdrift #443).
+        if not is_file_encrypted(env_file):
+            return EncryptionResult(
+                success=True,
+                changed=False,
+                message=f"Nothing to decrypt: {env_file} has no encrypted values.",
+                file_path=env_file,
+            )
+
         if not self.is_installed():
             raise EncryptionNotFoundError(
                 f"dotenvx is not installed.\n{self.install_instructions()}"
@@ -218,13 +235,27 @@ class DotenvxEncryptionBackend(EncryptionBackend):
                 env=kwargs.get("env"),
                 cwd=kwargs.get("cwd"),
             )
-            return EncryptionResult(
-                success=True,
-                message=f"Decrypted {env_file}",
-                file_path=env_file,
-            )
         except DotenvxError as e:
             raise EncryptionBackendError(f"dotenvx decryption failed: {e}") from e
+
+        # Verify the decryption took effect rather than trusting the exit code:
+        # dotenvx can exit 0 while leaving ciphertext in place when the key is
+        # missing or invalid. If any encrypted value survived, report failure.
+        if is_file_encrypted(env_file):
+            return EncryptionResult(
+                success=False,
+                message=(
+                    f"Decryption did not take effect: {env_file} still contains "
+                    "encrypted values. The decryption key may be missing or invalid."
+                ),
+                file_path=env_file,
+            )
+
+        return EncryptionResult(
+            success=True,
+            message=f"Decrypted {env_file}",
+            file_path=env_file,
+        )
 
     def detect_encryption_status(self, value: str) -> EncryptionStatus:
         """
--- a/src/envdrift/encryption/dotenvx.py
+++ b/src/envdrift/encryption/dotenvx.py
@@ -244,6 +244,7 @@ class DotenvxEncryptionBackend(EncryptionBackend):
         if is_file_encrypted(env_file):
             return EncryptionResult(
                 success=False,
+                changed=False,
                 message=(
                     f"Decryption did not take effect: {env_file} still contains "
                     "encrypted values. The decryption key may be missing or invalid."
--- a/src/envdrift/cli_commands/encryption.py
+++ b/src/envdrift/cli_commands/encryption.py
@@ -17,7 +17,11 @@ from envdrift.core.encryption import EncryptionDetector
 from envdrift.core.parser import EnvParser
 from envdrift.core.schema import SchemaLoader, SchemaLoadError
 from envdrift.encryption import EncryptionProvider, get_encryption_backend
-from envdrift.encryption.base import EncryptionBackendError, EncryptionNotFoundError
+from envdrift.encryption.base import (
+    EncryptionBackendError,
+    EncryptionNotFoundError,
+    EncryptionResult,
+)
 from envdrift.output.rich import (
     console,
     print_encryption_report,
@@ -491,6 +495,22 @@ def _verify_decryption_with_vault(
         return False
 
 
+def _report_decrypt_result(result: EncryptionResult, env_file: Path, backend_name: str) -> None:
+    """Print a decrypt outcome and exit nonzero on failure.
+
+    Distinguishes a real decryption from an honest no-op (``changed=False`` — e.g.
+    a file with no encrypted values) so the message is accurate instead of a
+    misleading "Decrypted".
+    """
+    if not result.success:
+        print_error(result.message)
+        raise typer.Exit(code=1)
+    if result.changed:
+        print_success(f"Decrypted {env_file} using {backend_name}")
+    else:
+        print_warning(result.message)
+
+
 def decrypt_cmd(
     env_file: Annotated[Path, typer.Argument(help="Path to encrypted .env file")] = Path(".env"),
     backend: Annotated[
@@ -672,15 +692,7 @@ def decrypt_cmd(
             raise typer.Exit(code=1)
 
         result = encryption_backend.decrypt(env_file)
-        if result.success:
-            if result.changed:
-                print_success(f"Decrypted {env_file} using {encryption_backend.name}")
-            else:
-                # Honest no-op: the file had no encrypted values to decrypt.
-                print_warning(result.message)
-        else:
-            print_error(result.message)
-            raise typer.Exit(code=1)
+        _report_decrypt_result(result, env_file, encryption_backend.name)
 
     except EncryptionNotFoundError as e:
         print_error(str(e))
--- a/src/envdrift/cli_commands/encryption.py
+++ b/src/envdrift/cli_commands/encryption.py
@@ -673,7 +673,11 @@ def decrypt_cmd(
 
         result = encryption_backend.decrypt(env_file)
         if result.success:
-            print_success(f"Decrypted {env_file} using {encryption_backend.name}")
+            if result.changed:
+                print_success(f"Decrypted {env_file} using {encryption_backend.name}")
+            else:
+                # Honest no-op: the file had no encrypted values to decrypt.
+                print_warning(result.message)
         else:
             print_error(result.message)
             raise typer.Exit(code=1)
--- a/src/envdrift/encryption/base.py
+++ b/src/envdrift/encryption/base.py
@@ -35,6 +35,10 @@ class EncryptionResult:
     success: bool
     message: str
     file_path: Path | None = None
+    # False when the operation succeeded but made no change (e.g. decrypting a
+    # file that has no encrypted values). Lets callers report an honest no-op
+    # instead of a misleading "Decrypted".
+    changed: bool = True
 
 
 class EncryptionBackend(ABC):
--- a/tests/integration/test_decrypt_outcome_bench.py
+++ b/tests/integration/test_decrypt_outcome_bench.py
@@ -1,0 +1,124 @@
+"""Real-outcome bench for the dotenvx ``decrypt`` seam (envdrift #443).
+
+Every test drives the REAL decryption seam (``DotenvxEncryptionBackend`` over the
+real ``dotenvx`` binary) and asserts the TRUE on-disk state. The adversarial
+sweep showed the pre-#443 backend reported ``[OK] Decrypted`` unconditionally:
+
+* on a plaintext file it claimed a decryption happened (and dotenvx's line-ending
+  normalisation silently rewrote bytes);
+* on a binary blob it corrupted the file outright while still exiting 0.
+
+The fix reports an honest no-op (``changed=False``) and never invokes dotenvx on
+a file that holds no ciphertext, so a non-encrypted/binary file is left byte-for-
+byte untouched. Tests that need the real binary skip-gate cleanly when absent.
+"""
+
+from __future__ import annotations
+
+import shutil
+from pathlib import Path
+
+import pytest
+
+from envdrift.encryption.base import EncryptionBackendError
+from envdrift.encryption.dotenvx import DotenvxEncryptionBackend
+
+# Mark all tests in this module
+pytestmark = [pytest.mark.integration]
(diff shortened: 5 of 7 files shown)


docs/cli/decrypt.md
```diff
@@ -20,6 +20,14 @@ It can also **verify** that a key stored in your vault can decrypt the file with
 - Viewing encrypted values
 - Migrating to a different encryption system
 
+### Honest no-op on non-encrypted files
+
+If the target file has no encrypted values, `decrypt` reports an honest no-op
+(`Nothing to decrypt: <file> has no encrypted values.`) instead of claiming a
+decryption happened. The file is left byte-for-byte untouched — the backend is
+not invoked at all — so a plaintext file (including one with CRLF line endings)
+or a non-`.env` binary file is never silently rewritten or corrupted.
```
jainal09: 🤖 **envdrift review-bot** · medium · doc-mismatch


The new doc section is written under the generic `decrypt` command Description (the page covers "dotenvx or SOPS"), and states an unconditional guarantee:

> "The file is left byte-for-byte untouched — **the backend is not invoked at all** — so a plaintext file (including one with CRLF line endings) or a non-`.env` binary file is never silently rewritten or corrupted."

This is only true for the dotenvx backend. The honest-no-op pre-check was added only to `DotenvxEncryptionBackend.decrypt` (src/envdrift/encryption/dotenvx.py). `SopsEncryptionBackend.decrypt` (src/envdrift/encryption/sops.py:380-456) has NO `is_file_encrypted` short-circuit — it builds `--decrypt` args and calls `self._run(...)` directly on whatever file it is given. A user can reach this via `envdrift decrypt plaintext.env --backend sops` (decrypt_cmd auto-detect/`--backend` selection at cli_commands/encryption.py:587-605). The PR body itself acknowledges the SOPS path is out of scope ("The binary→UnicodeDecodeError traceback on the auto-detect and SOPS paths (#443 #16/#17) is the separate ... sops theme"), and issue #443 lists the SOPS binary-blob decrypt as an unfixed uncaught-traceback finding. So the doc promises behavior the SOPS backend does not yet implement.

Impact: a reader following docs/cli/decrypt.md will believe `decrypt` (any backend) is a safe no-op on a plaintext/binary file, which is false for `--backend sops`. Suggested direction: scope the guarantee to the dotenvx backend (e.g. "with the dotenvx backend"), or note the SOPS backend does not yet share this guard.

src/envdrift/encryption/dotenvx.py
```diff
@@ -177,13 +194,27 @@ def decrypt(
...
+        # Verify the decryption took effect rather than trusting the exit code:
+        # dotenvx can exit 0 while leaving ciphertext in place when the key is
+        # missing or invalid. If any encrypted value survived, report failure.
+        if is_file_encrypted(env_file):
             return EncryptionResult(
-                success=True,
-                message=f"Decrypted {env_file}",
+                success=False,
+                message=(
+                    f"Decryption did not take effect: {env_file} still contains "
```
jainal09, line 249: 🤖 **envdrift review-bot** · medium · test-gap


The PR's most novel data-integrity addition is the post-decrypt verification: after a 0-exit dotenvx decrypt, it re-runs `is_file_encrypted(env_file)` and, if ciphertext survived, returns `EncryptionResult(success=False, message="Decryption did not take effect: ... still contains encrypted values ...")` (dotenvx.py ~199-211). This is the branch that protects against dotenvx exiting 0 while leaving secrets encrypted (a missing/invalid key that does not error). No test exercises it.

The integration bench's bad-key test (`test_decrypt_with_corrupted_key_errors_and_preserves_ciphertext`) takes the *other* path — dotenvx exits non-zero, raising `DotenvxError` → `EncryptionBackendError` — and asserts `pytest.raises(EncryptionBackendError)`, so it never reaches the `success=False` backstop. Grepping tests for the backstop message ("did not take effect" / "still contains") finds only the explanatory comment in the unit `test_decrypt_success`, not an assertion. Codecov flags this exact line as a partial (dotenvx.py patch 85.71%, 1 partial), and greptile's summary independently noted "no unit test for the exit-0-yet-ciphertext-survives backstop."

Impact: a regression that silently drops the post-check (reverting to trusting the exit code) would not be caught — defeating the contract this PR adds. Suggested direction: add a unit test with a mock `wrapper.decrypt` side_effect that leaves an `encrypted:` value on disk (exit 0), asserting `result.success is False` and the file still holds ciphertext. (This is a distinct test-gap from greptile's inline thread, which only argued the `changed=True` default on this result is semantically misleading.)

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 8df3d5b


The decrypt honest-no-op work is solid for dotenvx, but the guarantee is over-scoped in the docs and the core data-integrity backstop is untested. Both findings posted inline.

**Findings (2):**
- [medium] decrypt.md promises the no-corruption guarantee for all backends, but SOPS has no honest-no-op guard — `docs/cli/decrypt.md:29` (inline)
- [medium] Post-decrypt 'exit 0 but ciphertext survived' backstop branch is untested — `src/envdrift/encryption/dotenvx.py:207` (inline)

jainal09:
Addressed in a5ee408:
- **test-gap + codecov/patch**: added non-integration unit tests for both new backend branches — the no-op return, and the post-decrypt outcome check catching a 0-exit dotenvx that left ciphertext (no-op wrapper mock) — plus a CLI no-op-message test. New lines now covered without the binary.
- **changed flag**: the post-decrypt failure return now sets `changed=False`.
- **complexity**: extracted `_report_decrypt_result()`, cutting the decrypt command's branching.
- **doc-mismatch**: scoped the honest-no-op note to the dotenvx backend.

Full non-integration suite green (2471); decrypt bench green (5).

jainal09:
**envdrift review-bot** (posted via `gh` using Jainal's account) — re-review of latest head `8df3d5ba`


Reviewed the decrypt changes, current checks, existing bot/Greptile threads, and ran focused tests plus byte-level repros.

Focused tests: `uv run --extra dev pytest tests/unit/test_encryption_backends.py tests/integration/test_decrypt_outcome_bench.py -q` -> `84 passed`.

Current checks still failing: `codecov/patch`, `CodeScene Code Health Review (main)`.

**Findings**

1. **[medium] A failed dotenvx decrypt can still mutate encrypted file bytes before returning an error.** `DotenvxWrapper.decrypt()` normalizes line endings at `src/envdrift/integrations/dotenvx.py:953` before running `dotenvx decrypt`; if dotenvx then fails, the original bytes are not restored.

   Repro on this head: encrypt `.env`, convert the encrypted file to CRLF, corrupt `.env.keys`, then run `envdrift decrypt .env --backend dotenvx`. The command exits `1`, but the file hash changes and carriage returns are removed:

   ```text
   exit=1
   before_hash=635c1f6394bbb1485829b96eb8b3546a619938cbbd948950696167d9f630f4bf
   after_hash=97d2a25fa91461933eaaf3a48865705dd4bb9bf8ce175280ee13d52d114c869c
   before_cr_count=8
   after_cr_count=0
   ```

   That violates the PR's “never corrupt” goal for failed decrypt attempts. The wrapper/backend should preserve original bytes and restore them on subprocess failure and on the post-decrypt “ciphertext survived” failure path.

2. **[medium] Existing docs over-scope the no-op guarantee.** The current `docs/cli/decrypt.md` text says plaintext and binary files are never rewritten and that the backend is not invoked at all, but the implemented guard is dotenvx-specific. SOPS still has no equivalent no-op guard, and auto-detect can still fail before reaching the dotenvx backend for binary input. This matches the earlier inline bot comment and remains open on this head.

3. **[medium] Existing backstop branch is still untested.** `src/envdrift/encryption/dotenvx.py:203-211` handles “dotenvx exited 0 but ciphertext survived,” but the new corrupted-key integration test takes the `DotenvxError` path instead. Codecov is also flagging this area. Add a mock/fixture where `wrapper.decrypt()` returns without changing `KEY=encrypted:...` so the `success=False` result is asserted.

4. **[low] The post-check failure result reports `changed=True` by default.** The `EncryptionResult(success=False, ...)` at `src/envdrift/encryption/dotenvx.py:204` omits `changed=False`, so callers can observe a failed no-op/backstop as changed.

5. **[low] Plaintext no-op still requires dotenvx to be installed at the CLI layer.** `src/envdrift/cli_commands/encryption.py:669-672` checks `is_installed()` before calling `backend.decrypt()`, so with `PATH` excluding dotenvx, `envdrift decrypt plaintext.env --backend dotenvx --ci` exits `1` with “dotenvx is not installed” instead of the new “Nothing to decrypt” no-op. It does not corrupt data, but it weakens the “backend is not invoked” guarantee.

jainal09:
**envdrift review-bot** (posted via `gh` using Jainal's account) — current-head re-review of `a5ee4080`


The latest commit fixed several earlier items: docs are now scoped to dotenvx, the post-decrypt backstop has unit coverage, the backstop result sets `changed=False`, and Codecov is green.

Focused local checks on this head:

```sh
uv run --extra dev pytest tests/unit/test_encryption_backends.py tests/integration/test_decrypt_outcome_bench.py tests/unit/coverage/test_cov_cli_commands_encryption.py -q
# 111 passed
```

**Remaining findings**

1. **[medium] Failed dotenvx decrypt still mutates encrypted file bytes before returning an error.** `DotenvxWrapper.decrypt()` still normalizes line endings at `src/envdrift/integrations/dotenvx.py:953` before running the subprocess. If dotenvx then fails, the original bytes are not restored. Repro on `a5ee4080`: encrypt `.env`, convert the encrypted file to CRLF, corrupt `.env.keys`, then run `envdrift decrypt .env --backend dotenvx`. The command exits `1`, but bytes changed:

   ```text
   exit=1
   before_hash=f23350a77e162f3b492dc5da93d55853c7b0fd76e5f8fb31f1f28a2c359b0bdc
   after_hash=e3adb1157909d6d808ab6c1c368e178c7672966440e35982af84edb40363801f
   before_cr_count=8
   after_cr_count=0
   ```

   This still violates the PR's “never corrupt” goal for failed decrypt attempts. Preserve original bytes and restore them on `DotenvxError` and on the post-check failure path.

2. **[low] Plaintext no-op still requires dotenvx to be installed at the CLI layer.** The backend no-op pre-check runs before `DotenvxEncryptionBackend.is_installed()`, but the CLI checks install status first at `src/envdrift/cli_commands/encryption.py:689-692`. With `PATH` excluding dotenvx, `envdrift decrypt plaintext.env --backend dotenvx --ci` exits `1` with “dotenvx is not installed” instead of the new “Nothing to decrypt” no-op. This does not mutate data, but it contradicts the “dotenvx is not invoked” no-op contract at the command surface.

3. **[medium] Current CodeScene check is still failing.** The latest CodeScene review reports failed gates for `Low Cohesion`, plus advisory `Overall Code Complexity`, `Missing Arguments Abstractions`, `Lines of Code in a Single File`, and `Number of Functions in a Single Module`. GitHub Actions and Codecov are green; CodeScene is the remaining failing check.

docs/cli/decrypt.md
jainal09: 🤖 **envdrift review-bot** — resolved ✅

Verified at `a5ee408`: `docs/cli/decrypt.md` now accurately describes the non-encrypted no-op / sops behavior without overpromising. Resolved, thanks.

src/envdrift/encryption/dotenvx.py
jainal09: 🤖 **envdrift review-bot** — resolved ✅

Verified at `a5ee408`: the `success=False` post-decrypt backstop branch now has a load-bearing test (fails without the guard), and the changed-flag default was corrected without breaking the honest no-op. Resolved.

jainal09:
Re the CodeScene delta flags: these are **advisory** code-health deltas (CodeScene is non-required), and they come directly from the *same* review round's requests:
- **test-file size / function-count / cohesion** (`test_encryption_backends.py`, `test_cov_cli_commands_encryption.py`): the added regression + branch-coverage tests that closed the test-gap and the codecov/patch fail necessarily grow these modules; coverage-focused test files are intentionally broad.
- **encryption.py overall-complexity / arguments**: this commit *reduced* the decrypt command's branching by extracting `_report_decrypt_result`; the module-level delta is the artifact of adding one small, well-named helper (3 args, net-simpler call site).

No correctness impact and net readability is better, so resolving as accepted trade-offs.

jainal09:
The two new CodeScene flags (`test_encryption_backends.py` LOC=620, and 6 similarly-structured backend test functions) are the same advisory category as before: the rebase onto main combined #444's and this PR's backend tests into one module, and encryption-backend tests share an unavoidable shape (patch wrapper → call → assert on-disk state). Advisory + non-required; resolving as accepted.

PR #33, opened 2026-01-02
feat: Add partial encryption feature

jainal09:
## Summary

Add partial encryption feature for separating cleartext (non-sensitive) and encrypted (sensitive) environment variables while maintaining a single combined file for applications.

## Features

- **New commands:**
  - `envdrift push` - Encrypt secret files + combine with clear files
  - `envdrift pull-partial` - Decrypt secret files for editing

- **Configuration:** New `[partial_encryption]` section in `envdrift.toml`

## Workflow

```bash
# Edit source files
vim .env.production.clear    # Non-sensitive
vim .env.production.secret   # Sensitive (decrypted)

# Before commit
envdrift push

# Commit all 3 files
git add .env.production.clear .env.production.secret .env.production
```

## Files Added
- src/envdrift/core/partial_encryption.py
- src/envdrift/cli_commands/partial.py
- docs/guides/partial-encryption.md
- examples/partial-encryption-example.md
- tests/unit/test_partial_encryption.py

## Tests
- All 462 tests pass
- Pre-commit checks pass

## Summary by CodeRabbit

* **New Features**
  * Partial encryption for environment variables: keep cleartext and encrypted secrets in separate files
  * New CLI commands: push (encrypt & generate combined env files) and pull-partial (decrypt secrets for editing)

* **Documentation**
  * Added a comprehensive guide and a step-by-step example demonstrating setup, workflows, migration tips, and Git recommendations

* **Tests**
  * New unit tests covering partial encryption behaviors, file combinations, markers, and edge cases

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #449, opened 2026-06-09
fix(init/validate): handle non-identifier & Unicode env keys end-to-end

jainal09:
Supersedes #446. Closes the init→validate round-trip for keys the strict parser rejects, so the documented workflow is correct end-to-end. (#446 fixed only the init side; CodeRabbit correctly flagged that aliasing keys into the schema without teaching `validate` to match them makes init→validate fail.)

## The problem (verified)

The adversarial sweep (#443) found `init` silently **drops** every `.env` key the strict `EnvParser` rejects (dash/digit/dot/emoji + non-ASCII like `CAFÉ`). Fixing init to **alias** them in surfaced a second bug: `validate` parsed the .env strictly (dropping those keys) and matched by attribute name, so:

```
$ envdrift init .env -o settings.py -c Cfg   # X-API-KEY -> Field(alias='X-API-KEY')
$ envdrift validate .env --schema settings:Cfg
Validation FAILED — MISSING REQUIRED VARIABLES: X_API_KEY   # but X-API-KEY IS in the file
```

## The fix (init + validate, one theme)

| Layer | Change |
|---|---|
| `EnvParser` | opt-in **`lenient=True`** (`LENIENT_LINE_PATTERN`) recovers non-identifier / non-ASCII keys. Every other caller keeps the strict pattern. |
| `init` | parses `lenient=True` and aliases the recovered keys; drops its private `_unparsed_assignments` (now redundant — single-source). |
| `FieldMetadata` | carries the field **`alias`** (from `validation_alias`/`alias`). |
| `validate` | parses `lenient=True`; the validator matches each field against the .env **by alias** (when present) for the missing / extra / encryption / type checks — mirroring pydantic-settings. |

Result:

```
$ envdrift validate .env --schema settings:Cfg
Validation PASSED
```

## Tests / docs

- parser: strict drops vs `lenient=True` recovers non-identifier + Unicode keys
- schema: `extract_metadata` captures the alias
- validator: an aliased field present under its alias validates (and is reported missing when absent)
- **end-to-end `init`→`validate` round-trip** over `X-API-KEY` / `2FA_ENABLED` / `CAFÉ` that **PASSES** (failed before this change)

Full non-integration suite green (2479); ruff/format/pyrefly/bandit/markdownlint clean. Docs synced.

Scope note: this is the alias/round-trip slice of validate-fidelity. The remaining `validate` findings — constraint enforcement (#443 #12), broken-schema-module traceback (#34), multiline values (#33) — stay as follow-ups.

Refs #443

## Summary by CodeRabbit

* **New Features**
  * Lenient `.env` parsing retains non-standard keys (non-identifiers, non-ASCII) and emits fields with sanitized attribute names plus Pydantic aliases so every env var is captured and generated modules remain importable.

* **Bug Fixes**
  * Removed warnings about skipped/unparsed keys; aliasing now preserves those variables.
  * Generated settings files are written using explicit UTF-8 encoding.

* **Documentation**
  * Updated examples and docs to show aliased field handling for non-identifier environment variables.

<h3>Greptile Summary</h3>

This PR fixes the `init`→`validate` round-trip for non-identifier and Unicode `.env` keys by adding a `lenient=True` parsing mode to `EnvParser` and threading alias awareness through `FieldMetadata` and the `Validator`. Keys that were previously dropped (e.g. `X-API-KEY`, `2FA_ENABLED`, `CAFÉ`) are now aliased into the generated schema and matched back by alias during validation.

- **Parser**: adds `LENIENT_LINE_PATTERN` and a `lenient` keyword arg to `parse`/`parse_string`; `value_from_raw` is extracted as a public helper for consistent value normalization.
- **`init`**: switches to `lenient=True`, replaces the dropped-key warning path with NFKC-aware collision tracking in `_resolve_field_name`, and writes output as UTF-8.
- **`validate`**: parses leniently; the validator now resolves each field's lookup key via `field_meta.alias or field_meta.name`, mirroring how pydantic-settings binds env vars to aliased fields.

Safe to merge for the init→validate round-trip fix; one remaining defect in `generate_fix_template` means `--fix` output writes the wrong key name for aliased fields.

The core alias matching in `validate` and the lenient parser are correct and well-tested. However, `generate_fix_template` in `validator.py` (lines 325–346) still emits `var_name` (the attribute name like `X_API_KEY`) as the `.env` key rather than `field_meta.alias or var_name` (`X-API-KEY`). A user who runs `envdrift validate --fix` on a file with a missing aliased field will receive a template that instructs them to add the wrong key, and validation will still fail after they follow the advice. The `alias` field added to `FieldMetadata` in this PR is everything needed to fix it — only the two emission lines in `generate_fix_template` were not updated.

src/envdrift/core/validator.py — the `generate_fix_template` method needs its key-emission lines updated to use the alias.

| Filename | Overview |
|----------|----------|
| src/envdrift/core/validator.py | Validation matching updated to use alias for field lookup — correct. `generate_fix_template` still emits the attribute name instead of the alias as the .env key, so `--fix` output is wrong for aliased fields. |
| src/envdrift/core/parser.py | Adds `lenient=True` opt-in mode via `LENIENT_LINE_PATTERN` and exposes `value_from_raw` as a public helper; default strict behavior is unchanged. |
| src/envdrift/core/schema.py | Adds `alias` field to `FieldMetadata`, populated from `validation_alias` (with precedence) or `alias` during `extract_metadata`; clean and correct. |
| src/envdrift/cli_commands/init_cmd.py | Switches to `lenient=True` parsing, drops `_unparsed_keys`/`_raw_assignment_keys`, adds NFKC collision tracking in `_resolve_field_name`, and writes output as UTF-8; logic is solid. |
| src/envdrift/api.py | Removes the `UserWarning` emission for non-identifier keys and adds `encoding="utf-8"` to `write_text`; clean. |
| tests/unit/test_validator.py | New `TestValidatorAliasMatching` class covers alias extraction, aliased-present, and aliased-absent cases; good coverage of the new logic. |
| tests/unit/test_cli.py | Adds five new test cases including NFKC collision, unicode identifier, and end-to-end init→validate round-trip; comprehensive coverage of the new behavior. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant init_cmd
    participant EnvParser
    participant FieldResolver
    participant validate_cmd
    participant Validator

    User->>init_cmd: envdrift init .env -o settings.py
    init_cmd->>EnvParser: "parse(.env, lenient=True)"
    EnvParser-->>init_cmd: "EnvFile {X-API-KEY, 2FA_ENABLED, CAFÉ, ...}"
    init_cmd->>FieldResolver: _resolve_field_name(X-API-KEY, used_folded)
    FieldResolver-->>init_cmd: "(X_API_KEY, alias=X-API-KEY)"
    init_cmd->>FieldResolver: _resolve_field_name(2FA_ENABLED, used_folded)
    FieldResolver-->>init_cmd: "(field_2FA_ENABLED, alias=2FA_ENABLED)"
    init_cmd-->>User: settings.py written (UTF-8)

    User->>validate_cmd: envdrift validate .env --schema settings:Cfg
    validate_cmd->>EnvParser: "parse(.env, lenient=True)"
    EnvParser-->>validate_cmd: "EnvFile {X-API-KEY, 2FA_ENABLED, CAFÉ, ...}"
    validate_cmd->>Validator: validate(env, schema_meta)
    Note over Validator: _lookup_key(fm) = fm.alias or fm.name
    Validator-->>validate_cmd: "ValidationResult(valid=True)"
    validate_cmd-->>User: Validation PASSED
```

<details><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `src/envdrift/core/validator.py`, line 322-333 ([link](https://github.com/jainal09/envdrift/blob/3884d31a6259d2a67414f090c70ceb4550e315be/src/envdrift/core/validator.py#L322-L333)) 

   <a href="#">P1</a> **`generate_fix_template` emits the attribute name, not the real env-var name**

   `var_name` here is the Pydantic attribute name (e.g., `X_API_KEY`). For a field carrying `alias='X-API-KEY'` — exactly the kind of field this PR now adds to the schema — the template outputs `X_API_KEY=`, but the `.env` file expects `X-API-KEY=`. A user running `envdrift validate --fix` on a file missing that key would be told to add the wrong key and validation would still fail.

   The `alias` field added to `FieldMetadata` in this PR should be used here: replace `var_name` with `field_meta.alias or var_name` (or `var_name` when `field_meta` is None) as the key written to the template.

2. `src/envdrift/core/validator.py`, line 323-346 ([link](https://github.com/jainal09/envdrift/blob/0821ff0385b580e63c1d3822882af3b91ab03a81/src/envdrift/core/validator.py#L323-L346)) 

   <a href="#">P1</a> **`generate_fix_template` emits attribute name, not the real env-var key**

   `var_name` here is the Pydantic attribute name (e.g. `X_API_KEY`). For a field carrying `alias='X-API-KEY'` — exactly the kind of field this PR adds — the template writes `X_API_KEY=` in both the required and optional sections, but the `.env` file must use `X-API-KEY=`. A user running `envdrift validate --fix` on a file that's missing that key would be instructed to add the wrong key name, and validation would still fail after they follow the advice.

   Both loops should replace the bare `var_name` with `field_meta.alias or var_name` (falling back to `var_name` when `field_meta` is `None` or the alias is unset).

3. `src/envdrift/core/validator.py`, line 323-348 ([link](https://github.com/jainal09/envdrift/blob/5214c631a56a49990f755c7360cdcf518f4705a6/src/envdrift/core/validator.py#L323-L348)) 

   <a href="#">P1</a> **`generate_fix_template` emits the attribute name instead of the real env-var key**

   `var_name` here is the Pydantic attribute name (e.g. `X_API_KEY`). For a field carrying `alias='X-API-KEY'` — exactly what this PR now generates from `init` — the template writes `X_API_KEY=` in both the required and optional sections. A user running `envdrift validate --fix` on a file missing that key would be told to add `X_API_KEY=secret`, but validation would still fail because the `.env` file must contain `X-API-KEY=secret`. The `alias` field added to `FieldMetadata` in this PR is the fix: replace `var_name` in the key-emission lines with `(field_meta.alias or var_name) if field_meta else var_name`.

</details>

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into feat/init-valid..."](https://github.com/jainal09/envdrift/commit/5214c631a56a49990f755c7360cdcf518f4705a6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36471993)</sub>

### DIFF
--- a/src/envdrift/cli_commands/init_cmd.py
+++ b/src/envdrift/cli_commands/init_cmd.py
@@ -102,8 +102,7 @@ def _unparsed_assignments(env_file: Path, env: EnvFile) -> dict[str, str]:
         key = match.group(1)
         if key in env.variables or key in extra:
             continue
-        value = parser._unquote(parser._strip_inline_comment(match.group(2)).strip())
-        extra[key] = value
+        extra[key] = parser.value_from_raw(match.group(2))
     return extra
 
 
--- a/src/envdrift/cli_commands/init_cmd.py
+++ b/src/envdrift/cli_commands/init_cmd.py
@@ -14,12 +14,12 @@ from pydantic_settings import BaseSettings
 
 from envdrift.core.encryption import EncryptionDetector
 from envdrift.core.parser import EnvFile, EnvParser
-from envdrift.output.rich import console, print_error, print_success, print_warning
+from envdrift.output.rich import console, print_error, print_success
 
-# Matches the left-hand side of any `KEY=value` assignment in a raw .env file,
-# accepting keys the strict EnvParser.LINE_PATTERN rejects (leading digits,
-# dashes, etc.) so init can warn about variables it would otherwise drop.
-_RAW_ASSIGNMENT_PATTERN = re.compile(r"^\s*(?:export\s+)?([^\s=#]+)\s*=")
+# Matches `KEY=value` for ANY key the strict EnvParser.LINE_PATTERN rejects
+# (leading digits, dashes, dots, non-ASCII letters, …) so init can still emit a
+# field for it — sanitized + aliased — instead of silently dropping it.
+_RAW_ASSIGNMENT_PATTERN = re.compile(r"^\s*(?:export\s+)?([^\s=#]+)\s*=(.*)$")
 
 # Pydantic protects the ``model_`` attribute namespace: a BaseModel/BaseSettings
 # field whose name starts with this prefix either raises at import (``model_dump``
@@ -57,8 +57,6 @@ class SettingsGeneration:
     source: str
     sensitive_vars: set[str] = field(default_factory=set)
     aliased_count: int = 0
-    # .env keys present in the raw text but rejected by the strict parser.
-    unparsed_keys: list[str] = field(default_factory=list)
 
 
 def _sanitize_identifier(name: str) -> str:
@@ -82,17 +80,31 @@ def _sanitize_identifier(name: str) -> str:
     return sanitized
 
 
-def _raw_assignment_keys(text: str) -> list[str]:
-    """Extract every assignment key from raw .env text (parser-independent)."""
-    keys: list[str] = []
-    for raw_line in text.splitlines():
+def _unparsed_assignments(env_file: Path, env: EnvFile) -> dict[str, str]:
+    """Return ``name -> value`` for assignment lines the strict parser rejected.
+
+    ``EnvParser`` only accepts identifier-style ASCII keys, so ``2FA_ENABLED``,
+    ``X-API-KEY`` or a non-ASCII identifier such as ``CAFÉ`` never become
+    variables. Those keys are recovered here — values are unquoted with the
+    parser's own helpers so the value handling stays single-source — and fed
+    through the same sanitize/alias path as parsed keys instead of being dropped.
+    Keys the parser already accepted are skipped (they come from ``env``).
+    """
+    parser = EnvParser()
+    extra: dict[str, str] = {}
+    for raw_line in env_file.read_text(encoding="utf-8").splitlines():
         line = raw_line.strip()
         if not line or line.startswith("#"):
             continue
         match = _RAW_ASSIGNMENT_PATTERN.match(line)
-        if match:
-            keys.append(match.group(1))
-    return keys
+        if not match:
+            continue
+        key = match.group(1)
+        if key in env.variables or key in extra:
+            continue
+        value = parser._unquote(parser._strip_inline_comment(match.group(2)).strip())
+        extra[key] = value
+    return extra
 
 
 def _needs_sanitizing(name: str) -> bool:
@@ -210,43 +222,31 @@ def _module_header(class_name: str, env_file: Path) -> list[str]:
     ]
 
 
-def _unparsed_keys(env_file: Path, env: EnvFile) -> list[str]:
-    """Return raw .env keys present in the file but rejected by the strict parser.
-
-    The ``EnvParser`` only accepts identifier-style keys, so ``2FA_ENABLED`` or
-    ``MY-DASH-VAR`` never become variables. These are surfaced to the caller to
-    warn about rather than silently drop (``validate`` would later reject them as
-    forbidden extras under ``extra="forbid"``).
-    """
-    raw_keys = _raw_assignment_keys(env_file.read_text(encoding="utf-8"))
-    return [k for k in dict.fromkeys(raw_keys) if k not in env.variables]
-
-
-def _detect_sensitive_vars(env: EnvFile) -> set[str]:
+def _detect_sensitive_vars(values: dict[str, str]) -> set[str]:
     """Return the set of variable names flagged sensitive by name or value."""
     detector = EncryptionDetector()
     return {
         name
-        for name, var in env.variables.items()
-        if detector.is_name_sensitive(name) or detector.is_value_suspicious(var.value)
+        for name, value in values.items()
+        if detector.is_name_sensitive(name) or detector.is_value_suspicious(value)
     }
 
 
-def _render_fields(env: EnvFile, sensitive_vars: set[str]) -> tuple[list[str], int]:
+def _render_fields(values: dict[str, str], sensitive_vars: set[str]) -> tuple[list[str], int]:
     """Render every Settings field line; return the lines and the alias count.
 
     Walks the variables in sorted order so output is deterministic, resolving each
-    to a unique importable attribute name (with an alias when renamed) and an
-    inferred type/default.
+    to a unique importable attribute name (with an alias when the .env key is a
+    non-identifier, keyword, or collides) and an inferred type/default.
     """
     field_lines: list[str] = []
     aliased_count = 0
     used_names: set[str] = set()
-    for var_name, env_var in sorted(env.variables.items()):
+    for var_name, value in sorted(values.items()):
         field_name, alias = _resolve_field_name(var_name, used_names)
         if alias is not None:
             aliased_count += 1
-        type_hint, default_val = _infer_type(env_var.value)
+        type_hint, default_val = _infer_type(value)
         field_lines.append(
             _render_field_line(
                 field_name, type_hint, default_val, alias, var_name in sensitive_vars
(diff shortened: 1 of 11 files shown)


src/envdrift/cli_commands/init_cmd.py
```diff
@@ -210,43 +190,31 @@ def _module_header(class_name: str, env_file: Path) -> list[str]:
     ]
```
jainal09, line 198: 🤖 **envdrift review-bot** · high · bug


`_resolve_field_name`/`_needs_sanitizing` classify a key as a bare field whenever `str.isidentifier()` is True, and collision avoidance (`_bump_until_unique`) only tracks the *pre-compile* attribute names. But Python normalizes identifiers with **NFKC at compile time**, so two distinct env keys that fold to the same identifier collapse into one attribute when the generated module is imported.

Reproduced under the repo venv:
- NFC `CAFÉ` (É = U+00C9) and NFD `CAFÉ` (E + U+0301): both pass `isidentifier()`, both emitted as bare fields, generated `Cfg.model_fields` has **1** entry, not 2.
- `ﬁle` (U+FB01 ligature, NFKC→`file`) and plain `file`: both emitted bare, `model_fields` collapses to a single `file`.

The env parser keys `env.variables` by the raw string, so both keys survive parsing and `_render_fields` (iterating `sorted(values.items())`) emits **two** field lines — but only one survives the import. Impact: (1) one env var is silently dropped from the schema, directly contradicting the docstring's guarantee `"complete (no key dropped)"` / `"Every .env key becomes a field"` (init_cmd.py:236-238) and focus item (c) that collisions must be "handled (not silently merged)"; (2) at `validate`, the collapsed-away key has no schema field and is reported as an `extra_var` under `extra="forbid"`, a false validation FAILURE — breaking the very init→validate round-trip this PR exists to fix. No `unicodedata.normalize` exists anywhere in `src/` (grep confirms), so the focus's claim of deterministic NFC normalization is not implemented.

Suggested direction: detect post-fold collisions (e.g. compare against the set of already-emitted *NFKC-normalized* field names, or treat a key whose NFKC form differs from itself or collides with another emitted field as needing sanitize+alias+bump like any other collision), and/or normalize keys deterministically. Then add a real regression test (NFC-vs-NFD and ligature-vs-ASCII) asserting both keys end up as distinct fields with aliases — it should fail on current code.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 3884d31


This PR's init→validate round-trip for non-identifier keys has a real silent data-loss bug for Unicode keys that NFKC-fold to the same identifier, and the test suite lacks a regression test that would catch it.

**Findings (2):**
- [high] Unicode keys that NFKC-fold to the same identifier are silently merged into one field (key dropped) — `src/envdrift/cli_commands/init_cmd.py:192` (inline)
- [medium] No regression test for Unicode/NFKC field-name collisions; round-trip test only uses fold-distinct keys — `tests/unit/test_cli.py` (see below — could not anchor inline; target line is outside the PR diff hunk)

### Not anchored inline

**[medium · test-gap] No regression test for Unicode/NFKC field-name collisions; round-trip test only uses fold-distinct keys** — `tests/unit/test_cli.py` (around `test_init_unicode_identifier_keys_become_bare_fields`)

The PR adds tests for ASCII keyword collision (`test_init_keyword_collision_keeps_both_env_bindings`, class vs class_) and for valid Unicode identifiers becoming bare fields (`test_init_unicode_identifier_keys_become_bare_fields` using `CAFÉ`/`ΑΛΦΑ`/`KEY🔑`), but none of them cover **two keys that fold to the same identifier** (NFC vs NFD `CAFÉ`, or `ﬁle` vs `file`). The end-to-end round-trip test (`test_init_then_validate_round_trip_with_non_identifier_keys`) uses only `DATABASE_URL`/`2FA_ENABLED`/`X-API-KEY`/`CAFÉ`, all distinct after NFKC folding, so it passes even though the silent-merge bug is present. Per CLAUDE.md, the collision behavior the PR claims to handle needs a regression test that fails without the fix.

Suggested direction: add an init test writing e.g. `CAFÉ=a\nCAFÉ=b\n` (one NFC, one NFD) or `ﬁle=a\nfile=b\n` and assert the imported `Cfg.model_fields` contains two distinct fields each bound to its original key (and/or that init surfaces the collision rather than dropping a key). On current code this assertion fails (model_fields collapses to one), making it a valid fail-without-fix regression test.

jainal09:
Fixed the HIGH NFKC-collision bug in 0821ff0 — excellent catch. Identifier name tracking now uses the **NFKC-folded** form (the shape Python actually compiles to), so NFC `CAFÉ` vs NFD `CAFÉ` and ligature `ﬁle` vs `file` bump to distinct attributes + alias back to their exact original keys instead of collapsing. Regression test asserts all 4 survive as 4 distinct fields (collapses to 2 on the old code). The CodeScene 'Complex Method' delta is advisory (the resolver gained the fold-aware bump loop) — net behaviour is correct; resolving as accepted.

src/envdrift/cli_commands/init_cmd.py
jainal09: 🤖 **envdrift review-bot** — resolved ✅

Verified at `0821ff0`: init now tracks NFKC-folded names, so two distinct source keys that fold to the same identifier are handled deterministically (no silent merge / secret drop) and the originals round-trip. The added unicode-collision test is load-bearing (fails without the fix), which also closes the test-gap finding. Resolved, thanks.

jainal09:
The CodeScene 'Complex Conditional' is the NFKC alias guard (`sanitize or bumped or _nfkc(field_name) != var_name`) — it's a flat 3-term disjunction expressing one idea ('the attribute Python actually binds differs from the original key'), each term commented. Advisory; the correctness (proven by the NFC/NFD/ligature regression test) outweighs the metric. Resolving as accepted.

jainal09:
@coderabbitai review

PR #451, opened 2026-06-09
ci: run binary-backed integration tests on macOS and Windows

jainal09:
The enabler for your goal: **real cross-platform test coverage of the binary-backed features**.

## Why

The cross-platform matrix (#442) only runs `-m "not integration"`. So the features that drive real binaries — **encrypt, decrypt, diff, partial-encryption, sops, and the secret scanners** — have **zero** cross-platform coverage today; their integration tests run only on Linux. This makes them run on macOS + Windows too.

## What

- **`.github/scripts/install_test_binaries.py`** — a standalone, cross-platform installer. Reads the pinned versions/URLs from `constants.json` (single source, no hardcoded versions), installs **dotenvx, sops, gitleaks, trufflehog, talisman, trivy, infisical** into a bin dir on PATH. Resilient (a tool that fails to install is logged + skipped so the rest run). **Verified locally on macOS — all 7 install and report `--version`.**
- **`cross-platform-integration.yml`** — runs the non-container integration suite (`-m "integration and not aws and not vault and not azure and not gcp"`) on `macos-latest` + `windows-latest`. Container-backed tests (aws/vault/azure) stay Linux-only by infra necessity.

## Method (deliberate)

**Not a required check.** It is *expected to fail initially* — each failing test here is a **real cross-platform feature bug to fix**, not something to skip or xfail. Keeping it non-required lets those honest failures surface as findings **without blocking the merge queue**. The follow-up work fixes the tool until this is green — and that green *is* the reliability goal.

No spoofing (real binaries, real CLI, no mocks), no cheating (no skip-to-hide), no save-for-later.

Refs #443, #445

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UTF‑8 handling across the CLI and subprocess interactions to avoid encoding errors on Windows and when external tools emit non‑UTF8 bytes.

* **Chores**
  * Added cross‑platform integration test workflow and a helper to install required test binaries automatically on CI runners.

* **Tests**
  * Updated integration and unit tests to ensure deterministic UTF‑8 behavior and to validate in‑memory decryption and environment injection.

<h3>Greptile Summary</h3>

This PR adds a cross-platform integration test workflow (`cross-platform-integration.yml`) that runs binary-backed tests on macOS and Windows, paired with a new Python installer script that downloads pinned tool binaries at CI time. It also applies a broad UTF-8 hardening pass across all `subprocess.run` calls in the scanner, encryption, and CLI-install paths — guarding against Windows cp1252 mis-decoding.

- **New CI workflow**: runs `pytest -m \"integration and not aws/vault/azure/gcp\"` on `macos-latest` and `windows-latest` with per-test `--timeout=300`; intentionally non-required so initial failures surface as real cross-platform bug findings rather than blocking merges.
- **`install_test_binaries.py`**: standalone installer that reads pinned versions from `constants.json`, downloads tarballs/zips/raw binaries for 7 tools, validates archives against zip-slip before extraction, and appends the bin dir to `GITHUB_PATH`; resilient per-tool so a single failed install doesn't abort the rest.
- **`sops.py` `exec_env` rewrite**: replaces `sops exec-env` (shell-expansion-fragile on Windows) with a `sops -d` decrypt-to-memory step followed by a direct `subprocess.run(command, env=...)` call — no shell, no disk write, cross-platform quoting safe.

Safe to merge; changes are additive CI infrastructure and defensive UTF-8 encoding fixes with no production logic regressions.

All production-path changes are strictly additive: encoding utf-8 and errors replace on subprocess calls is a net-correctness improvement with no behavior change on POSIX. The exec_env rewrite is well-tested with a real Python child process. The one outstanding item — no timeout on urlretrieve in the CI installer — can only delay a CI step; it has no effect on the production library.

.github/scripts/install_test_binaries.py: the _download function uses urlretrieve without a timeout — a stalled binary download would silently consume the runner's wall-clock budget instead of surfacing as a timed-out per-tool failure.

| Filename | Overview |
|----------|----------|
| .github/scripts/install_test_binaries.py | New cross-platform binary installer for CI; resilient per-tool failure handling and zip/tar-slip guards are solid, but urlretrieve lacks a timeout so a stalled download blocks the whole step. |
| .github/workflows/cross-platform-integration.yml | New macOS+Windows integration workflow; --timeout=300 is present, fail-fast disabled, container markers correctly excluded. |
| src/envdrift/encryption/sops.py | exec_env rewritten to decrypt via sops -d to memory and run the child as argv (no shell); encoding=utf-8/errors=replace and a 300s timeout are all present. |
| tests/integration/conftest.py | Added autouse _utf8_subprocess_on_windows fixture that patches subprocess.run/Popen on Windows to inject encoding=utf-8; correctly no-ops on POSIX. |
| tests/unit/test_encryption_backends.py | Tests updated to validate new decrypt-to-memory exec_env; new tests cover UTF-8 round-trip, missing file, bad decrypt, caller env merging. |
| src/envdrift/cli.py | Added _force_utf8_output() that reconfigures stdout/stderr to UTF-8 on Windows at import time; POSIX is untouched. |
| src/envdrift/integrations/dotenvx.py | Added encoding=utf-8/errors=replace to subprocess.run calls in installer and wrapper; no logic changes. |
| src/envdrift/scanner/gitleaks.py | Added encoding=utf-8/errors=replace to all subprocess.run calls; no logic changes. |
| src/envdrift/cli_commands/install.py | Added encoding=utf-8/errors=replace to agent install/status subprocess calls; no logic changes. |

</details>

<sub>Reviews (10): Last reviewed commit: ["test(sops): cover exec\_env error/edge br..."](https://github.com/jainal09/envdrift/commit/458f90ae1eeb2525b107393ae1af04c8277d74fb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36480316)</sub>

### DIFF
--- a/.github/scripts/install_test_binaries.py
+++ b/.github/scripts/install_test_binaries.py
@@ -78,6 +78,34 @@ def _find_binary(root: Path, binary: str) -> Path:
     raise FileNotFoundError(f"{binary!r} not found in extracted archive under {root}")
 
 
+def _is_within(base: Path, target: Path) -> bool:
+    """True if resolved ``target`` stays inside ``base`` (no path traversal)."""
+    try:
+        target.resolve().relative_to(base.resolve())
+    except ValueError:
+        return False
+    return True
+
+
+def _safe_extract_tar(tf: tarfile.TarFile, dest: Path) -> None:
+    """Extract a tar, rejecting any member that escapes ``dest`` (zip/tar-slip)."""
+    for member in tf.getmembers():
+        if member.islnk() or member.issym() or not _is_within(dest, dest / member.name):
+            raise RuntimeError(f"unsafe path in archive: {member.name!r}")
+    try:
+        tf.extractall(dest, filter="data")  # 3.12+: blocks traversal/abs/special
+    except TypeError:  # Python < 3.12 has no filter= kwarg
+        tf.extractall(dest)  # members already validated above
+
+
+def _safe_extract_zip(zf: zipfile.ZipFile, dest: Path) -> None:
+    """Extract a zip, rejecting any member that escapes ``dest``."""
+    for name in zf.namelist():
+        if not _is_within(dest, dest / name):
+            raise RuntimeError(f"unsafe path in archive: {name!r}")
+    zf.extractall(dest)
+
+
 def _install_tool(bindir: Path, binary: str, urls_key: str, version_key: str) -> None:
     urls = CONST[urls_key]
     key = _platform_key(urls)
@@ -92,13 +120,13 @@ def _install_tool(bindir: Path, binary: str, urls_key: str, version_key: str) ->
         _download(url, archive)
         extract_dir = bindir / f"_{binary}_x"
         with tarfile.open(archive) as tf:
-            tf.extractall(extract_dir)
+            _safe_extract_tar(tf, extract_dir)
         dest.write_bytes(_find_binary(extract_dir, binary).read_bytes())
     elif url.endswith(".zip"):
         _download(url, archive)
         extract_dir = bindir / f"_{binary}_x"
         with zipfile.ZipFile(archive) as zf:
-            zf.extractall(extract_dir)
+            _safe_extract_zip(zf, extract_dir)
         dest.write_bytes(_find_binary(extract_dir, binary).read_bytes())
     else:
         # Raw, single-file binary (e.g. sops, talisman).
--- a/.github/scripts/install_test_binaries.py
+++ b/.github/scripts/install_test_binaries.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python3
+"""Install the external CLI binaries the integration tests drive, cross-platform.
+
+Reads pinned versions + download URLs from ``src/envdrift/constants.json`` (the
+single source of truth, kept current by Renovate — no hardcoded versions here)
+and installs each tool into a per-user bin directory, which it appends to
+``GITHUB_PATH`` so later steps and the test subprocesses find them on ``PATH``.
+
+Works on Linux, macOS and Windows. Standalone (no ``envdrift`` import) so the
+production installer's own cross-platform behaviour is exercised by *its* tests,
+not entangled with the feature tests this enables.
+
+Resilient by design: each tool installs independently and a failure is logged
+and skipped (it becomes a real finding — the test that needs it then skips for an
+honest "binary absent" reason) rather than aborting every other tool's install.
+Exit code is non-zero only if *nothing* installed, so the workflow surfaces a
+total failure but tolerates a partial one.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import platform
+import stat
+import sys
+import tarfile
+import urllib.request
+import zipfile
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parents[2]
+CONST = json.loads((REPO / "src" / "envdrift" / "constants.json").read_text(encoding="utf-8"))
+IS_WIN = sys.platform == "win32"
+
+# (binary name, URL-map key in constants.json, version key in constants.json)
+TOOLS: list[tuple[str, str, str]] = [
+    ("dotenvx", "download_urls", "dotenvx_version"),
+    ("sops", "sops_download_urls", "sops_version"),
+    ("gitleaks", "gitleaks_download_urls", "gitleaks_version"),
+    ("trufflehog", "trufflehog_download_urls", "trufflehog_version"),
+    ("talisman", "talisman_download_urls", "talisman_version"),
+    ("trivy", "trivy_download_urls", "trivy_version"),
+    ("infisical", "infisical_download_urls", "infisical_version"),
+]
+
+
+def _platform_key(urls: dict[str, str]) -> str:
+    osname = {"linux": "linux", "darwin": "darwin", "win32": "windows"}[sys.platform]
+    machine = platform.machine().lower()
+    arch = "arm64" if machine in ("arm64", "aarch64") else "amd64"
+    key = f"{osname}_{arch}"
+    # Fall back to amd64 when a tool publishes no arm64 build for this OS.
+    if key not in urls and arch == "arm64":
+        key = f"{osname}_amd64"
+    return key
+
+
+def _download(url: str, dest: Path) -> None:
+    print(f"    GET {url}")
+    urllib.request.urlretrieve(url, dest)
+
+
+def _make_executable(path: Path) -> None:
+    path.chmod(path.stat().st_mode | stat.S_IEXEC | stat.S_IXGRP | stat.S_IXOTH)
+
+
+def _find_binary(root: Path, binary: str) -> Path:
+    """Locate the extracted binary by exact name, then by prefix."""
+    target = f"{binary}.exe" if IS_WIN else binary
+    for candidate in (target, binary):
+        matches = [p for p in root.rglob(candidate) if p.is_file()]
+        if matches:
+            return matches[0]
+    matches = [p for p in root.rglob(f"{binary}*") if p.is_file()]
+    if matches:
+        return matches[0]
+    raise FileNotFoundError(f"{binary!r} not found in extracted archive under {root}")
+
+
+def _install_tool(bindir: Path, binary: str, urls_key: str, version_key: str) -> None:
+    urls = CONST[urls_key]
+    key = _platform_key(urls)
+    if key not in urls:
+        raise RuntimeError(f"no download URL for platform {key}")
+    url = urls[key].format(version=CONST[version_key])
+    final_name = f"{binary}.exe" if IS_WIN else binary
+    dest = bindir / final_name
+    archive = bindir / Path(url).name
+
+    if url.endswith((".tar.gz", ".tgz")):
+        _download(url, archive)
+        extract_dir = bindir / f"_{binary}_x"
+        with tarfile.open(archive) as tf:
+            tf.extractall(extract_dir)
+        dest.write_bytes(_find_binary(extract_dir, binary).read_bytes())
+    elif url.endswith(".zip"):
+        _download(url, archive)
+        extract_dir = bindir / f"_{binary}_x"
+        with zipfile.ZipFile(archive) as zf:
+            zf.extractall(extract_dir)
+        dest.write_bytes(_find_binary(extract_dir, binary).read_bytes())
+    else:
+        # Raw, single-file binary (e.g. sops, talisman).
+        _download(url, dest)
+
+    if not IS_WIN:
+        _make_executable(dest)
+    print(f"    -> {dest}")
+
+
+def main() -> int:
+    bindir = Path.home() / ".envdrift-test-bin"
+    bindir.mkdir(parents=True, exist_ok=True)
+    print(f"platform={sys.platform}/{platform.machine()}  bindir={bindir}\n")
+
+    installed: list[str] = []
+    failed: list[str] = []
+    for binary, urls_key, version_key in TOOLS:
+        print(f"installing {binary} {CONST[version_key]}")
+        try:
+            _install_tool(bindir, binary, urls_key, version_key)
+            installed.append(binary)
+        except Exception as exc:
+            print(f"    !! FAILED to install {binary}: {exc}")
+            failed.append(binary)
+
+    print(f"\ninstalled: {', '.join(installed) or '(none)'}")
+    if failed:
+        print(f"failed:    {', '.join(failed)}")
+
+    gh_path = os.environ.get("GITHUB_PATH")
+    if gh_path and installed:
+        with open(gh_path, "a", encoding="utf-8") as fh:
+            fh.write(f"{bindir}\n")
+
+    return 0 if installed else 1
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/.github/workflows/cross-platform-integration.yml
+++ b/.github/workflows/cross-platform-integration.yml
@@ -55,6 +55,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Set up Python
         uses: actions/setup-python@v6
@@ -81,4 +83,4 @@ jobs:
         run: |
           uv run pytest tests/integration/ \
             -m "integration and not aws and not vault and not azure and not gcp" \
-            -p no:cacheprovider -v
+            -p no:cacheprovider --timeout=300 -v
--- a/.github/workflows/cross-platform-integration.yml
+++ b/.github/workflows/cross-platform-integration.yml
@@ -1,0 +1,84 @@
+name: Cross-Platform Integration
+
+# Runs the BINARY-BACKED integration tests (real dotenvx / sops / scanners over
+# real .env files and the real CLI) on macOS and Windows, so the encrypt,
+# decrypt, diff, partial-encryption, sops and scanner features get real
+# cross-platform coverage — not just Linux (integration-tests.yml).
+#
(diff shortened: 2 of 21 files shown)


.github/workflows/cross-platform-integration.yml
```diff
@@ -0,0 +1,84 @@
...
+      # add them to PATH. Resilient: a tool that fails to install is logged and
+      # its tests then skip for an honest "binary absent" reason.
+      - name: Install external CLI binaries
+        run: uv run python .github/scripts/install_test_binaries.py
+
+      - name: Run binary-backed integration tests (Linux containers excluded)
+        shell: bash
+        run: |
+          uv run pytest tests/integration/ \
+            -m "integration and not aws and not vault and not azure and not gcp" \
```
jainal09, line 85: 🤖 **envdrift review-bot** · medium · test-gap


The workflow runs `-m "integration and not aws and not vault and not azure and not gcp"` and its header asserts "Container-backed integration (the aws / vault / azure markers ...) ... is excluded here." That exclusion is NOT achieved by markers.

`tests/integration/test_e2e_workflows.py` is marked only `pytestmark = [pytest.mark.integration]` at module level, yet `TestPullDecryptWorkflow.test_e2e_pull_decrypt_workflow`, `TestLockPushWorkflow.test_e2e_lock_push_workflow`, and `TestMonorepoMultiService.test_e2e_monorepo_multi_service` consume the `localstack_endpoint`/`aws_secrets_client` fixtures (Docker/LocalStack) and have NO `@pytest.mark.aws`. The same is true for vault-dependent tests in `test_error_handling.py`/`test_concurrency.py` that lack a per-test `@pytest.mark.vault`. Under the workflow's selector these tests are SELECTED, not excluded — they pass only because the `localstack_endpoint`/`vault_endpoint` fixtures call `pytest.skip()` when the port is closed (conftest.py). So the "excluded" guarantee is incidental, not marker-enforced: any container-dependent test added later without an aws/vault/azure marker will actually run on mac/Windows and either fail or stall (bounded by the fixture's 1s port check / 40-min job timeout). Direction: mark every container-dependent test with its `aws`/`vault`/`azure` marker so the `-m` selector genuinely excludes them (matches CLAUDE.md's skip-gate convention), making the selector—not fixture luck—the contract.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit ff11d8b


The new cross-platform integration workflow documents a marker-based exclusion of container-backed tests, but that exclusion is enforced by fixture port-checks rather than by the `-m` selector — a latent test-gap worth tightening.

**Findings (1):**
- [medium] Marker selector does not actually exclude container-backed e2e tests; they only skip via fixture port-check — `.github/workflows/cross-platform-integration.yml:83` (inline)

jainal09:
Addressed the review threads in 66241b5:
- **Critical — archive path traversal**: the installer now validates every tar/zip member stays inside the extract dir (rejects `..`/absolute/symlink members) and passes `filter="data"` to tarfile on 3.12+.
- **exec-env child timeout**: restored (default 300s, overridable) — the argv reimplement had dropped `_run`'s 120s bound.
- **pytest `--timeout=300`** added (matches integration-tests.yml); **`persist-credentials: false`** on checkout.
- **tarfile `filter` DeprecationWarning**: silenced by the same `filter="data"`.

**test-gap (marker exclusion)**: the aws/vault/azure container tests are excluded in practice by the conftest port-availability skip-gates (no Docker on the runners), not only by markers — that's why Windows shows **227 passed, 8 skipped**. The marker filter is belt, the skip-gate is suspenders.

**Checksum verification (CodeRabbit Major)**: left open for a decision — see my reply on that thread.

.github/scripts/install_test_binaries.py
```diff
@@ -0,0 +1,141 @@
...
+    key = f"{osname}_{arch}"
+    # Fall back to amd64 when a tool publishes no arm64 build for this OS.
+    if key not in urls and arch == "arm64":
+        key = f"{osname}_amd64"
+    return key
+
+
+def _download(url: str, dest: Path) -> None:
+    print(f"    GET {url}")
+    urllib.request.urlretrieve(url, dest)
```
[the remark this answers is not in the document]
jainal09: Leaving this for a maintainer decision rather than burying it. Risk analysis for this specific script:

- It is **CI-only** (installs test binaries for the cross-platform job) — not the production install path users run (that is the separate `DotenvxInstaller`/`SopsInstaller`, with its own code/tests).
- URLs are **version-pinned** and fetched over **HTTPS from the official GitHub release** of each tool. TLS already protects transport integrity.
- Checksums published by the **same** GitHub release share the trust root, so they add limited protection against the scenario that matters most (a compromised release/maintainer — the attacker updates the checksum too). Real defense there is signature verification (cosign/sigstore), a much heavier lift.
- The higher-value extraction hardening (path traversal / `filter="data"`) is now in place.

So I scoped this as accepted risk for a CI installer. If we want it anyway, the clean path is a Renovate-maintained checksum map in `constants.json` (the same place versions/URLs live) + a verify step — happy to do that as a focused follow-up. Flagging to @jainal09 to decide: accept (resolve) or implement.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 66241b5


The `exec_env` rewrite (decrypt-to-memory) changed the argv contract, but the existing unit test that pins the old `sops exec-env <file> <command>` shell-string contract was not updated in this PR, so the standard `-m "not integration"` lane (and the Publish full-suite run) will go red.

**Findings (1):**
- [blocker] exec_env rewrite breaks existing unit test test_exec_env_builds_command (not updated in this PR) — `tests/unit/test_encryption_backends.py` (file not in diff — see below)

### Not anchored inline

This finding targets `tests/unit/test_encryption_backends.py`, which is not part of this PR's diff (the rewrite landed only in `src/envdrift/encryption/sops.py`), so there is no RIGHT-side diff line to anchor an inline comment to. Intra-file reference: `def test_exec_env_builds_command` at line 725, failing assertion at line 745.



**[blocker · regression] exec_env rewrite breaks existing unit test test_exec_env_builds_command (not updated in this PR)**

Commit 928bc7f rewrote `SOPSEncryptionBackend.exec_env` to decrypt-to-memory: it now calls `self._run(["-d", "--input-type", "dotenv", "--output-type", "dotenv", str(env_file)])` and then runs the child via `subprocess.run`. But the existing unit test `TestSOPSEncryptionBackend.test_exec_env_builds_command` (tests/unit/test_encryption_backends.py:725) still asserts the OLD shell-string contract:

```python
result = backend.exec_env(env_file, ["echo", "ok"])
assert captured["args"] == ["exec-env", str(env_file), "echo ok"]
```

The entire PR (`git diff main...66241b5 --stat`) does not touch this test file. I extracted the HEAD `sops.py` into a scratch dir and ran the test's mock harness against it: `captured["args"]` is now `['-d', '--input-type', 'dotenv', '--output-type', 'dotenv', <file>]`, so the assertion is `False` and the test FAILS.

Impact: this is a plain unit test (no markers/skips, class `TestSOPSEncryptionBackend`) that runs in the standard `-m "not integration"` CI lane on every platform (Linux included), so PR `CI` and the `Publish` full-suite run will go red. It is also a behavior-asserting mock test (CLAUDE.md flags mocking the thing under test) that now encodes a stale contract.

Direction: update `test_exec_env_builds_command` to assert the new decrypt-argv invocation (e.g. that `_run` is called with the `-d --input-type dotenv --output-type dotenv <file>` decrypt args and that the child command is executed with the decrypted secrets in its env), so the unit test matches the new exec_env behavior.

_Re-verified at PR head 88a5386: exec_env still uses the new decrypt argv (sops.py:589) and the test assertion is still the old contract (test_encryption_backends.py:745); the only file changed since 66241b5 is src/envdrift/encryption/sops.py, and the test file remains absent from the PR diff._

tests/unit/test_encryption_backends.py
```diff
@@ -733,16 +735,23 @@ def test_exec_env_builds_command(self, tmp_path, monkeypatch):
         captured = {}
 
         def fake_run(args, env=None, cwd=None):
+            # sops is invoked to DECRYPT to stdout, not via exec-env.
             captured["args"] = args
-            return MagicMock(returncode=0, stderr="", stdout="")
+            return MagicMock(returncode=0, stderr="", stdout="KEY=secretval\n")
```
jainal09, line 790: 🤖 **envdrift review-bot** · high · test-gap


The commit is explicitly framed as a bug fix ("a non-ASCII byte would corrupt a finding or raise UnicodeDecodeError"), but it adds **no** test that fails without the `encoding="utf-8"` lines and passes with them — violating CLAUDE.md's rule that every bug fix ships a real regression test failing on pre-fix code.

The two test edits do not exercise UTF-8 decoding at all: (1) `test_exec_env_decrypts_to_memory_and_injects_into_argv` mocks `_run` to return ASCII `"KEY=secretval\n"` and asserts the `-d` argv contract + env injection (a test for the *separate* exec_env refactor, not for subprocess decoding); (2) the `test_hook_check` change only widens `fake_run` to `**kwargs` so it tolerates the new encoding kwargs and still asserts env-stripping. Neither feeds non-ASCII bytes, so both pass unchanged if the `encoding="utf-8"` lines were reverted.

Worse, the regression cannot be caught by the new cross-platform workflow either: the autouse `_utf8_subprocess_on_windows` fixture in tests/integration/conftest.py force-injects `encoding="utf-8"` into every text-mode subprocess on Windows, so the integration suite would stay green even if the production scanner code reverted to locale decoding — the exact masking the commit message warns about. Suggested direction: add a focused unit/integration test that runs a real subprocess emitting non-ASCII UTF-8 bytes (a secret value or path with é/emoji) through one scanner/`_run` and asserts the decoded value round-trips (and would raise/garble under the prior locale path), without relying on the conftest patch.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 5b01fb5


The UTF-8 subprocess-decoding hardening is solid across the 9 scanners + guard/install/dotenvx/hook_check, but the "every tool-internal subprocess" claim has two gaps: a scanner-subsystem module (`scanner/engine.py`) was missed entirely, and the fix ships no regression test that fails on pre-fix code — and the new Windows integration fixture actively masks that gap.

**Findings (2):**
- [medium] Two scanner git check-ignore subprocess calls were MISSED by the UTF-8 fix — `src/envdrift/scanner/engine.py` (see below — not anchored; file not in PR diff)
- [high] UTF-8 'fix' ships no regression test that fails on pre-fix code; integration harness masks it — `tests/unit/test_encryption_backends.py:740` (inline)

### Not anchored inline

**[medium · bug] Two scanner git check-ignore subprocess calls were MISSED by the UTF-8 fix** — `src/envdrift/scanner/engine.py`

The commit claims to add `encoding="utf-8"` to *every* tool-internal text-mode subprocess call across all scanners, but two `subprocess.run(["git", "check-ignore", ...], text=True, ...)` calls in `scanner/engine.py` were missed: `_filter_gitignored_files` (around line 935) and `check_combined_files_security` (around line 1006). Both still run with `text=True` and **no** `encoding="utf-8"`, so on Windows their stdout is decoded with the cp1252 locale — exactly the latent bug the commit message says it is eliminating tool-wide.

This is not cosmetic: both calls parse git's stdout as file **paths** and match them against config/finding paths to decide scan results. `_filter_gitignored_files` splits stdout on NUL and drops findings whose path is in the ignored set; `check_combined_files_security` splits stdout on `\n` and uses `combined_file not in gitignored` to decide whether to emit a SECURITY WARNING. A non-ASCII repo/file path mis-decoded under cp1252 breaks the path equality, so a finding from a gitignored file may be wrongly *kept* (or a real one wrongly dropped), and a combined-file security warning may be wrongly suppressed or wrongly emitted. These calls live in the scanner subsystem the cross-platform workflow is meant to exercise, yet they were left inconsistent with the 9 scanners the commit did fix. Suggested direction: add `encoding="utf-8", errors="replace"` to both engine.py git check-ignore calls so scan-result filtering decodes git output identically on Windows.

_Evidence: scanner/engine.py ~935: `subprocess.run(["git","check-ignore","--stdin","-z"], input=..., capture_output=True, text=True, timeout=30, cwd=str(root))` then `ignored = [p for p in result.stdout.split("\0") if p]`; and ~1006: `subprocess.run(["git","check-ignore","--stdin"], ..., text=True, ...)` then `gitignored = set(result.stdout.strip().split("\n"))` — neither has `encoding="utf-8"`, unlike the scanners the commit touched. `scanner/engine.py` is not part of this PR's diff, so there is no RIGHT-side line to anchor an inline comment to._

src/envdrift/encryption/sops.py
```diff
@@ -574,13 +579,43 @@ def exec_env(
...
+        try:
+            return subprocess.run(  # nosec B603
+                command,
+                env=child_env,
+                cwd=str(kwargs["cwd"]) if kwargs.get("cwd") else None,
+                capture_output=True,
+                text=True,
+                encoding="utf-8",
+                timeout=kwargs.get("timeout", 300),
+            )
```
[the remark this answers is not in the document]
jainal09: Fixed in acb45a5: added `errors="replace"` to the child `subprocess.run`, so a malformed UTF-8 byte from the child degrades to U+FFFD instead of raising `UnicodeDecodeError`. Regression covered by the new `test_exec_env_child_stdout_decoded_as_utf8`.

tests/unit/test_encryption_backends.py
jainal09: Addressed in acb45a5: added `test_exec_env_child_stdout_decoded_as_utf8`. The child writes raw UTF-8 *bytes* for a non-ASCII secret (`café-→-μ`) and the test asserts the parent decodes them losslessly — exercising the `encoding="utf-8"` decode path on every platform, not just where the locale already defaults to UTF-8.

jainal09:
Resolving the CodeScene Code Health advisory threads on this PR with rationale:

**9× "Complex Method — Getting worse" (scanner/guard subprocess methods):** these methods _already_ carry high cyclomatic complexity (CodeScene says so on each), and this PR adds **zero branches** to them — the delta is purely +2 lines of code per method (`encoding="utf-8", errors="replace"`) for the cross-platform UTF-8 decode fix. e.g. `GitleaksScanner.scan`: LOC 99→101, complexity unchanged. The added kwargs are a correctness requirement on Windows (cp1252 would otherwise mis-decode the CLI/tool UTF-8 output); refactoring these long-standing scanner methods is out of scope for a CI-enablement + decode-safety PR.

**2× test-file size deltas (`test_encryption_backends.py`):** the file grew because this PR _adds_ real regression tests (incl. the non-ASCII UTF-8 round-trip test). Growing test coverage is the intent, not a regression to fix.

These are advisory (the `CodeScene Code Health Review` status check is green); resolving so the conversation-resolution merge gate reflects that they are reviewed and accepted.

tests/unit/test_encryption_backends.py
jainal09: 🤖 **envdrift review-bot**


Verified resolved against `4190c22`. `test_exec_env_child_stdout_decoded_as_utf8` is a genuine cross-platform regression test: the child writes the raw UTF-8 bytes of `café-→-μ` and the test asserts the parent reads back `result.stdout == "café-→-μ"`. Under the pre-fix locale decode (cp1252 on Windows) `→` (`E2 86 92`) would surface as mojibake (`â\x86\x92`), so the assertion fails without `encoding="utf-8"` and passes with it — exactly the gap that was open. The `exec_env` child `subprocess.run` now carries `encoding="utf-8", errors="replace"` (sops.py). 👍

One low-severity residual (optional, not blocking): the same `encoding="utf-8"` change was applied to ~12 other modules (the scanners, dotenvx, guard, install), and those lines aren't individually regression-tested — but they reuse this exact, now-proven decode pattern, so a per-module test would be redundant. Closing this thread as resolved.

Issue #458, opened 2026-06-09
validate/parser: multiline quoted values are split per physical line → truncation + phantom vars + false PASS


jainal09:
From the #443 re-audit (finding #20). **Deferred from the #18 validator PR** (P3) because it changes the core `EnvParser`, which underpins encrypt/decrypt/diff/guard too — high blast radius, deserves its own PR + a full-suite run.

## Summary

`EnvParser.parse_string` (`src/envdrift/core/parser.py`, docstring L128–132 still says *"Multiline values are not currently supported. Each line is parsed independently"*) `splitlines()` the content and matches `KEY=value` per **physical** line. A multiline quoted value is therefore:
- **truncated** to its first line (keeping the stray opening quote — `_unquote` only strips a matching pair), and
- its interior lines are **re-parsed as new assignments** (phantom variables).

This diverges from `pydantic-settings` and can flip `validate` to a **false PASS**.

## Evidence (current main, pydantic-settings 2.14.1)

Schema: `CERT: str` and `DB_PASSWORD: str` (both required).

```
CERT="-----BEGIN CERT-----
DB_PASSWORD=oops_interpreted_as_assignment
-----END CERT-----"
```

- **pydantic-settings (ground truth):** REJECTS — `DB_PASSWORD Field required [type=missing]` (the interior line is part of `CERT`'s value, not a separate var).
- **envdrift:** fabricates a phantom `DB_PASSWORD='oops_interpreted_as_assignment'` and truncates `CERT` to `'"-----BEGIN CERT-----'`. `validate … --no-check-encryption` → **"Validation PASSED" exit 0**, and **even with `--ci`** → PASSED exit 0.

So a `.env` that pydantic-settings rejects (a required secret actually absent) sails through envdrift's CI gate, and a variable the user never wrote is invented. (`validate` is read-only, so the fabrication is in the parse/validation logic, not on disk.)

Also: a `PRIVATE_KEY="-----BEGIN…"` PEM value is type-checked as `'"-----BEGIN PRIVATE KEY-----'` (stray leading quote, END marker lost), and an interior `PORT=` line yields a mangled `TYPE ERRORS: PORT: Expected integer, got '"8080'` plus a spurious `Extra variable 'WORKERS' not in schema`.

## Fix direction

[...]

Issue #18, opened 2025-12-31
Add vault push command to upload keys to vault


jainal09:
## Summary

Add a new `envdrift vault push` command to upload private keys to cloud vaults (Azure Key Vault, AWS Secrets Manager, HashiCorp Vault).

This is the reverse of `envdrift sync` - instead of pulling keys from vault to local, it pushes keys from local to vault.

### Option 1: From .env.keys file

```bash
envdrift vault push <folder> <secret-name> --env <environment>
```

**Example:**
```bash
envdrift vault push ./services/soak soak-machine --env soak
```

- Reads `DOTENV_PRIVATE_KEY_SOAK` from `./services/soak/.env.keys`
- Pushes to vault secret `soak-machine` with value `DOTENV_PRIVATE_KEY_SOAK=...`
- Uses vault config (provider, url, region) from `envdrift.toml`

### Option 2: Direct value

```bash
envdrift vault push --direct <secret-name> <value>
```

**Example:**
```bash
envdrift vault push --direct soak-machine "DOTENV_PRIVATE_KEY_SOAK=abc123..."
```

## Implementation Notes

- Reuse existing vault clients (Azure, AWS, HashiCorp) from `src/envdrift/vault/`
- Read vault config from `envdrift.toml`
- Handle edge cases:
  - Permission errors
  - Network errors
  - Secret already exists (update vs create)
  - Invalid key format

## Related

- Mirrors existing `az keyvault secret set` workflow:
  ```bash
  az keyvault secret set --vault-name galileo-kv-dev-001 --name soak-machine --value "DOTENV_PRIVATE_KEY_SOAK=..."
  ```

PR #459, opened 2026-06-09
fix(validate): enforce Pydantic field constraints so a rejected config can't pass

jainal09:
## What

P3 of the #443 re-audit (#18): `validate` reports **"Validation PASSED" / exit 0** on a config the real Pydantic Settings class **rejects** — field constraints (`ge`/`le`, `Literal`, `min_length`, `pattern`, …) were never enforced. A false pass on the CI gate.

(The companion finding #20 — multiline quoted values misparsed — is split out to #458 because it changes the core `EnvParser`.)

## Why

The validator only did name-based base-type checks (`int`/`float`/`bool`/`list`) and **never instantiated the Settings class**, so constraints were invisible.

## Reproduction (real CLI)

```
# schema: PORT:int=Field(ge=1,le=65535); LOG_LEVEL:Literal[...]; APP_NAME:str=Field(min_length=3); VERSION:str=Field(pattern=...)
printf 'PORT=99999\nLOG_LEVEL=trace\nAPP_NAME=ab\nVERSION=not-a-version\n' > .env
envdrift validate .env --schema config:Settings --no-check-encryption --ci
# BEFORE: Validation PASSED, exit 0   (real pydantic raises 4 ValidationErrors)
# AFTER:  Validation FAILED, exit 1   (PORT le, LOG_LEVEL literal, APP_NAME min_length, VERSION pattern)
```

## Fix

`SchemaMetadata` now carries the live `model_class` (set by `extract_metadata`). The validator keeps the heuristic base-type pass (preserving its messages and the encrypted/empty skips), then validates the type-valid, non-encrypted, non-empty values against the real class via `model_validate` and surfaces the constraint errors the heuristic can't see. `missing`/`extra` pydantic errors are ignored (already reported as `missing_required`/`extra_vars`), and a base-type message is never overridden. `model_validate` applies real constraints, coerces string env values, and does not read `os.environ`.

## Tests (dogfood, test-first)

- **New** `test_validate_enforces_pydantic_field_constraints` — **red** before the fix (PASSED), green after (all four constraints flagged).
- The three edge-case tests for float/bool messages and encrypted/empty skipping still pass (caught + fixed during the change).
- Full non-integration suite (**2492**) green; verified end-to-end with the real CLI.

Part of the #443 backlog (P3 of 9). See the [re-audit](https://github.com/jainal09/envdrift/issues/443#issuecomment-4663943129).

<h3>Greptile Summary</h3>

This PR closes a false-pass gap in `envdrift validate`: field constraints (`ge`/`le`, `Literal`, `min_length`, `pattern`, custom validators) were invisible to the validator because it only ran name-based type checks and never instantiated the Settings class. A two-stage approach is now used — the existing heuristic runs first (preserving its messages and encrypted/empty skips), then the live class is validated with `model_validate` on a filtered subset of values.

- `SchemaMetadata` gains `model_class` (the live Settings class) and `has_constraints` (set when any field has constraint metadata, a non-plain-scalar type, or a custom validator); the expensive `model_validate` pass is skipped entirely for trivially-typed schemas.
- `_is_string_coercible` guards against false-positives by excluding complex types (list/dict/nested models) that pydantic-settings parses via its JSON env source rather than from a raw string.
- Five new unit tests cover the fixed path end-to-end: enforced constraints, custom field validators, skip-on-encrypted/empty/absent, base-type message priority, and no false-fail on list fields.

Safe to merge. The two-stage validation approach is well-guarded: complex types are explicitly skipped to avoid false-positives, `missing`/`extra_forbidden` pydantic errors are filtered to avoid double-reporting, the heuristic message takes priority over constraint messages, and a broad `Exception` guard prevents a user-defined after-validator from crashing the CLI.

The change is narrowly scoped: the existing heuristic path is untouched, and the new constraint pass only runs when `model_class` and `has_constraints` are both set. Every identified edge case (encrypted values, empty values, complex types, partial dicts raising missing errors, base-type message clobbering, non-ValidationError exceptions) has both handling code and a corresponding test. No new data paths are exposed externally.

No files require special attention. `validator.py` carries the most logic change but is thoroughly tested.

| Filename | Overview |
|----------|----------|
| src/envdrift/core/schema.py | Adds `model_class` and `has_constraints` to `SchemaMetadata`, and detects constraints/validators during `extract_metadata` to decide whether the deep-validate pass is needed. Logic is correct and the decorator sniffing via `__pydantic_decorators__` cleanly handles custom field/model validators. |
| src/envdrift/core/validator.py | Adds `_is_string_coercible` helper and the constraint-validation pass via `model_validate`. Correctly skips encrypted/empty values, complex types (list/dict/nested), and filters `missing`/`extra_forbidden` pydantic errors. Base-type message preservation and broad Exception guard for non-ValidationError model validators are both in place. |
| tests/unit/test_validator.py | Five new tests cover the fix comprehensively: enforcing constraints, custom field validators, skip-on-encrypted/empty/absent, base-type message preservation, and no false-fail on complex list fields. Edge cases are well-exercised. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant CLI
    participant Validator
    participant BaseTypeCheck as _check_type (heuristic)
    participant Pydantic as model_validate (real Pydantic)

    CLI->>Validator: validate(env_file, schema)
    Validator->>Validator: build env_by_lower index
    Validator->>Validator: check missing required / extra vars / encryption

    loop for each field
        Validator->>BaseTypeCheck: _check_type(value, field_type)
        BaseTypeCheck-->>Validator: type_error or None
        Note over BaseTypeCheck: skips encrypted / empty values
    end

    alt schema.model_class is not None AND schema.has_constraints
        Validator->>Validator: build values dict (coercible, non-encrypted, non-empty fields only)
        Validator->>Pydantic: model_class.model_validate(values)
        alt ValidationError raised
            Pydantic-->>Validator: list of field errors
            loop for each pydantic error
                Note over Validator: skip missing / extra_forbidden
                Note over Validator: skip if base-type error already set
                Validator->>Validator: map alias to field_name
                Validator->>Validator: "result.type_errors[field_name] = msg"
            end
        else non-ValidationError raised
            Pydantic-->>Validator: Exception (caught, suppressed)
        else no error
            Pydantic-->>Validator: validated model (discarded)
        end
    end

    Validator-->>CLI: ValidationResult (valid / type_errors / missing / extras)
```

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/validate-en..."](https://github.com/jainal09/envdrift/commit/793f658e06791ced135994636cd94b187609c2a1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36590216)</sub>

### DIFF
--- a/src/envdrift/core/validator.py
+++ b/src/envdrift/core/validator.py
@@ -4,9 +4,31 @@ from __future__ import annotations
 
 import re
 from dataclasses import dataclass, field
+from typing import Literal, Union, get_args, get_origin
 
 from envdrift.core.parser import EncryptionStatus, EnvFile
 from envdrift.core.schema import FieldMetadata, SchemaMetadata
+
+_COERCIBLE_SCALARS = (str, int, float, bool)
+
+
+def _is_string_coercible(tp: object) -> bool:
+    """True if model_validate can validate a *raw env string* against ``tp``.
+
+    Pydantic-settings parses complex types (``list``/``dict``/nested models) from
+    their env string via JSON in its env source; ``model_validate`` of the raw
+    string does not, so feeding it such a field would wrongly reject a config the
+    real schema accepts. Only scalars, ``Literal``, and Optionals of those accept
+    a string directly, so only those are checked via model_validate.
+    """
+    if tp in _COERCIBLE_SCALARS:
+        return True
+    origin = get_origin(tp)
+    if origin is Literal:
+        return True
+    if origin is Union:  # Optional[X] / X | None
+        return all(arg is type(None) or _is_string_coercible(arg) for arg in get_args(tp))
+    return False
 
 
 @dataclass
@@ -239,6 +261,12 @@ class Validator:
                 # Mirror _check_type's skips: ciphertext and empty (= unset).
                 if value == "" or value.startswith(("encrypted:", "ENC[")):
                     continue
+                # Only feed fields whose raw env string model_validate can check.
+                # Complex types (list/dict/nested) are JSON-parsed by the env source
+                # at runtime, which model_validate of the raw string does not do, so
+                # including them would wrongly reject a valid config (#443 review).
+                if not _is_string_coercible(field_meta.field_type):
+                    continue
                 values[field_meta.alias or field_name] = value
 
             try:
@@ -255,6 +283,11 @@ class Validator:
                     field_name = alias_to_field.get(key, key)
                     if field_name in schema.fields and field_name not in result.type_errors:
                         result.type_errors[field_name] = err.get("msg", "invalid value")
+            except Exception:
+                # A model-level @model_validator / model_post_init can raise a
+                # non-ValidationError; the base-type check already ran, so a
+                # constraint-pass failure must not crash validate (#443 review).
+                pass
 
         # Determine overall validity
         # Note: unencrypted_secrets are warnings, not errors
--- a/src/envdrift/core/validator.py
+++ b/src/envdrift/core/validator.py
@@ -225,8 +225,9 @@ class Validator:
         # The heuristic above only parses base types, so a config the real schema
         # rejects on a *constraint* used to pass as valid (#443). Instantiate the
         # live Settings class with the type-valid, non-encrypted, non-empty values
-        # and surface the constraint errors the heuristic cannot see.
-        if schema.model_class is not None:
+        # and surface the constraint errors the heuristic cannot see. Skipped for
+        # trivially-typed schemas (no constraints), where it would only add cost.
+        if schema.model_class is not None and schema.has_constraints:
             from pydantic import ValidationError
 
             values: dict[str, str] = {}
--- a/src/envdrift/core/validator.py
+++ b/src/envdrift/core/validator.py
@@ -210,7 +210,8 @@ class Validator:
                                 "but is not marked sensitive in schema"
                             )
 
-        # Basic type validation
+        # Basic type validation (name-based). Produces envdrift's own messages and
+        # skips encrypted/empty values; kept as the source of base-type errors.
         for field_name, field_meta in schema.fields.items():
             env_var = env_by_lower.get(_lookup_key(field_meta))
             if env_var is None:
@@ -219,6 +220,40 @@ class Validator:
             type_error = self._check_type(env_var.value, field_meta.field_type)
             if type_error:
                 result.type_errors[field_name] = type_error
+
+        # Field-constraint validation (ge/le, Literal, min_length, pattern, ...).
+        # The heuristic above only parses base types, so a config the real schema
+        # rejects on a *constraint* used to pass as valid (#443). Instantiate the
+        # live Settings class with the type-valid, non-encrypted, non-empty values
+        # and surface the constraint errors the heuristic cannot see.
+        if schema.model_class is not None:
+            from pydantic import ValidationError
+
+            values: dict[str, str] = {}
+            for field_name, field_meta in schema.fields.items():
+                env_var = env_by_lower.get(_lookup_key(field_meta))
+                if env_var is None:
+                    continue
+                value = env_var.value
+                # Mirror _check_type's skips: ciphertext and empty (= unset).
+                if value == "" or value.startswith(("encrypted:", "ENC[")):
+                    continue
+                values[field_meta.alias or field_name] = value
+
+            try:
+                schema.model_class.model_validate(values)
+            except ValidationError as exc:
+                alias_to_field = {(fm.alias or name): name for name, fm in schema.fields.items()}
+                for err in exc.errors():
+                    # missing/extra are reported via missing_required / extra_vars;
+                    # don't override a base-type message the heuristic already set.
+                    if err.get("type") in ("missing", "extra_forbidden"):
+                        continue
+                    loc = err.get("loc") or ()
+                    key = str(loc[0]) if loc else ""
+                    field_name = alias_to_field.get(key, key)
+                    if field_name in schema.fields and field_name not in result.type_errors:
+                        result.type_errors[field_name] = err.get("msg", "invalid value")
 
         # Determine overall validity
         # Note: unencrypted_secrets are warnings, not errors
--- a/tests/unit/test_validator.py
+++ b/tests/unit/test_validator.py
@@ -312,6 +312,35 @@ NEW_FEATURE_FLAG=enabled
         assert result.valid is False
         assert "PORT" in result.type_errors
 
+    def test_validate_enforces_pydantic_field_constraints(self, tmp_path):
+        """#18: field constraints (ge/le/Literal/min_length/pattern) must be enforced.
+
+        validate previously did only name-based type checks and never instantiated
+        the Settings class, so a config the real Pydantic schema REJECTS sailed
+        through as 'Validation PASSED' (exit 0) — a false pass on the CI gate.
+        """
+        from typing import Literal
+
+        from pydantic import Field
+        from pydantic_settings import BaseSettings
+
+        class ConstrainedSettings(BaseSettings):
+            PORT: int = Field(ge=1, le=65535)
+            LOG_LEVEL: Literal["debug", "info", "warning", "error"]
+            APP_NAME: str = Field(min_length=3)
+            VERSION: str = Field(pattern=r"^v\d+\.\d+\.\d+$")
+
+        # Type-correct values that every constraint nonetheless rejects.
+        env_file = tmp_path / ".env"
+        env_file.write_text("PORT=99999\nLOG_LEVEL=trace\nAPP_NAME=ab\nVERSION=not-a-version\n")
+
+        env = EnvParser().parse(env_file)
+        schema = SchemaLoader().extract_metadata(ConstrainedSettings)
+        result = Validator().validate(env, schema, check_encryption=False)
+
+        assert result.valid is False
+        assert set(result.type_errors) >= {"PORT", "LOG_LEVEL", "APP_NAME", "VERSION"}
+
     def test_validate_suspicious_plaintext(self, tmp_path, permissive_settings_class):
         """Warn about plaintext values matching secret patterns."""
         content = """
--- a/tests/unit/test_validator.py
+++ b/tests/unit/test_validator.py
@@ -341,6 +341,37 @@ NEW_FEATURE_FLAG=enabled
         assert result.valid is False
         assert set(result.type_errors) >= {"PORT", "LOG_LEVEL", "APP_NAME", "VERSION"}
 
+    def test_validate_runs_custom_validator_on_plain_field(self, tmp_path):
+        """A custom @field_validator on a plain-typed field is still enforced.
+
+        The field is a bare ``str`` (no constraint metadata), so the only
+        validation is the custom validator — has_constraints must flag it so the
+        real model_validate pass runs and surfaces the rejection.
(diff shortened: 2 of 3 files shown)


src/envdrift/core/validator.py
```diff
@@ -220,6 +221,40 @@ def _lookup_key(fm: FieldMetadata) -> str:
...
+                if env_var is None:
+                    continue
+                value = env_var.value
+                # Mirror _check_type's skips: ciphertext and empty (= unset).
+                if value == "" or value.startswith(("encrypted:", "ENC[")):
+                    continue
+                values[field_meta.alias or field_name] = value
+
+            try:
+                schema.model_class.model_validate(values)
```
jainal09, line 273: 🤖 **envdrift review-bot** · high · regression


The constraint block calls `schema.model_class.model_validate(values)` where `values` is a `dict[str, str]` of raw .env strings. But `model_validate()` is NOT the pydantic-settings *env source* — the env source JSON-decodes list/dict/nested-model string values BEFORE validation, while `model_validate` of a plain string does not. So for any `list[...]` / `dict[...]` field, a value the real Settings class loads fine now produces a spurious `list_type`/`dict_type` error and `result.valid` flips to False (exit 1). This is the over-strict CI-red regression CLAUDE.md's 'full suite on validation changes' note warns about — `validate` now REJECTS a config the real Pydantic schema ACCEPTS, the mirror-image of the very false-PASS this PR fixes.

Reproduced end-to-end with the head validator + real pydantic 2.12.5:
```
class Cfg(BaseSettings):
    ALLOWED_HOSTS: list[str] = ["localhost"]
# .env: ALLOWED_HOSTS=["localhost","example.com"]
validate(...).valid          -> False  type_errors={'ALLOWED_HOSTS': 'Input should be a valid list'}
Cfg().ALLOWED_HOSTS (real)    -> ['localhost', 'example.com']   # accepted
```
The heuristic `_check_type` list branch intentionally `pass`es (accepts anything), so the `field_name not in result.type_errors` guard does NOT protect list/dict fields — the constraint block sets the bogus error unguarded. This also contradicts documented behavior: docs/cli/diff.md:97 lists `list[str]` as a supported field type and docs/guides/migrating.md:165,192 show `ALLOWED_HOSTS: list[str]` as a first-class pattern.

Suggested direction: for fields whose annotation is a list/dict/JSON-complex type, either skip the constraint pass, or pre-decode JSON (and fall back gracefully) so model_validate sees the same value the env source would — i.e. validate through a path that mirrors `BaseSettings` env-source coercion rather than raw `model_validate` of strings.

tests/unit/test_validator.py
```diff
@@ -312,6 +312,35 @@ def test_validate_type_mismatch(self, tmp_path, test_settings_class):
         assert result.valid is False
         assert "PORT" in result.type_errors
 
+    def test_validate_enforces_pydantic_field_constraints(self, tmp_path):
```
jainal09, line 315: 🤖 **envdrift review-bot** · high · test-gap


The new test only asserts that a *constraint-violating* config now FAILS. There is no positive test that a *valid* config containing a `list[str]`/`dict[...]` field (the kind whose value the real env source JSON-decodes) still PASSES. That missing case is exactly why the over-strict false-FAIL above slipped through — a valid `ALLOWED_HOSTS=["a","b"]` config now incorrectly reports FAILED. Per CLAUDE.md, a stricter validator must be proven not to newly reject configs that are actually valid.

Suggested direction: add a regression test that builds a Settings class with a `list[str]` (and/or `dict[str,str]`) field, writes a JSON-encoded value the real class accepts, and asserts `result.valid is True` and the field is NOT in `type_errors`. It will fail on the current code, pinning the fix for the finding above.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 2726ae4


This PR fixes a real false-PASS in `validate`, but the new constraint pass calls `model_class.model_validate()` on raw .env strings instead of mirroring the pydantic-settings env source — introducing the mirror-image regression: valid `list[...]`/`dict[...]` (JSON-typed) configs that the real Settings class accepts are now wrongly REJECTED, and no positive test covers that case.

**Findings (2):**
- [high] validate now FAILS on valid list/dict (JSON-typed) configs the real Settings class accepts — `src/envdrift/core/validator.py:244` (inline)
- [high] No test asserts a valid complex-typed (list/dict) config still PASSES under the stricter validator — `tests/unit/test_validator.py:315` (inline)

src/envdrift/core/validator.py
```diff
@@ -210,7 +210,8 @@ def _lookup_key(fm: FieldMetadata) -> str:
                                 "but is not marked sensitive in schema"
                             )
 
-        # Basic type validation
+        # Basic type validation (name-based). Produces envdrift's own messages and
+        # skips encrypted/empty values; kept as the source of base-type errors.
```
jainal09, line 236: Advisory (non-required). The added constraint pass is a flat, well-scoped block (build values → one model_validate → map errors); it is gated behind has_constraints so trivially-typed schemas skip it entirely. The complexity delta is the cost of enforcing real Pydantic constraints, which closes a false-PASS on the validate CI gate. Refactoring the long-standing validate method is out of scope.

src/envdrift/core/validator.py
jainal09: 🤖 **envdrift review-bot**


Re-checked at `f892c9c` — this finding is **not resolved** by the `perf(validate): skip model_validate for trivially-typed schemas (no constraints)` change. The skip is gated on `schema.has_constraints`, but that flag is set whenever a field's `annotation not in _PLAIN_SCALARS` (= `(str, int, float, bool)`, schema.py:275). A `list[str]`/`dict`/`Literal`/nested field is **not** a plain scalar, so it always forces `has_constraints=True` — which means the constraint block still runs and still calls `model_validate()` on the raw `dict[str, str]` for exactly the list/dict fields that can't survive it. The optimization only short-circuits pure str/int/float/bool schemas, which never had a list/dict field to false-fail in the first place.

Reproduced against the head logic with pydantic-settings (a schema with `TAGS: list[str]` + a constrained `PORT`, `.env: TAGS=["a","b"]`):

```
model_validate(raw strings): FAIL -> [(('TAGS',), 'list_type')]   # validator.py sets type_errors['TAGS'] -> valid=False
real Cfg() from env:         PASSED, TAGS = ['a', 'b']            # the env source JSON-decodes it
```

So `validate` still reports a config the real Settings class **accepts** as invalid (exit 1). Direction is unchanged: for list/dict/JSON-complex annotations, either exclude them from the constraint pass or JSON-pre-decode their value so `model_validate` sees what the `BaseSettings` env source would. (The companion test-gap also stands — no test asserts a list/dict config *passes*.)

jainal09: Fixed in 9ac6778: complex-typed fields (list/dict/nested) are skipped from the model_validate pass via `_is_string_coercible`, since pydantic-settings JSON-parses them at runtime from the env string and model_validate of the raw string does not. `TAGS=["a","b","c"]` now validates correctly. Positive regression test added.

tests/unit/test_validator.py
jainal09: Added in 9ac6778: `test_validate_does_not_false_fail_on_complex_typed_field` asserts a valid config with a JSON list field PASSES (red before this commit).

src/envdrift/core/validator.py
```diff
@@ -220,6 +221,40 @@ def _lookup_key(fm: FieldMetadata) -> str:
...
+                for err in exc.errors():
+                    # missing/extra are reported via missing_required / extra_vars;
+                    # don't override a base-type message the heuristic already set.
+                    if err.get("type") in ("missing", "extra_forbidden"):
+                        continue
+                    loc = err.get("loc") or ()
+                    key = str(loc[0]) if loc else ""
+                    field_name = alias_to_field.get(key, key)
+                    if field_name in schema.fields and field_name not in result.type_errors:
+                        result.type_errors[field_name] = err.get("msg", "invalid value")
```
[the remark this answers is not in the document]
jainal09: Fixed in 9ac6778: a broad `except Exception` now wraps model_validate, so a non-ValidationError from a model_post_init / @model_validator no longer crashes validate (the base-type check already ran).

src/envdrift/core/validator.py
```diff
@@ -220,6 +221,41 @@ def _lookup_key(fm: FieldMetadata) -> str:
...
```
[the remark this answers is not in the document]
jainal09: Addressed in 9ac6778: model-validator errors raised as ValidationError are still mapped to type_errors; a non-ValidationError is now caught by the broad except rather than propagating.

src/envdrift/core/validator.py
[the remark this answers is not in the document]
jainal09: Acknowledged: model_validate constructs the model, so after-validators/model_post_init run (true of the prior code too). For `validate`, running the schema's own validators is the intended strongest check; a side-effecting post_init is rare and now its errors are caught rather than crashing.

src/envdrift/core/validator.py
jainal09, line 236: Advisory (non-required). The constraint pass is a scoped block gated behind has_constraints; the added complexity is the cost of enforcing real Pydantic constraints + skipping complex types. Refactoring the long-standing validate method is out of scope.

src/envdrift/core/validator.py
jainal09: 🤖 **envdrift review-bot**


Resolved at `9ac6778` — thanks for the proper fix. The constraint pass now skips any field where `not _is_string_coercible(field_meta.field_type)`, so `list`/`dict`/nested fields are excluded from `model_validate` (which can't JSON-decode a raw env string the way pydantic-settings' env source does), while scalars / `Literal` / `Optional`-of-those are still fed in — so the #443 constraint enforcement is preserved (a constrained scalar like `PORT: int = Field(ge=1)` keeps its `int` annotation and is still checked). I confirmed `TAGS: list[str]` with `TAGS=["a","b","c"]` now reports valid, and the companion test-gap is closed by `test_validate_does_not_false_fail_on_complex_typed_field`. Both review-bot findings here are addressed.

(Minor, non-blocking, your call: a *container-level* constraint on a complex field — e.g. `Field(min_length=1)` on a `list` — is now also skipped, so that one narrow class of constraint goes unchecked again. Almost certainly the right trade-off vs. false-failing every valid list config; flagging only for completeness.)

tests/unit/test_validator.py
jainal09: 🤖 **envdrift review-bot**


Resolved at `9ac6778` — `test_validate_does_not_false_fail_on_complex_typed_field` (a `list[str]` config that must report valid) covers exactly this gap. Thanks.

src/envdrift/core/validator.py
```diff
@@ -210,7 +232,8 @@ def _lookup_key(fm: FieldMetadata) -> str:
                                 "but is not marked sensitive in schema"
                             )
 
-        # Basic type validation
+        # Basic type validation (name-based). Produces envdrift's own messages and
+        # skips encrypted/empty values; kept as the source of base-type errors.
```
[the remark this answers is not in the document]
jainal09: Advisory (non-required). The added branches are the cost of replacing uncaught-traceback / false-PASS paths with clean, typed handling; each is small and scoped. Refactoring the long-standing method is out of scope for this fix.

PR #464, opened 2026-06-10
test(decrypt): cover the sops binary-blob stderr-decode path (#443 #14)

jainal09:
P8 of the #443 re-audit. #14 was classified **fixed-but-untested**: the sops backend's binary-blob decrypt used to raise an uncaught `UnicodeDecodeError` decoding sops' non-UTF-8 stderr; #451 fixed it (`sops._run` reads with `encoding="utf-8", errors="replace"`) but left no regression test.

Adds a real-`sops` integration test (skip-gated on the binary): decrypting a 1KB binary blob must fail **cleanly** (a failure result or a typed `EncryptionBackendError`), never an uncaught `UnicodeDecodeError`, and leave the file byte-for-byte intact.

Part of the #443 backlog (P8 of 9). See the [re-audit](https://github.com/jainal09/envdrift/issues/443#issuecomment-4663943129).

<h3>Greptile Summary</h3>

Adds a skip-gated integration test (`test_decrypt_sops_binary_blob_does_not_crash`) that exercises the SOPS backend's non-UTF-8 stderr-decode path (issue #14), previously fixed in `sops._run` via `encoding="utf-8", errors="replace"` but left untested.

- Writes a binary blob carrying non-UTF-8 bytes (`\xff\xfe\x80\x81…`), calls `backend.decrypt()`, and asserts it raises `EncryptionBackendError` (not an uncaught `UnicodeDecodeError`); a `UnicodeDecodeError` propagating out of `_run` would miss `pytest.raises` and fail the test.
- Follows the `pytest.raises` context with a byte-level file-integrity assertion to confirm sops did not silently corrupt the blob on failure.

Test-only change that adds a regression guard for a previously fixed but untested bug; no production code is modified.

The added test is correctly structured: pytest.raises(EncryptionBackendError) tightly covers the failure path so a reintroduced UnicodeDecodeError (which _run does not catch) would propagate past the context manager and fail the test. The sops_on_path skip gate is consistent with the existing dotenvx_on_path fixture. The post-raise file-integrity assertion runs in the right place. No logic touches production code.

| Filename | Overview |
|----------|----------|
| tests/integration/test_decrypt_outcome_bench.py | Adds sops_on_path fixture and test_decrypt_sops_binary_blob_does_not_crash; regression guard for issue #14, using pytest.raises(EncryptionBackendError) correctly after the previous dead-code assertion was removed. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant T as Test
    participant B as SOPSEncryptionBackend
    participant R as _run()
    participant S as sops binary

    T->>B: "decrypt(env_file, cwd=tmp_path)"
    B->>R: _run(["--decrypt", "--in-place", "--input-type", "dotenv", ...])
    R->>S: "subprocess.run(..., encoding="utf-8", errors="replace")"
    S-->>R: "returncode=1, stderr=<non-UTF-8 bytes replaced with U+FFFD>"
    R-->>B: "CompletedProcess(returncode=1)"
    B-->>T: raises EncryptionBackendError
    Note over T: pytest.raises catches EncryptionBackendError ✓
    T->>T: "assert env_file.read_bytes() == blob ✓"
```

<sub>Reviews (6): Last reviewed commit: ["test(decrypt): drive sops to emit non-UT..."](https://github.com/jainal09/envdrift/commit/1d9a684d20e0e9cf3b0f22eaa8bba34fd37ab498) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36601212)</sub>

### DIFF
--- a/tests/integration/test_decrypt_outcome_bench.py
+++ b/tests/integration/test_decrypt_outcome_bench.py
@@ -51,7 +51,11 @@ def test_decrypt_sops_binary_blob_does_not_crash(tmp_path: Path, sops_on_path: N
     the blob; the point is it fails *cleanly* and leaves the file intact.
     """
     env_file = tmp_path / ".env"
-    blob = bytes(range(256)) * 4  # NUL + non-UTF-8 bytes
+    # The *invalid* first line (no '=') must carry non-UTF-8 bytes so sops echoes
+    # them back on stderr ("invalid dotenv input line: \xff\xfe..."). That is the
+    # exact path the #14 fix guards: sops._run decoding that non-UTF-8 stderr. An
+    # all-valid-UTF-8 blob (e.g. bytes(range(256))) would NOT exercise it.
+    blob = b"\xff\xfe\x80\x81\xc0\xc1notavalidline\n"
     env_file.write_bytes(blob)
 
     backend = SOPSEncryptionBackend()
--- a/tests/integration/test_decrypt_outcome_bench.py
+++ b/tests/integration/test_decrypt_outcome_bench.py
@@ -22,6 +22,7 @@ import pytest
 
 from envdrift.encryption.base import EncryptionBackendError
 from envdrift.encryption.dotenvx import DotenvxEncryptionBackend
+from envdrift.encryption.sops import SOPSEncryptionBackend
 
 # Mark all tests in this module
 pytestmark = [pytest.mark.integration]
@@ -32,6 +33,35 @@ def dotenvx_on_path() -> None:
     """Skip the test if the real ``dotenvx`` binary is not installed."""
     if shutil.which("dotenvx") is None:
         pytest.skip("dotenvx binary not found on PATH")
+
+
+@pytest.fixture
+def sops_on_path() -> None:
+    """Skip the test if the real ``sops`` binary is not installed."""
+    if shutil.which("sops") is None:
+        pytest.skip("sops binary not found on PATH")
+
+
+def test_decrypt_sops_binary_blob_does_not_crash(tmp_path: Path, sops_on_path: None) -> None:
+    """#14: decrypting a binary blob with the SOPS backend must not raise an
+    uncaught UnicodeDecodeError while decoding sops' (non-UTF-8) stderr.
+
+    sops._run reads the subprocess with encoding='utf-8', errors='replace', so a
+    non-UTF-8 stderr byte degrades to U+FFFD instead of crashing. sops rejects
+    the blob; the point is it fails *cleanly* and leaves the file intact.
+    """
+    env_file = tmp_path / ".env"
+    blob = bytes(range(256)) * 4  # NUL + non-UTF-8 bytes
+    env_file.write_bytes(blob)
+
+    backend = SOPSEncryptionBackend()
+    try:
+        result = backend.decrypt(env_file, cwd=tmp_path)
+        assert result.success is False
+    except EncryptionBackendError:
+        pass  # a clean, typed failure is the acceptable outcome
+    # No silent corruption: the blob is byte-for-byte intact.
+    assert env_file.read_bytes() == blob
 
 
 # --- A non-encrypted file is an honest no-op, never mutated --------------------
--- a/tests/integration/test_decrypt_outcome_bench.py
+++ b/tests/integration/test_decrypt_outcome_bench.py
@@ -55,11 +55,11 @@ def test_decrypt_sops_binary_blob_does_not_crash(tmp_path: Path, sops_on_path: N
     env_file.write_bytes(blob)
 
     backend = SOPSEncryptionBackend()
-    try:
-        result = backend.decrypt(env_file, cwd=tmp_path)
-        assert result.success is False
-    except EncryptionBackendError:
-        pass  # a clean, typed failure is the acceptable outcome
+    # sops rejects the blob; the point is it surfaces a clean, typed
+    # EncryptionBackendError — not a UnicodeDecodeError from decoding sops'
+    # non-UTF-8 stderr (which would fail this pytest.raises and flag the bug).
+    with pytest.raises(EncryptionBackendError):
+        backend.decrypt(env_file, cwd=tmp_path)
     # No silent corruption: the blob is byte-for-byte intact.
     assert env_file.read_bytes() == blob
 


tests/integration/test_decrypt_outcome_bench.py
```diff
@@ -34,6 +35,35 @@ def dotenvx_on_path() -> None:
...
+    env_file = tmp_path / ".env"
+    blob = bytes(range(256)) * 4  # NUL + non-UTF-8 bytes
+    env_file.write_bytes(blob)
+
+    backend = SOPSEncryptionBackend()
+    try:
+        result = backend.decrypt(env_file, cwd=tmp_path)
+        assert result.success is False
+    except EncryptionBackendError:
+        pass  # a clean, typed failure is the acceptable outcome
```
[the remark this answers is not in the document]
jainal09: Fixed in 30f2d7b: sops decrypt of a blob always raises, so the success=False assertion was dead. Switched to `pytest.raises(EncryptionBackendError)` — tighter, and a UnicodeDecodeError would fail it (the regression being guarded).

tests/integration/test_decrypt_outcome_bench.py
```diff
@@ -34,6 +35,35 @@ def dotenvx_on_path() -> None:
...
+def test_decrypt_sops_binary_blob_does_not_crash(tmp_path: Path, sops_on_path: None) -> None:
+    """#14: decrypting a binary blob with the SOPS backend must not raise an
+    uncaught UnicodeDecodeError while decoding sops' (non-UTF-8) stderr.
+
+    sops._run reads the subprocess with encoding='utf-8', errors='replace', so a
+    non-UTF-8 stderr byte degrades to U+FFFD instead of crashing. sops rejects
+    the blob; the point is it fails *cleanly* and leaves the file intact.
+    """
+    env_file = tmp_path / ".env"
+    blob = bytes(range(256)) * 4  # NUL + non-UTF-8 bytes
```
jainal09: 🤖 **envdrift review-bot** · high · test-gap


This test is the regression guard for #14 (the pre-#451 `sops._run` raised an uncaught `UnicodeDecodeError` decoding sops' **non-UTF-8 stderr**). But the chosen input does not drive sops to emit any non-UTF-8 byte on stderr, so it does not exercise the bug.

I drove the real `sops` 3.13.1 binary in an isolated scratch dir (not the test suite / no shared containers) with the exact args envdrift's `decrypt` uses (`--decrypt --in-place --input-type dotenv --output-type dotenv`). For `blob = bytes(range(256)) * 4` sops writes to stderr:

```
b'invalid dotenv input line: \x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n'
```

That is **valid UTF-8**: the blob begins `\x00..\x09\x0a`, and the `\x0a` (newline at byte index 10) truncates sops' echoed "invalid dotenv input line" message at bytes 0x00–0x09 — none of which are non-UTF-8. Consequences I verified empirically:
- Decoding this stderr under the **pre-#451** path (`subprocess.run(..., text=True)` on a UTF-8 locale = strict UTF-8) **succeeds with no `UnicodeDecodeError`**. So pre-fix `_run` returns normally, `decrypt` sees rc!=0 and raises `EncryptionBackendError`, and `pytest.raises(EncryptionBackendError)` **passes — even without the fix**.
- `errors='replace'` (the #451 fix) yields **zero** U+FFFD replacement chars for this stderr, i.e. the fixed code path produces a byte-identical result. The fix is never exercised.

Impact: per CLAUDE.md ("Every bug fix ships a real regression test [that] FAILS without the fix"), this test is green with the bug present — it guards nothing. It also misleads future maintainers (the docstring and the `# NUL + non-UTF-8 bytes` comment claim a path that isn't reached).

Suggested direction (do not implement here): make the malformed line contain a non-UTF-8 byte **before** the first newline so sops echoes it. I confirmed `bytes([0xff, 0x80, 0xfe]) + b'\n'` produces stderr `b'invalid dotenv input line: \xff\x80\xfe\n'` (genuinely non-UTF-8) — that input crashes the pre-#451 strict decode with `UnicodeDecodeError` and only passes once `errors='replace'` is in place. Optionally assert the surfaced `EncryptionBackendError` message contains U+FFFD to prove the replace path ran. Note sops' message is truncated at the first newline, so the high bytes must precede it (a blob like `bytes(b for b in range(256) if b != 0x0a)` instead fails later with the ASCII-only `sops metadata not found`).

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit a6906cb


This PR's regression test for #14 does not actually reproduce the bug it guards: the chosen blob drives sops to emit *valid-UTF-8* stderr, so the test passes even with the pre-#451 fix reverted. The production fix itself is correct and in place — this is a test-only gap, but it defeats the stated purpose of the test.

**Findings (1):**
- [high] Chosen blob does not reproduce #14: sops' stderr for `bytes(range(256))*4` is valid UTF-8, so the test passes WITHOUT #451's fix — `tests/integration/test_decrypt_outcome_bench.py:54` (inline)

tests/integration/test_decrypt_outcome_bench.py
jainal09: Fixed in 1d9a684 — confirmed your point: `bytes(range(256))` is all valid UTF-8, so sops emitted valid-UTF-8 stderr and the test never exercised the decode. Now uses `b"\xff\xfe\x80\x81\xc0\xc1notavalidline\n"` — an invalid (no `=`) line of non-UTF-8 bytes that sops echoes verbatim on stderr (`invalid dotenv input line: \xff\xfe...`), forcing `sops._run`'s errors="replace" decode. Verified the raw stderr is non-UTF-8.

jainal09: 🤖 **envdrift review-bot**


Verified resolved at `1d9a684`. The new blob `b"\xff\xfe\x80\x81\xc0\xc1notavalidline\n"` puts genuinely non-UTF-8 bytes in the *first* (invalid) line, so sops echoes them on stderr before any newline — the pre-#451 strict `text=True` decode raises `UnicodeDecodeError` there and the old code never reaches `EncryptionBackendError`, so `pytest.raises` now fails without the fix and passes once `errors="replace"` degrades them to U+FFFD. The test now genuinely guards #14. Thanks for the quick turnaround.

Issue #467, opened 2026-06-10
Merged #443 follow-ups: partial-encryption still bypasses secret-lockout guard (#457) + init.md isidentifier docs stale (#460)


jainal09:
_🤖 Filed by the envdrift review-bot (via Jainal's account)._

Two review-bot findings from the #443 wave merged without being addressed. Both still reproduce on `main`.

---

## 1. Secret-lockout guard is backend-only — partial-encryption (`push`/`lock`) still bypasses it  ·  _medium · data-loss_

Originally raised on #457 (merged `6a60a63`). #457 added a filename-slug guard to **`DotenvxEncryptionBackend.encrypt`** (`src/envdrift/encryption/dotenvx.py`, the `[A-Za-z0-9._-]` check), and the merge also routed `DotenvxWrapper.encrypt` through `self._validate_filename`. **But `_validate_filename` only rejects the Windows `.env.local` pattern**, not space/non-ASCII slugs:

```python
# src/envdrift/integrations/dotenvx.py (main)
PROBLEMATIC_FILENAME_PATTERNS = [
    r"^\.env\.local$",          # <- the ONLY pattern
]
def _validate_filename(file_path):
    for pattern in PROBLEMATIC_FILENAME_PATTERNS:
        if re.match(pattern, file_path.name.lower(), re.IGNORECASE):
            raise DotenvxFilenameError(...)
```

`DotenvxWrapper.encrypt` has no `[A-Za-z0-9._-]` slug check of its own. The partial-encryption path uses the wrapper, not the guarded backend:

- `encrypt_secret_file` → `DotenvxWrapper().encrypt(secret_file)` (`src/envdrift/core/partial_encryption.py`)
- `_encrypt_secrets_only_file` (from `push_secrets_only`) → `dotenvx.encrypt(file)`, where `file` comes from `secrets_dir.glob(pattern)` — **arbitrary on-disk filenames**.

[...]

PR #468, opened 2026-06-10
fix(partial-encryption): refuse filenames dotenvx can't turn into a valid key (#467)

jainal09:
### The bug (medium · **data-loss**)
#457 added the `[A-Za-z0-9._-]` filename guard to **`DotenvxEncryptionBackend.encrypt`**, but the
partial-encryption push/lock paths reach dotenvx through **`DotenvxWrapper.encrypt`** directly, *not*
the guarded backend:

- `encrypt_secret_file` → `DotenvxWrapper().encrypt(secret_file)` (combine mode)
- `push_secrets_only` → `_encrypt_secrets_only_file` → `dotenvx.encrypt(file)`, where `file` comes
  from `secrets_dir.glob(pattern)` — arbitrary on-disk filenames.

`DotenvxWrapper._validate_filename` only rejects the Windows `.env.local` pattern — it has no slug
check. So a `my secret.env` / `café.env.secret` file pushed via `envdrift push`/`lock` still hit the
exact permanent lockout #457 set out to fix.

**Confirmed live** through the real `push_secrets_only`:
```
push_secrets_only → {'encrypted': 1}        # claimed success
on disk           → DOTENV_PRIVATE_KEY_MY SECRET... (invalid: space in key name)
plaintext         → DESTROYED
pull_secrets_only → MISSING_PRIVATE_KEY      # permanent lockout
```

### The fix
- Extract a single shared predicate **`is_dotenvx_safe_filename(name)`** in `integrations/dotenvx.py`.
- `DotenvxWrapper.encrypt` now calls `_validate_encryptable_filename` **on every platform** (the
  lockout is not Windows-specific) and **before invoking dotenvx**, so the plaintext is preserved.
- `DotenvxEncryptionBackend.encrypt` reuses the same predicate (no more inline regex) so the two
  guards cannot drift.
- The partial-encryption callers wrap the resulting `DotenvxFilenameError` into a clean
  `PartialEncryptionError` (no raw traceback).

### Tests (real, no mock of the seam under test)
Two regression tests drive the **real** wrapper guard through `push_secrets_only` (space name) and
`encrypt_secret_file` (non-ASCII name) and assert refusal with the plaintext **byte-for-byte intact**.
Both fail RED on `main` (`DID NOT RAISE`; the file is silently encrypted and destroyed). The guard
fires before dotenvx is invoked, so the tests need no dotenvx binary.

Full non-integration suite green (2521 passed); ruff / pyrefly / bandit clean.

> Finding #2 of #467 (the `docs/cli/init.md` `isidentifier()` claim) is docs-only and a different
> subsystem — it ships as a separate small docs PR so this stays themed.

## Release Notes

* **New Features**
  * Added filename validation system to prevent encryption of files with incompatible character sets for key derivation.

* **Bug Fixes**
  * Improved error handling for encryption failures related to invalid filenames while preserving original data.
  * Enhanced cross-platform filename safety checks to prevent undecryptable encryption results.

* **Tests**
  * Added test coverage for filename validation during encryption operations.

<h3>Greptile Summary</h3>

This PR closes a data-loss gap left by #457: the `[A-Za-z0-9._-]` filename guard existed only in `DotenvxEncryptionBackend`, but the partial-encryption push/lock paths (`push_secrets_only` → `_encrypt_secrets_only_file`, `encrypt_secret_file`) call `DotenvxWrapper.encrypt` directly, bypassing that backend entirely.

- Extracts `is_dotenvx_safe_filename(name)` into `integrations/dotenvx.py` as a single shared predicate; `DotenvxEncryptionBackend` replaces its inline `re.fullmatch` with it so the two guards cannot drift.
- Adds `_validate_encryptable_filename` to `DotenvxWrapper.encrypt`, called on every platform before dotenvx is invoked, so unsafe filenames are refused with the plaintext byte-for-byte intact.
- Both partial-encryption callers now catch `DotenvxFilenameError` alongside `DotenvxError` and wrap it into a clean `PartialEncryptionError` instead of a raw traceback.

Safe to merge — the guard fires before any dotenvx invocation, no file mutation can occur on the unsafe-filename path, and two regression tests verify plaintext survival byte-for-byte through the real wrapper.

The change is a focused, pre-flight guard addition. The shared predicate is functionally identical to the inline regex it replaces (both use fullmatch on the same character class). The exception plumbing is correct: DotenvxFilenameError is unrelated to DotenvxError in the hierarchy, so widening the except tuple is non-redundant. Regression tests drive the real wrapper without mocking the seam under test and confirm the plaintext is untouched on refusal.

| Filename | Overview |
|----------|----------|
| src/envdrift/integrations/dotenvx.py | Adds is_dotenvx_safe_filename predicate and _validate_encryptable_filename guard in DotenvxWrapper.encrypt; regex is correct (fullmatch with [A-Za-z0-9._-]+), guard fires before dotenvx invocation on all platforms. |
| src/envdrift/encryption/dotenvx.py | Replaces inline re.fullmatch with shared is_dotenvx_safe_filename predicate — functionally identical, removes drift risk. |
| src/envdrift/core/partial_encryption.py | Imports DotenvxFilenameError and widens two except clauses to catch it alongside DotenvxError, wrapping both into PartialEncryptionError cleanly. |
| tests/unit/test_partial_encryption.py | Adds a parametrized regression test driving the real guard (no wrapper mock) for both push_secrets_only and encrypt_secret_file, asserting PartialEncryptionError is raised and plaintext survives byte-for-byte. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A([push_secrets_only / encrypt_secret_file]) --> B[DotenvxWrapper.encrypt]
    B --> C{env_file exists?}
    C -- No --> D[raise DotenvxError]
    C -- Yes --> E{is_dotenvx_safe_filename?}
    E -- No --> F[raise DotenvxFilenameError\nplaintext preserved]
    E -- Yes --> G{Windows + problematic\nfilename?}
    G -- Yes --> H[raise DotenvxFilenameError]
    G -- No --> I[normalize line endings\nclean headers\nrun dotenvx encrypt]
    I --> J{encrypt succeeded?}
    J -- No --> K[raise DotenvxError]
    J -- Yes --> L([file encrypted])
    F --> M[PartialEncryptionError.encrypt_failed]
    H --> M
    D --> M
    K --> M
```

<sub>Reviews (2): Last reviewed commit: ["refactor(encrypt): hoist slug-guard impo..."](https://github.com/jainal09/envdrift/commit/2f04ecbc420a1be2512080201d308cc7c3062d85) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36727309)</sub>

### DIFF
--- a/src/envdrift/encryption/dotenvx.py
+++ b/src/envdrift/encryption/dotenvx.py
@@ -143,8 +143,12 @@ class DotenvxEncryptionBackend(EncryptionBackend):
         # undecryptable: the original plaintext is destroyed and the secret is
         # locked out for good, and the plaintext-survival check below cannot
         # catch it (the value *is* encrypted, only the key name is unusable).
-        # Only [A-Za-z0-9._-] filenames round-trip safely (#443).
-        if not re.fullmatch(r"[A-Za-z0-9._-]+", env_file.name):
+        # Only [A-Za-z0-9._-] filenames round-trip safely (#443). The same
+        # predicate guards DotenvxWrapper.encrypt so the partial-encryption paths
+        # that bypass this backend are covered too (#467).
+        from envdrift.integrations.dotenvx import is_dotenvx_safe_filename
+
+        if not is_dotenvx_safe_filename(env_file.name):
             return EncryptionResult(
                 success=False,
                 message=(
--- a/src/envdrift/encryption/dotenvx.py
+++ b/src/envdrift/encryption/dotenvx.py
@@ -14,6 +14,7 @@ from envdrift.encryption.base import (
     EncryptionResult,
     EncryptionStatus,
 )
+from envdrift.integrations.dotenvx import is_dotenvx_safe_filename
 
 
 class DotenvxEncryptionBackend(EncryptionBackend):
@@ -146,8 +147,6 @@ class DotenvxEncryptionBackend(EncryptionBackend):
         # Only [A-Za-z0-9._-] filenames round-trip safely (#443). The same
         # predicate guards DotenvxWrapper.encrypt so the partial-encryption paths
         # that bypass this backend are covered too (#467).
-        from envdrift.integrations.dotenvx import is_dotenvx_safe_filename
-
         if not is_dotenvx_safe_filename(env_file.name):
             return EncryptionResult(
                 success=False,
--- a/tests/unit/test_partial_encryption.py
+++ b/tests/unit/test_partial_encryption.py
@@ -14,6 +14,7 @@ from envdrift.config import PartialEncryptionEnvironmentConfig
 from envdrift.core.partial_encryption import (
     PartialEncryptionError,
     combine_files,
+    encrypt_secret_file,
     file_has_assignment,
     has_plaintext_secret_value,
     is_file_encrypted,
@@ -180,8 +181,6 @@ def test_encrypt_secret_file_reencrypts_mixed_state(tmp_path: Path):
     encrypt_secret_file early-returned and never encrypted the newly-added
     plaintext value -> it leaked. This asserts dotenvx.encrypt IS invoked.
     """
-    from envdrift.core.partial_encryption import encrypt_secret_file
-
     secret_file = tmp_path / ".env.secret"
     secret_file.write_text('API_KEY="encrypted:abc..."\nNEW_LEAKED_SECRET=plaintext_leak_value\n')
     config = PartialEncryptionEnvironmentConfig(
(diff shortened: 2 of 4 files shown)


src/envdrift/encryption/dotenvx.py
```diff
@@ -143,8 +143,12 @@ def encrypt(
...
         # locked out for good, and the plaintext-survival check below cannot
         # catch it (the value *is* encrypted, only the key name is unusable).
-        # Only [A-Za-z0-9._-] filenames round-trip safely (#443).
-        if not re.fullmatch(r"[A-Za-z0-9._-]+", env_file.name):
+        # Only [A-Za-z0-9._-] filenames round-trip safely (#443). The same
+        # predicate guards DotenvxWrapper.encrypt so the partial-encryption paths
+        # that bypass this backend are covered too (#467).
+        from envdrift.integrations.dotenvx import is_dotenvx_safe_filename
+
+        if not is_dotenvx_safe_filename(env_file.name):
```
jainal09, line 150: Fixed in 2f04ecb: moved the `is_dotenvx_safe_filename` import to module level (integrations.dotenvx has no envdrift imports, so no cycle), so `DotenvxEncryptionBackend.encrypt` keeps its prior length and the inline guard is a single line — no added complexity.

tests/unit/test_partial_encryption.py
```diff
@@ -700,6 +700,69 @@ def test_push_secrets_only_raises_on_encrypt_failure(secrets_only_config, secret
             push_secrets_only(secrets_only_config)
 
 
+# ---------------------------------------------------------------------------
```
jainal09, line 702: Fixed in 2f04ecb: folded the two near-identical lockout tests into one parametrized test over both entry points, bringing the module back under the 75-function threshold.

tests/unit/test_partial_encryption.py
```diff
@@ -700,6 +700,69 @@ def test_push_secrets_only_raises_on_encrypt_failure(secrets_only_config, secret
...
+    # unrecoverable file (pre-fix this destroyed it and raised nothing).
+    assert bad.read_bytes() == before
+    assert b"encrypted:" not in bad.read_bytes()
+
+
+def test_encrypt_secret_file_refuses_unicode_named_file_preserves_plaintext(tmp_path: Path):
+    """#467: a non-ASCII secret filename is refused too, with plaintext intact."""
+    from envdrift.core.partial_encryption import encrypt_secret_file
+
+    secret_file = tmp_path / "café.env.secret"  # non-ASCII -> invalid key slug
```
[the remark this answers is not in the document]
jainal09: Done in 2f04ecb: hoisted `encrypt_secret_file` (and a redundant pre-existing function-local import) into the module's top-level import block.

PR #502, opened 2026-06-12
fix(scanner): stop trivy --skip-duplicate collapsing distinct findings (#479)

### LINKED ISSUE #479: guard: trivy adapter hashes the pre-redacted Match line, so --skip-duplicate drops distinct real secrets
## Summary

The trivy adapter derives `secret_hash` from trivy's `Match` field, which trivy returns with the secret already
redacted to asterisks. Two different secrets with the same variable name and length therefore produce identical
hashes, and under `--skip-duplicate` the engine's hash-keyed dedup silently drops one of two distinct real
secrets — a false negative in a security scan. The redacted Match also makes trivy's `secret_preview`
uninformative and prevents the documented cross-scanner duplicate collapse against native/gitleaks (which hash
the raw value). Every item below was reproduced twice (original reviewer + independent verifier) on origin/main
commit cb68c97, with the real trivy binary.

## Findings

- [ ] **`--skip-duplicate` collapses two distinct trivy-found secrets into one because `secret_hash` is computed
  from the redacted Match line.** `src/envdrift/scanner/trivy.py:551`
  **Repro:** `a.txt`/`b.txt` each with a different `ghp_<36 alnum>` token (same var name `TOKEN`, same length);
  commit; `envdrift guard --trivy --no-gitleaks --no-detect-secrets --no-auto-install --json .` → 2 findings;
  add `--skip-duplicate` → 1 finding.
  **Expected:** each distinct real secret stays a distinct finding under `--skip-duplicate`; dedup must key on the
  actual secret value, not a redacted line.
  **Actual:** trivy returns `Match = 'TOKEN=****...'` (`Secret: null`); `hash_secret(matched)` at trivy.py:551-553
  is identical for both, and the engine dedups on `secret_hash` alone (engine.py:612-617) — `b.txt`'s secret is
  silently suppressed.
  **Verifier note:** cause confirmed in code and live with real trivy; `secret_preview` also shows the redacted
  line rather than the secret, and trivy findings never collapse with the same secret found by native/gitleaks.

## Suggested fix direction

- In the trivy adapter, build `secret_hash` from fields that distinguish findings when only a redacted match is
  available (e.g. file path + line + column/offsets + rule id), or extract the raw matched span from the file at
  `StartLine`/`Code` offsets before hashing — never hash the redacted `Match` string as if it were the secret.
- Consider having the engine's `--skip-duplicate` dedup (engine.py:612-617) treat hashes derived from redacted
  content as non-collapsible across distinct locations.

_Part of the 2026-06 final deep-review (batch: GUARD-5). Each item was independently reproduced and re-verified before filing._

jainal09:
## Summary

The trivy adapter hashed trivy's `Match` field as if it were the secret — but trivy emits `Match`
with the secret already redacted to a same-length `*` run (`Secret` is always null). Two distinct
secrets with the same variable name and length therefore produced identical `secret_hash` values,
and the engine's hash-keyed `--skip-duplicate` dedup silently dropped one of two real secrets — a
false negative in a security scan.

## Fixes

Issue #479 has one checklist item; this PR resolves it and the verifier-noted side effects:

- **`--skip-duplicate` collapses two distinct trivy-found secrets (trivy.py:551)** — the adapter
  now never hashes the redacted `Match` line as the secret. `TrivyScanner._recover_secret_value`
  re-reads the flagged line from the scanned file and aligns it with `Match` (the longest common
  prefix/suffix bound the masked span); the recovered raw value is hashed instead. The exact issue
  repro (two `ghp_<36 alnum>` tokens, same var name/length, `--skip-duplicate`) now reports both.
- **Unrecoverable matches can never collapse** — when alignment cannot be validated (multi-line
  finding, truncated match, file changed or unreadable), `_location_qualified_hash` qualifies the
  hash with file + line span + rule + match, so distinct findings at different locations always
  stay distinct (the engine's `--skip-duplicate` keys on `secret_hash` alone), while re-scans of
  the same finding stay stable.
- **Verifier note: uninformative `secret_preview`** — with the raw value recovered, the preview is
  now `redact_secret(raw_secret)` (e.g. `ghp_****...EPCr`) instead of the fully-masked Match line.
- **Verifier note: no cross-scanner collapse** — the recovered-value hash matches what
  native/gitleaks compute, so the same secret found by trivy and another scanner collapses to one
  finding under `--skip-duplicate` (regression-tested against the real native scanner).
- **Docs** — `docs/cli/guard.md` documents the redacted-match handling under `--skip-duplicate`.

## Test evidence

Regression tests were written first and run against unfixed code (fail), then after the fix (pass).

Unit — `tests/scanner/test_trivy.py::TestRedactedMatchHashing` (11 tests, real files in `tmp_path`,
secret dicts shaped exactly like real `trivy fs --format json` output):

```
UNFIXED: 7 failed, 3 passed
  FAILED ...::test_distinct_secrets_with_identical_redacted_match_hash_differently
  FAILED ...::test_recovered_hash_and_preview_use_raw_secret_value
  FAILED ...::test_redacted_match_line_is_never_hashed_as_the_secret
  FAILED ...::test_fallback_hash_is_location_qualified_when_file_is_missing
  FAILED ...::test_fallback_when_file_line_no_longer_aligns_with_match
  FAILED ...::test_multi_line_secret_falls_back_to_location_hash
  FAILED ...::test_recovery_works_with_crlf_line_endings
FIXED:   all pass (file total: 61 passed)
```

Integration — `tests/integration/test_scanner_external_binaries.py` (real trivy binary, real
envdrift CLI subprocess, real `ScanEngine`; skip-gated on trivy being installed):

```
UNFIXED: 3 failed
  test_trivy_distinct_secrets_survive_skip_duplicate_cli
    AssertionError: --skip-duplicate must keep both distinct secrets, got ['a.txt']
  test_trivy_hash_and_preview_recover_raw_secret_from_real_output
    AssertionError: assert 'c9ce2958d662...' == '7b8072392191...'   (hash != hash of raw token)
  test_trivy_finding_collapses_with_native_for_same_secret_under_skip_duplicate
    AssertionError: a finding still hashes the redacted Match line as if it were the secret
FIXED:   all pass (file total: 11 passed)
```

Gates: `ruff check` / `ruff format --check` / `pyrefly` / `bandit` clean; full unit lane
`pytest -m "not integration"` 2536 passed; guard CLI e2e + engine dedup integration suites pass;
`make lint-docs` 0 errors; `src/envdrift/scanner/trivy.py` at 99% line coverage.

Fixes #479

<a href="https://cubic.dev/pr/jainal09/envdrift/pull/502?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **New Features**
  * Improved secret recovery from Trivy scans; distinct secrets with identical redaction patterns are now properly handled as separate findings instead of false duplicates.

* **Bug Fixes**
  * Enhanced `--skip-duplicate` deduplication logic to correctly identify unique secrets across scanners while preventing collapse of genuinely different values.

* **Documentation**
  * Added guidance on how secret deduplication behaves with Trivy findings and redacted matches.

<h3>Greptile Summary</h3>

This PR fixes a false-negative in trivy's secret deduplication: trivy emits `Match` with the secret already replaced by a same-length run of `*` (the `Secret` field is always null), so hashing `Match` as if it were the secret caused two distinct secrets of identical shape to share the same `secret_hash`, and `--skip-duplicate` silently dropped one of them.

- **Core fix in `trivy.py`**: A new `_ParseState` dataclass (shared across one `_parse_output` call) carries a per-parse `line_cache` and a `fallback_counts` occurrence map. `_hash_and_preview` attempts to recover the raw value via `_recover_secret_value` → `_align_redacted_match` (longest-common-prefix/suffix alignment with asterisk-boundary ambiguity guards); on success, it hashes and redacts the raw value so cross-scanner dedup works correctly.
- **Fallback path**: when recovery is not possible (multi-line finding, `EndLine: null`, file unreadable, line changed, ambiguous `*` boundary), `_location_qualified_hash` qualifies the hash with file + line span + rule + a per-location occurrence index, ensuring byte-identical trivy dicts for two same-line same-rule secrets never collapse.
- **Tests**: 11 unit tests in the new `test_trivy_dedup.py` and 4 `@requires_trivy` integration tests covering the exact #479 repro, cross-scanner collapse, same-line fallback stability, and CRLF recovery.

Safe to merge — the change is isolated to the trivy adapter's hash/preview computation, all previously raised review concerns are addressed, and the logic is well-tested.

The fix is self-contained: it replaces a single wrong hash computation with a well-guarded recovery path and a deterministic fallback. Every edge case (null EndLine, CRLF, edge-asterisk ambiguity, file changed, same-line same-rule duplicates) is covered by unit tests that were confirmed failing before the fix and passing after. All prior review thread items are resolved in this diff.

| Filename | Overview |
|----------|----------|
| src/envdrift/scanner/trivy.py | Core fix: adds _ParseState dataclass, _hash_and_preview, _recover_secret_value, _align_redacted_match, _read_lines, and _location_qualified_hash to correctly compute secret_hash and secret_preview from trivy's pre-redacted Match field. |
| tests/scanner/test_trivy_dedup.py | New unit test module with 11 tests covering the full space of hash/preview recovery: distinct-token collision, same-line same-rule fallback with occurrence index, CRLF recovery, edge-asterisk ambiguity guard, null EndLine, and stability across re-parses. |
| tests/integration/test_scanner_external_binaries.py | Adds four @requires_trivy integration tests exercising the exact #479 repro through the real CLI, the real trivy binary, and the real ScanEngine. |
| src/envdrift/scanner/engine.py | One-line doc update to the public-key filter comment to mention trivy now hashes with hash_secret once raw value is recovered. |
| docs/cli/guard.md | Adds a paragraph documenting how trivy's pre-redacted Match is handled under --skip-duplicate, including the fallback to per-occurrence location-qualified hashes. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["_parse_output(scan_data, base_path)"] --> B["Create _ParseState\n(line_cache, fallback_counts)"]
    B --> C["For each result → each secret"]
    C --> D["_parse_secret(secret, target, base_path, state)"]
    D --> E["_hash_and_preview(secret, file_path, rule_id, state)"]
    E --> F["_recover_secret_value(file_path, secret, state.line_cache)"]
    F --> G{Single-line? Has '*'?\nValid start_line?}
    G -->|No| H["Return None (fallback)"]
    G -->|Yes| I["_read_lines via line_cache"]
    I --> J["_align_redacted_match(raw_line, match)"]
    J --> K{Length match?\nUnambiguous '*' boundary?\nMasked span all '*'?}
    K -->|No| H
    K -->|Yes| L["raw_secret = candidate span"]
    L --> M["hash_secret(raw_secret)\nredact_secret(raw_secret)"]
    H --> N{match empty?}
    N -->|Yes| O["Return '', ''"]
    N -->|No| P["key = (file, start, end, rule, match)\noccurrence = fallback_counts[key]++"]
    P --> Q["_location_qualified_hash(key, occurrence)"]
    M --> R["ScanFinding"]
    Q --> R
    O --> R
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A["_parse_output(scan_data, base_path)"] --> B["Create _ParseState\n(line_cache, fallback_counts)"]
    B --> C["For each result → each secret"]
    C --> D["_parse_secret(secret, target, base_path, state)"]
    D --> E["_hash_and_preview(secret, file_path, rule_id, state)"]
    E --> F["_recover_secret_value(file_path, secret, state.line_cache)"]
    F --> G{Single-line? Has '*'?\nValid start_line?}
    G -->|No| H["Return None (fallback)"]
    G -->|Yes| I["_read_lines via line_cache"]
    I --> J["_align_redacted_match(raw_line, match)"]
    J --> K{Length match?\nUnambiguous '*' boundary?\nMasked span all '*'?}
    K -->|No| H
    K -->|Yes| L["raw_secret = candidate span"]
    L --> M["hash_secret(raw_secret)\nredact_secret(raw_secret)"]
    H --> N{match empty?}
    N -->|Yes| O["Return '', ''"]
    N -->|No| P["key = (file, start, end, rule, match)\noccurrence = fallback_counts[key]++"]
    P --> Q["_location_qualified_hash(key, occurrence)"]
    M --> R["ScanFinding"]
    Q --> R
    O --> R
```

</a>

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/479-trivy-d..."](https://github.com/jainal09/envdrift/commit/8d6fce6ea6fc96821b31782cb14ff6394b00a087) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37198099)</sub>


src/envdrift/scanner/trivy.py
```diff
@@ -579,3 +594,77 @@ def _parse_secret(
...
+
+        The redacted ``Match`` line alone is identical for two distinct
+        secrets of the same shape, so it must be qualified with the finding's
+        file, line span and rule: distinct findings then can never collapse
+        under the engine's hash-keyed ``--skip-duplicate`` dedup, while
+        re-scans of the same finding stay stable. The constant prefix and
+        NUL separators keep this synthetic key from ever colliding with a
+        real secret's ``hash_secret`` value.
+        """
+        parts = (
```
jainal09: 🤖 **envdrift review-bot** · medium · bug


Trivy censors **all** secret spans in a line before building each finding's `Match` (its censored-content model never leaks one finding's secret in another finding's Match). So when one line holds two distinct same-rule secrets (the exact #348 scenario: "two AWS keys on one `.env` line"), trivy emits two findings whose dicts are **byte-identical**: same `StartLine`/`EndLine`/`RuleID` and the same fully-censored `Match` (e.g. `T1=**** T2=****`).

I extracted `_recover_secret_value` and `_location_qualified_hash` verbatim into /tmp and ran this case:

- Recovery is rejected for both findings (the differing span `****... T2=...****` contains the unmasked ` T2=` text, so `set(masked_span) != {"*"}` → `None`). Correct so far.
- Both findings then take `_location_qualified_hash`, whose `parts = ("envdrift-trivy-redacted", str(file_path), StartLine, EndLine, rule_id, matched)` are identical for both → **identical `secret_hash`**.

Result: under `--skip-duplicate` (engine `_dedup_key` returns `(finding.secret_hash,)`) AND under the default key `(file, line, rule, secret_hash)`, the two distinct real secrets collapse to ONE finding — the same false-negative class #479 fixes, surviving in the same-line case. No deterministic pure function of the (identical) finding dicts can distinguish them, so this needs an occurrence index: e.g. in `_parse_output`, count repeats of identical `(file, StartLine, EndLine, RuleID, Match)` tuples and feed the index into `_location_qualified_hash` so N identical-looking trivy findings keep N distinct hashes.

This also makes two shipped docs inaccurate:
- `docs/cli/guard.md:199` (added in this PR): "the finding keeps a location-qualified hash that **never collapses with any other finding**" — false for this case.
- `engine.py` `_deduplicate` docstring: "Two *distinct* secrets matching the same rule on the same line (e.g. two AWS keys on one ``.env`` line, #348) are both reported" — now false for trivy fallback findings.

No test covers the same-line two-secrets case either (the new suite tests same-shape secrets only in *separate* files). Suggested direction: add the occurrence-index qualifier + a regression test planting two same-rule tokens on one line, and soften/fix the guard.md sentence. If you choose not to fix here, per repo convention add an `xfail(strict=False)` test referencing a filed issue.

src/envdrift/scanner/trivy.py
```diff
@@ -579,3 +594,77 @@ def _parse_secret(
...
+        except OSError:
+            return None
+        lines = text.splitlines()
+        if start_line > len(lines):
+            return None
+        raw_line = lines[start_line - 1]
+        length = len(match)
+        if len(raw_line) != length or raw_line == match:
+            return None
+        prefix = 0
```
jainal09, line 715: 🤖 **envdrift review-bot** · low · bug


The prefix/suffix scan in `_recover_secret_value` advances while `raw_line[i] == match[i]`. Inside trivy's masked span, `match[i]` is `'*'` — so if the raw secret itself starts or ends with literal `*` characters (possible for generic-api-key/password-style rules), the scan walks **into** the masked span and the recovered `candidate` is the secret with its leading/trailing `*` chars stripped. The remaining `masked_span` is still all `'*'`, so validation passes and the truncated value is accepted as recovered.

Verified by running the extracted algorithm: raw line `PASSWORD=*abc-secret-xyz*` with Match `PASSWORD=` + 16x`*` recovers `abc-secret-xyz` (not `*abc-secret-xyz*`).

Impact: `secret_hash = hash_secret(truncated)` no longer equals what native/gitleaks compute for the same secret, so the documented cross-scanner collapse under `--skip-duplicate` (guard.md and the PR's own headline behavior) silently fails for these secrets — the same secret shows as two findings. (A distinct-secret hash collision is also constructible but contrived.) Findings still stay distinct per value, so no false negative — this is a correctness gap in the recovery contract, not a dropped finding.

Suggested direction: treat the alignment as ambiguous (return `None` → location-qualified fallback) when the characters immediately adjacent to the differing region are `'*'` on both sides in `match` and `raw_line` — i.e. stop trusting recovery when the mask boundary cannot be determined. Add a unit test with a `*`-bracketed secret value.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 4faf391


The recovery + fallback-hash design is solid for the cross-file cases it targets, but two verified edge cases remain in the trivy redacted-match handling: identical same-line same-rule secrets still collapse under the fallback hash (the residual #348/#479 class, contradicting the guard.md claim added here), and boundary `*` characters in a secret silently truncate the recovered value.

**Findings (2):**
- [medium] Two distinct same-rule secrets on the SAME line still collapse to one finding (fallback hash is identical for both) — `src/envdrift/scanner/trivy.py:662` (inline)
- [low] Alignment consumes literal '*' at the secret's boundaries, recovering a truncated value and breaking cross-scanner hash collapse for such secrets — `src/envdrift/scanner/trivy.py:634` (inline)

src/envdrift/scanner/trivy.py
jainal09: Fixed in c38e45c. Verified against the real trivy 0.72.0 binary first: two distinct same-rule PATs on one line do produce byte-identical finding dicts (same StartLine/EndLine/RuleID, same fully-censored Match). `_location_qualified_hash` now takes a per-location occurrence index threaded from `_parse_output` (`fallback_counts` keyed on file + line span + rule + Match), so N byte-identical unrecoverable findings keep N distinct hashes while re-parses stay stable (trivy output order is deterministic). Regression tests: `test_two_distinct_same_rule_secrets_on_one_line_keep_distinct_hashes` (unit, via `_parse_output`, asserts distinct + stable hashes) and `test_trivy_two_distinct_secrets_on_one_line_keep_distinct_hashes` (integration, real binary, one-line two-token file). The guard.md sentence is rewritten to describe the occurrence-qualified fallback instead of the old overclaim; the engine.py `_deduplicate` docstring claim is accurate again now that the same-line findings carry distinct hashes.

src/envdrift/scanner/trivy.py
jainal09: Fixed in c38e45c, following the suggested direction: `_recover_secret_value` now rejects the alignment as ambiguous (returns None, so the caller uses the location-qualified fallback) whenever the character immediately adjacent to the differing span is `*` — i.e. the common prefix/suffix may have walked into the masked run. Your repro (`PASSWORD=*abc-secret-xyz*` with `PASSWORD=` + 16x`*`) was reproduced before the fix (recovered the truncated `abc-secret-xyz`) and now falls back. Regression test: `test_secret_with_literal_edge_asterisks_is_not_silently_truncated` asserts the hash is non-empty and is NOT `hash_secret("abc-secret-xyz")`.

src/envdrift/scanner/trivy.py
```diff
@@ -579,3 +594,77 @@ def _parse_secret(
...
+        match = secret.get("Match") or ""
+        start_line = secret.get("StartLine")
+        end_line = secret.get("EndLine", start_line)
+        if (
+            not match
+            or "*" not in match
+            or not isinstance(start_line, int)
+            or start_line < 1
+            or (isinstance(end_line, int) and end_line != start_line)
+        ):
```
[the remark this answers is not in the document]
jainal09: Fixed in c38e45c: the guard now requires `isinstance(end_line, int) and end_line == start_line`, so an explicit `"EndLine": null` is treated as an unknown (multi-line) span and recovery is rejected instead of assumed single-line. Regression test: `test_explicit_endline_null_is_treated_as_multi_line_and_falls_back` (asserts the finding falls back to the location-qualified hash rather than recovering the raw value).

src/envdrift/scanner/trivy.py
```diff
@@ -579,3 +594,77 @@ def _parse_secret(
...
+        suffix = 0
+        while (
+            suffix < length - prefix and raw_line[length - 1 - suffix] == match[length - 1 - suffix]
+        ):
+            suffix += 1
+        masked_span = match[prefix : length - suffix]
+        candidate = raw_line[prefix : length - suffix]
+        if not candidate or set(masked_span) != {"*"}:
+            return None
+        return candidate
```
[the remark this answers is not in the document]
jainal09: Fixed in c38e45c: `_parse_output` now builds a per-parse `line_cache: dict[Path, list[str] | None]` shared across all findings and threads it into `_parse_secret` / `_recover_secret_value`, so n findings in one file cost a single read. Scoped per parse (not `lru_cache`) so a file changing between scans can never serve stale lines. Exercised by the new same-line test where two findings hit the same cached file.

tests/integration/test_scanner_external_binaries.py
```diff
@@ -379,3 +379,134 @@ def test_scan_engine_filters_dotenvx_public_keys_from_real_output(tmp_path):
...
+            ".",
+        ],
+        cwd=str(tmp_path),
+        env=env,
+        capture_output=True,
+        text=True,
+    )
+
+    out = proc.stdout
+    payload = _json.loads(out[out.index("{") : out.rindex("}") + 1])
```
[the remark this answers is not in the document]
jainal09: Fixed in c38e45c: the brace slicing is gone — the test now calls `json.loads(proc.stdout)` strictly (with a `pytest.fail` that dumps stdout/stderr on `JSONDecodeError`). This also enforces the repo contract that `guard --json` stdout is a single clean JSON document with no ANSI/prose. Verified passing against the real trivy 0.72.0 binary.

tests/integration/test_scanner_external_binaries.py
```diff
@@ -379,3 +379,134 @@ def test_scan_engine_filters_dotenvx_public_keys_from_real_output(tmp_path):
...
+            "--no-auto-install",
+            "--skip-duplicate",
+            "--json",
+            ".",
+        ],
+        cwd=str(tmp_path),
+        env=env,
+        capture_output=True,
+        text=True,
+    )
```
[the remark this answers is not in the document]
jainal09: Fixed in c38e45c: added `timeout=300` to the subprocess.run call (generous headroom for CI cold starts; a hang now fails the test cleanly via TimeoutExpired instead of stalling the job). Verified passing against the real trivy 0.72.0 binary.

tests/scanner/test_trivy.py
```diff
@@ -706,6 +707,226 @@ def fake_download(url: str, dest: str) -> None:
...
+    def test_multi_line_secret_falls_back_to_location_hash(
+        self, scanner: TrivyScanner, tmp_path: Path
+    ):
+        """Multi-line findings (StartLine != EndLine) never hash a partial span."""
+        (tmp_path / "key.pem").write_text(
+            "-----BEGIN PRIVATE KEY-----\nabc\n-----END PRIVATE KEY-----\n",
+            encoding="utf-8",
+        )
+        masked = "-----BEGIN PRIVATE KEY-----"
+        finding_a = scanner._parse_secret(
```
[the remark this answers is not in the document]
jainal09: Fixed in c38e45c: the BEGIN/END private-key markers are now built by concatenation (`"-----BEGIN " + "PRIVATE KEY-----"`), matching the token-fixture approach, so no key-shaped literal appears whole in source. (The test now lives in tests/scanner/test_trivy_dedup.py — see the CodeScene threads.)

tests/scanner/test_trivy.py
```diff
@@ -11,6 +11,7 @@
 import pytest
 
 from envdrift.scanner.base import FindingSeverity
+from envdrift.scanner.patterns import hash_secret, redact_secret
```
jainal09: Addressed in c38e45c: the entire #479 hashing/dedup suite (`TestRedactedMatchHashing` plus its token constants) moved to a new focused module, tests/scanner/test_trivy_dedup.py (precedent: test_engine_dedup.py). test_trivy.py drops from 947 to 726 physical lines, putting its LOC back under the 600 threshold.

jainal09: Addressed in c38e45c: the #479 secret-hash recovery/dedup tests were split out into tests/scanner/test_trivy_dedup.py, so test_trivy.py is back to its original install/scan-plumbing responsibilities and the new module owns a single concern.

src/envdrift/scanner/trivy.py
```diff
@@ -518,21 +525,38 @@ def _parse_output(
...
         return findings, files_scanned
 
     def _parse_secret(
-        self, secret: dict[str, Any], target: str, base_path: Path
+        self,
+        secret: dict[str, Any],
+        target: str,
+        base_path: Path,
+        fallback_counts: dict[tuple[str, str, str, str, str], int] | None = None,
+        line_cache: dict[Path, list[str] | None] | None = None,
```
jainal09: Addressed in f8a1836: the hash/preview branch moved out of `_parse_secret` into `_hash_and_preview`, the two threaded dicts collapsed into a `_ParseState` dataclass (so `_parse_secret` is back to 4 args and one nested block), `_recover_secret_value` was split into `_read_lines` + `_align_redacted_match`, and `_location_qualified_hash` now takes the prebuilt fallback-identity tuple plus the occurrence index (2 args). Behavior unchanged; all trivy suites and the real-binary integration tests still pass.

jainal09: Addressed in f8a1836: the hash/preview branch moved out of `_parse_secret` into `_hash_and_preview`, the two threaded dicts collapsed into a `_ParseState` dataclass (so `_parse_secret` is back to 4 args and one nested block), `_recover_secret_value` was split into `_read_lines` + `_align_redacted_match`, and `_location_qualified_hash` now takes the prebuilt fallback-identity tuple plus the occurrence index (2 args). Behavior unchanged; all trivy suites and the real-binary integration tests still pass.

src/envdrift/scanner/trivy.py
```diff
@@ -507,6 +507,13 @@ def _parse_output(
         findings: list[ScanFinding] = []
         files_scanned = 0
 
+        # Shared per-parse state: occurrence counts so byte-identical
```
jainal09: Addressed in f8a1836: the hash/preview branch moved out of `_parse_secret` into `_hash_and_preview`, the two threaded dicts collapsed into a `_ParseState` dataclass (so `_parse_secret` is back to 4 args and one nested block), `_recover_secret_value` was split into `_read_lines` + `_align_redacted_match`, and `_location_qualified_hash` now takes the prebuilt fallback-identity tuple plus the occurrence index (2 args). Behavior unchanged; all trivy suites and the real-binary integration tests still pass.

src/envdrift/scanner/trivy.py
jainal09: Addressed in f8a1836: the hash/preview branch moved out of `_parse_secret` into `_hash_and_preview`, the two threaded dicts collapsed into a `_ParseState` dataclass (so `_parse_secret` is back to 4 args and one nested block), `_recover_secret_value` was split into `_read_lines` + `_align_redacted_match`, and `_location_qualified_hash` now takes the prebuilt fallback-identity tuple plus the occurrence index (2 args). Behavior unchanged; all trivy suites and the real-binary integration tests still pass.

src/envdrift/scanner/trivy.py
```diff
@@ -579,3 +630,109 @@ def _parse_secret(
...
+        parts = (
+            "envdrift-trivy-redacted",
+            str(file_path),
+            str(secret.get("StartLine") or 0),
+            str(secret.get("EndLine") or 0),
+            rule_id,
+            matched,
+            str(occurrence),
+        )
+        return hash_secret("\x00".join(parts))
```
jainal09: Addressed in f8a1836: the hash/preview branch moved out of `_parse_secret` into `_hash_and_preview`, the two threaded dicts collapsed into a `_ParseState` dataclass (so `_parse_secret` is back to 4 args and one nested block), `_recover_secret_value` was split into `_read_lines` + `_align_redacted_match`, and `_location_qualified_hash` now takes the prebuilt fallback-identity tuple plus the occurrence index (2 args). Behavior unchanged; all trivy suites and the real-binary integration tests still pass.

jainal09:
@coderabbitai full review

PR #504, opened 2026-06-12
fix(agent): make auto-encryption pipeline functional end-to-end (#481)

### LINKED ISSUE #481: agent: auto-encryption pipeline non-functional end-to-end (lock argv, kqueue self-block, config crash, skips)
## Summary

The Go agent's core promise — watch registered projects and auto-encrypt idle `.env` files — is broken at every
stage of the pipeline: project discovery silently drops registered projects, the documented global config crashes
the agent at startup, macOS lock detection blocks forever on the agent's own watcher fds, mixed-state files are
dropped from tracking, and the final encrypt step is an argv the CLI rejects. Net effect: no agent user gets any
file encrypted, while registration and status output report success and plaintext secrets sit on disk. Every item
below was reproduced twice (original reviewer + independent verifier) on origin/main commit cb68c97.

## Findings

- [ ] **The agent's encrypt step can never succeed: it runs `envdrift lock <file>` (usage error, exit 2), and the
  Python fallback `python -m envdrift lock <file>` also fails (no `__main__.py`, exit 1).**
  `envdrift-agent/internal/encrypt/encrypt.go:94`
  **Repro:** harness `main` calling `encrypt.Encrypt(file)` in a module copy; `go build`; run against a plaintext
  `.env.production` with the real `envdrift` on PATH.
  **Expected:** the encrypt step succeeds for an idle plaintext env file (e.g. `envdrift encrypt <file>`).
  **Actual:** exit 2 ("Got unexpected extra argument(s)"), file unchanged; the python fallback is doubly dead.
  **Verifier note:** `encrypt_test.go:152` asserts the broken argv tail rather than real behavior — why CI never
  caught it.

- [ ] **macOS: lock detection sees the agent's OWN fsnotify/kqueue fd, so every watched file is permanently "still
  open" and encryption is never attempted.** `envdrift-agent/internal/lockcheck/lockcheck.go:69`
  **Repro:** project with `[guardian] enabled = true`, `idle_timeout = "1s"`; run the agent foreground on macOS;
  append to `.env`; each 30s tick logs "File still open, skipping"; `lsof -- <project>/.env` shows envdrift-agent
  itself holding the fd.
  **Expected:** lock detection ignores the agent's own fds; idle files actually get encrypted.
  **Actual:** `isFileOpenUnix` returns true on any `lsof` output with no self-PID filter; a PATH-injected fake
  `envdrift` recorded zero invocations.
  **Verifier note:** after killing the agent the same `lsof` exits 1 — the agent held the only fd.

- [ ] **One encrypted value anywhere makes `IsEncrypted` return true, and `checkIdleFiles` then drops the file from
  tracking — plaintext secrets added to an already-encrypted file are never re-encrypted.**
  `envdrift-agent/internal/encrypt/encrypt.go:44`
  **Repro:** `.env` with a `DOTENV_PUBLIC_KEY` header, one `KEY="encrypted:..."` value and one plaintext
  `NEW_SECRET=...`; `encrypt.IsEncrypted(path)` → true; `guardian.go:420-423` does `RemoveFile` + `continue`.
  **Expected:** any plaintext non-comment value means not-fully-encrypted; hand the file to the encrypt step.
  **Actual:** the file is silently dropped with no encrypt attempt, log, or notification; the secret stays plaintext.
  **Verifier note:** the Python CLI handles partial encryption; this Go gate is an independent defect with no
  mixed-state handling.

- [ ] **The documented `idle_timeout = "5m"` in guardian.toml crashes the agent at startup (TOML string cannot decode
  into `time.Duration`); `Save` writes raw nanoseconds instead.** `envdrift-agent/internal/config/config.go:21`
  **Repro:** `printf '[guardian]\nenabled = true\nidle_timeout = "5m"\n' > ~/.envdrift/guardian.toml;` then
  `envdrift-agent start`
  **Expected:** the documented duration-string form parses (the per-project parser already handles it via
  `IdleTimeoutStr` + `ParseIdleTimeout`).
  **Actual:** `start` and `config` exit 1 with a Go decode error; a fresh `config` run writes
  `idle_timeout = 300000000000`.
  **Verifier note:** under launchd `KeepAlive=true` / systemd `Restart=always` this becomes a perpetual
  crash-respawn loop.

- [ ] **Projects registered via `pyproject.toml [tool.envdrift.guardian]` or a parent-dir `envdrift.toml` are silently
  never watched: the Go agent only reads `<path>/envdrift.toml`.** `envdrift-agent/internal/project/config.go:67`
  **Repro:** project whose guardian config lives in `pyproject.toml`; `envdrift agent register <path>` prints
  "✓ Registered" with no warning; agent log shows "Loading N registered projects" but no "Watching project:" line.
  **Expected:** the agent honors the same config-discovery contract the CLI used to bless the registration, or
  register/list warn that the agent will ignore the project.
  **Actual:** CLI reports success and `agent list` shows "Has Config ✓"; the agent drops the project with zero
  log output.
  **Verifier note:** reproduced end-to-end with 3 registrations; only the local-envdrift.toml project was watched.

## Suggested fix direction

- `buildEncryptCommand` (`encrypt.go:78-97`): invoke `envdrift encrypt <file>` (which takes a positional ENV_FILE)
  or a lock form that exists; fix `encrypt_test.go:152` to assert behavior against the real CLI contract.
- `isFileOpenUnix` (`lockcheck.go:38-69`): use `lsof -t` and filter out `os.Getpid()` so the agent's own kqueue fds
  don't count as "file open".
- `encrypt.IsEncrypted`: return false when any non-comment value lacks the `encrypted:` prefix (mirror the CLI's
  partial-encryption awareness), or at minimum log/notify the skipped-with-plaintext case.
- `config.Load`/`Save` (`config.go:21,80,98`): decode duration strings using the `IdleTimeoutStr` + `ParseIdleTimeout`
  pattern from `project/config.go:34,105-112`, and serialize back as a string.
- `LoadProjectConfig` (`project/config.go:67`): implement pyproject/parent-dir discovery matching
  `src/envdrift/config.py find_config`, or add a registration-time warning in `cli_commands/agent.py`.

_Part of the 2026-06 final deep-review (batch: AGENT-1). Each item was independently reproduced and re-verified before filing._

jainal09:
## Summary

Issue #481: the Go background agent's core promise — watch registered projects and auto-encrypt idle
`.env` files — was broken at every stage of the pipeline, so no agent user ever got a file encrypted
while registration and status output reported success. This PR fixes all five defects in one themed
change, each with a regression test that fails on the unfixed code, plus a real-CLI contract test and
an end-to-end test against the real `envdrift` binary.

## Fixes

- **Encrypt argv** (`internal/encrypt/encrypt.go`): the agent ran `envdrift lock <file>`, but `lock`
  takes no positional argument — every invocation exited 2. It now runs `envdrift encrypt <file>`,
  the CLI's per-file encryption path (positional `ENV_FILE`). The existing unit tests that asserted
  the broken argv now assert the correct one, and a new contract test validates the emitted
  subcommand against the **real** CLI's `--help` (a future CLI change that drops the positional
  breaks it loudly). `TestEncryptEndToEnd_RealCLI` drives the real binary end-to-end; both
  skip-gate cleanly when `envdrift`/`dotenvx` are absent (e.g. the cross-OS Go CI matrix).
- **Dead Python fallback** (`src/envdrift/__main__.py`, new): the agent's documented
  `python -m envdrift ...` fallback always failed with "No module named envdrift.\_\_main\_\_"
  (exit 1). Added the module entry point plus a real subprocess test.
- **macOS self-block** (`internal/lockcheck/lockcheck.go`): the fsnotify/kqueue watcher holds an fd
  on every watched file, so the lsof check saw the agent's own descriptor and reported every file
  permanently "still open". Lock detection now uses `lsof -t` and excludes `os.Getpid()`; only other
  processes' handles count. The PID lister is an injectable seam so the exclusion logic is
  unit-tested on all platforms; conservative behavior on missing/failing `lsof` is preserved.
- **Mixed-state files dropped** (`internal/encrypt/encrypt.go`): `IsEncrypted` returned true on the
  first `encrypted:` value, so a plaintext secret added to an already-encrypted file made the
  guardian drop the file from tracking without re-encrypting. The predicate now requires **every**
  value to be ciphertext, exempting `DOTENV_PUBLIC_KEY*` and SOPS metadata keys (mirrors
  `src/envdrift/core/partial_encryption.py`).
- **Config crash loop** (`internal/config/config.go`): the documented `idle_timeout = "5m"` failed
  go-toml decoding into `time.Duration`, crashing the agent at startup (a perpetual respawn loop
  under launchd `KeepAlive` / systemd `Restart=always`), and `Save` wrote raw nanoseconds. Load now
  accepts the documented duration strings (same `ParseIdleTimeout` as the per-project parser,
  including `"2d"`) **and** the legacy nanosecond integers older Saves wrote; Save writes the
  documented string form.
- **Projects silently never watched** (`internal/project/config.go`): the agent read only
  `<path>/envdrift.toml`, ignoring projects registered via `pyproject.toml [tool.envdrift.guardian]`
  or a parent-dir `envdrift.toml` — both blessed by the CLI's `find_config` at registration time.
  Discovery now mirrors that contract: walk up to the filesystem root; per level `envdrift.toml`
  wins, then a `pyproject.toml` with `[tool.envdrift]`; malformed `pyproject.toml` files are
  skipped like the CLI does.

Docs updated in the same PR: `docs/guides/agent-setup.md`, `envdrift-agent/README.md`,
`docs/agent-spec.md`.

Regression tests run against unfixed `origin/main` (e93c954) sources (real `envdrift` 10.16.0 +
`dotenvx` + `lsof` on PATH, macOS):

```text
--- FAIL: TestIsEncrypted/mixed_state:_encrypted_values_plus_fresh_plaintext_secret
        IsEncrypted(...) = true, expected false
--- FAIL: TestBuildEncryptCommand_PythonUsesModuleFlag
        args tail = [-m envdrift lock .env], want [-m envdrift encrypt .env]
--- FAIL: TestBuildEncryptCommand_BinaryNoModuleFlag
        args = [.../envdrift lock .env], want [.../envdrift encrypt .env]
--- FAIL: TestBuildEncryptCommand_MatchesRealCLIContract
        `envdrift lock` does not accept a positional ENV_FILE argument (#481)
--- FAIL: TestEncryptEndToEnd_RealCLI
        EncryptSilent(.../.env) failed against the real CLI: exit status 2
--- FAIL: TestIsFileOpenIgnoresOwnProcess
        IsFileOpen = true for a file held open only by our own process; want false (#481)
--- FAIL: TestLoadDocumentedDurationString
        Load() rejected the documented duration-string form (#481): toml: cannot decode TOML
        string into struct field config.GuardianConfig.IdleTimeout of type time.Duration
--- FAIL: TestSaveWritesDocumentedDurationString
        saved idle_timeout = 300000000000 (int64); want "5m" (#481)
--- FAIL: TestLoadProjectConfig_PyprojectGuardian
        Enabled = false: pyproject.toml [tool.envdrift.guardian] was ignored (#481)
--- FAIL: TestLoadProjectConfig_ParentDirEnvdriftToml
        Enabled = false: parent-dir envdrift.toml was ignored (#481)
--- FAIL: TestLoadProjectConfig_PyprojectWinsOverParentEnvdriftToml
--- FAIL: TestLoadProjectConfig_PyprojectWithoutEnvdriftSkipped
--- FAIL: TestLoadProjectConfig_MalformedPyprojectSkipped
--- FAIL: TestLoadProjectConfig_PyprojectVaultMappings
```

Python fallback before the fix:

```text
$ python -m envdrift --version
.../python: No module named envdrift.__main__; 'envdrift' is a package and cannot be directly executed
exit=1
```

With the fix, the full suites are green:

```text
go test -race ./...      → ok (all 10 packages, incl. TestEncryptEndToEnd_RealCLI passing
                           against the real envdrift + dotenvx; skips cleanly without them)
go vet ./... / gofmt -l . / golangci-lint v2.12.2 run ./...   → clean / empty / 0 issues
uv run pytest -m "not integration"   → 2525 passed, 4 skipped
ruff check + format / pyrefly / bandit / make lint-docs       → all green
```

Fixes #481

<a href="https://cubic.dev/pr/jainal09/envdrift/pull/504?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Project settings are now discovered more flexibly, including support for configuration in either project or parent directories.
  * File encryption now uses the current `encrypt` command flow and respects project-level settings.

* **Bug Fixes**
  * Improved detection of already-encrypted files, including mixed plaintext/encrypted content and common comment formats.
  * Fixed file-lock checks so the agent ignores its own open handles and handles unavailable system tools more safely.
  * `idle_timeout` values are now saved in a clearer, documented duration format while still supporting older saved values.

<h3>Greptile Summary</h3>

This PR fixes five independent bugs that rendered the agent's auto-encryption pipeline completely non-functional, and adds regression tests for each fix. All changes are tightly scoped to the affected packages.

- **encrypt.go**: fixes the CLI subcommand (`lock` → `encrypt`), extends `IsEncrypted` to require every secret value to be ciphertext (not just the first), and broadens SOPS ciphertext detection to all `ENC[` algorithms.
- **lockcheck.go**: excludes the agent's own PID from the open-file check (fsnotify holds a kqueue fd on every watched file), refactored around an injectable seam for cross-platform unit testing.
- **config.go**: adds a two-phase load (raw `any` + merge) that accepts both the documented `\"5m\"` string form and the legacy nanosecond integer, saving always writes the documented string.
- **project/config.go**: mirrors the CLI's `find_config` walk (up to filesystem root, `envdrift.toml` wins then `pyproject.toml [tool.envdrift]`), and surfaces load errors in `LoadAllProjectConfigs` instead of silently dropping projects.

All five pipeline bugs are correctly fixed with proper conservative fallbacks preserved; the changes are safe to merge.

Every fix is well-scoped to the broken path, backwards-compatible (legacy nanosecond configs still load, the self-PID exclusion is additive), and backed by targeted regression tests that demonstrate the before/after behavior. The refactored lsof path preserves the existing conservative treat-as-open policy for all ambiguous failures. The config discovery walk correctly stops at the filesystem root and matches the CLI contract. No behavioral changes outside the documented bug fixes were introduced.

| Filename | Overview |
|----------|----------|
| envdrift-agent/internal/encrypt/encrypt.go | Fixes `lock` → `encrypt` subcommand, expands `IsEncrypted` to require fully-encrypted files, broadens SOPS detection to all `ENC[` algo variants, adds inline-comment stripping for quoted values. Logic is correct and well-tested. |
| envdrift-agent/internal/lockcheck/lockcheck.go | Refactors lsof invocation to use `-t` (terse PID-per-line output) and excludes the agent's own PID from the open check. Injectable `openPIDs` seam enables cross-platform unit tests. Conservative error handling preserved. |
| envdrift-agent/internal/config/config.go | Two-phase load (unmarshal to `rawConfig`, then merge) fixes crash-on-string-duration. Pointer-to-slice fields correctly distinguish absent key from explicit empty array. `Save` serializes via `savedConfig` to write human-readable duration strings. |
| envdrift-agent/internal/project/config.go | Adds `discoverEnvdriftConfig` that walks toward filesystem root matching CLI's `find_config` contract. Handles both `envdrift.toml` and `pyproject.toml [tool.envdrift]`. Load errors are now logged before the project is skipped. |
| envdrift-agent/internal/encrypt/encrypt_test.go | Extensive regression tests for all `IsEncrypted` scenarios (mixed-state, SOPS variants, inline comments, empty assignments), plus a real-CLI contract test and end-to-end test that skip cleanly without binaries. |
| envdrift-agent/internal/lockcheck/lockcheck_test.go | Adds unit tests for the self-PID exclusion via fake lister seam (cross-platform), plus a real-lsof integration test that confirms the agent's own open handle is ignored. |
| envdrift-agent/internal/config/config_test.go | New tests cover all duration form variants, explicit-empty-slice clearing vs. absent-key defaults, invalid duration strings, and round-trip Save/Load verification. |
| envdrift-agent/internal/project/config_test.go | Regression tests for pyproject.toml discovery, parent-dir envdrift.toml, precedence ordering, malformed/missing-section skipping, vault mappings, and log output when a project is dropped. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[File change detected\nfsnotify/kqueue] --> B{Idle timeout\nelapsed?}
    B -- No --> A
    B -- Yes --> C[IsFileOpen check\nlsof -t -- path]
    C --> D{Foreign PID\nholding file?}
    D -- Yes\nother process has it open --> A
    D -- No\nonly own PID or none --> E[IsEncrypted check\nparse .env line by line]
    E --> F{Every secret\nvalue is ciphertext?}
    F -- Yes\nfully encrypted --> G[Skip — already\nfully encrypted]
    F -- No\nplaintext or mixed state --> H[envdrift encrypt file\nper-file encrypt path]
    H --> I{Success?}
    I -- Yes --> J[Notify user\nif enabled]
    I -- No --> K[Log error\nretry next cycle]

    subgraph Config Discovery
        L[projectPath] --> M{envdrift.toml\nin dir?}
        M -- Yes --> N[Use it]
        M -- No --> O{pyproject.toml with\ntool.envdrift?}
        O -- Yes --> P[Use it]
        O -- No/malformed --> Q[Walk up to\nparent dir]
        Q --> M
        Q -- reached root --> R[DefaultGuardianConfig\nenabled=false]
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[File change detected\nfsnotify/kqueue] --> B{Idle timeout\nelapsed?}
    B -- No --> A
    B -- Yes --> C[IsFileOpen check\nlsof -t -- path]
    C --> D{Foreign PID\nholding file?}
    D -- Yes\nother process has it open --> A
    D -- No\nonly own PID or none --> E[IsEncrypted check\nparse .env line by line]
    E --> F{Every secret\nvalue is ciphertext?}
    F -- Yes\nfully encrypted --> G[Skip — already\nfully encrypted]
    F -- No\nplaintext or mixed state --> H[envdrift encrypt file\nper-file encrypt path]
    H --> I{Success?}
    I -- Yes --> J[Notify user\nif enabled]
    I -- No --> K[Log error\nretry next cycle]

    subgraph Config Discovery
        L[projectPath] --> M{envdrift.toml\nin dir?}
        M -- Yes --> N[Use it]
        M -- No --> O{pyproject.toml with\ntool.envdrift?}
        O -- Yes --> P[Use it]
        O -- No/malformed --> Q[Walk up to\nparent dir]
        Q --> M
        Q -- reached root --> R[DefaultGuardianConfig\nenabled=false]
    end
```

</a>

<sub>Reviews (13): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/481-agent-e..."](https://github.com/jainal09/envdrift/commit/14eb0793941b58b49722b1df0c50399ade08cfff) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37198261)</sub>


envdrift-agent/internal/config/config.go
```diff
@@ -30,6 +33,46 @@ type DirectoriesConfig struct {
...
+	Directories rawDirectoriesConfig `toml:"directories"`
+}
+
+type rawGuardianConfig struct {
+	Enabled     *bool    `toml:"enabled"`
+	IdleTimeout any      `toml:"idle_timeout"`
+	Patterns    []string `toml:"patterns"`
+	Exclude     []string `toml:"exclude"`
+	Notify      *bool    `toml:"notify"`
+}
```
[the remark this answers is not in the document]
jainal09: Fixed in 6fce261 — rawGuardianConfig.Patterns/Exclude (and rawDirectoriesConfig.Watch) are now *[]string, so a nil pointer = absent key (keep default) and a non-nil pointer to an empty slice = explicit clear.

envdrift-agent/internal/config/config.go
```diff
@@ -77,15 +120,76 @@ func Load() (*Config, error) {
...
+			return nil, fmt.Errorf("%s: guardian.idle_timeout: %w", configPath, err)
+		}
+		cfg.Guardian.IdleTimeout = d
+	}
+	if len(raw.Guardian.Patterns) > 0 {
+		cfg.Guardian.Patterns = raw.Guardian.Patterns
+	}
+	if len(raw.Guardian.Exclude) > 0 {
+		cfg.Guardian.Exclude = raw.Guardian.Exclude
+	}
```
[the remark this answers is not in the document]
jainal09: Fixed in 6fce261 — Load now checks raw.*.Patterns != nil (pointer presence) instead of len() > 0, so an explicit patterns = [] / exclude = [] / watch = [] clears the default. Regression: TestLoadExplicitEmptySlicesClearDefaults (proven failing pre-fix) + TestLoadAbsentSlicesKeepDefaults guards the other half.

envdrift-agent/internal/encrypt/encrypt.go
```diff
@@ -8,13 +8,60 @@ import (
...
+// DOTENV_PUBLIC_KEY_<ENV>). The public key is stored plaintext by design and is
+// not a secret, so it must never count as an unencrypted value (mirrors the
+// CLI's rules in src/envdrift/core/partial_encryption.py).
+const dotenvxPublicKeyPrefix = "DOTENV_PUBLIC_KEY"
+
+// Ciphertext prefixes of an assigned dotenv VALUE: dotenvx writes
+// `KEY="encrypted:..."`, SOPS writes `KEY=ENC[AES256_GCM,...]`.
+const (
+	dotenvxCiphertextPrefix = "encrypted:"
+	sopsCiphertextPrefix    = "ENC[AES256_GCM,"
```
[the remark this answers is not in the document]
jainal09: Fixed in 6fce261 — sopsCiphertextPrefix broadened from ENC[AES256_GCM, to the generic ENC[ tag, so AES256_SIV/CTR (and future suites) are recognized as ciphertext and the guardian no longer re-encrypts them every cycle. Regression: TestIsEncryptedSOPS now covers SIV and CTR (proven failing pre-fix).

envdrift-agent/internal/project/config.go
```diff
@@ -61,25 +61,95 @@ type vaultSyncMappingToml struct {
...
+		case err == nil:
+			var cfg envdriftConfig
+			if err := toml.Unmarshal(data, &cfg); err != nil {
+				return nil, false, err
+			}
+			return &cfg, true, nil
+		case !os.IsNotExist(err):
+			// An existing-but-unreadable envdrift.toml is an error, not a
+			// silent skip (pre-existing behavior for the project's own file).
+			return nil, false, err
```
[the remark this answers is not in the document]
jainal09: Fixed in 6fce261 — LoadAllProjectConfigs now logs 'Skipping project <path>: cannot load config: <err>' instead of a bare continue, so an unreadable ancestor envdrift.toml taking down a child project is visible in the agent log. Regression: TestLoadAllProjectConfigs_LogsDroppedProjectOnLoadError (captures the log buffer; proven failing pre-fix).

envdrift-agent/internal/config/config.go
```diff
@@ -77,15 +120,76 @@ func Load() (*Config, error) {
 	}
 
 	cfg := DefaultConfig()
-	if err := toml.Unmarshal(data, cfg); err != nil {
+	var raw rawConfig
+	if err := toml.Unmarshal(data, &raw); err != nil {
```
jainal09, line 129: Fixed in 6fce261 — extracted mergeGuardian and mergeDirectories from Load; its cyclomatic complexity is back under the threshold.

jainal09, line 129: Fixed in 6fce261 — the same mergeGuardian/mergeDirectories extraction removes the nested conditional blocks Load carried.

envdrift-agent/internal/config/config_test.go
```diff
@@ -142,6 +145,153 @@ func TestSaveAndLoad(t *testing.T) {
...
+	writeGuardianToml(t, "[guardian]\nidle_timeout = \"2d\"\n")
+
+	cfg, err := Load()
+	if err != nil {
+		t.Fatalf("Load() error: %v", err)
+	}
+	if cfg.Guardian.IdleTimeout != 48*time.Hour {
+		t.Errorf("IdleTimeout = %v, want 48h", cfg.Guardian.IdleTimeout)
+	}
+}
```
jainal09: Fixed in 6fce261 — TestLoadExtendedDurationUnits and TestLoadLegacyNanosecondInteger merged into one table-driven TestLoadIdleTimeoutForms.

envdrift-agent/internal/encrypt/encrypt.go
```diff
@@ -26,30 +73,73 @@ func IsEncrypted(path string) (encrypted bool, err error) {
...
 	}
+	return sopsMetadataGroupKey.MatchString(key)
+}
 
-	return false, scanner.Err()
+// unquoteValue strips surrounding whitespace and a single layer of matching
+// quotes, the two styles dotenv/SOPS emit (`"..."` and `'...'`).
+func unquoteValue(value string) string {
+	v := strings.TrimSpace(value)
+	if len(v) >= 2 && (v[0] == '"' || v[0] == '\'') && v[len(v)-1] == v[0] {
```
jainal09: Fixed in 6fce261 — extracted isWrappedInMatchingQuotes from unquoteValue, removing the 3-operand conditional.

envdrift-agent/internal/encrypt/encrypt_test.go
```diff
@@ -189,3 +250,105 @@ func TestBuildEncryptCommand_BinaryNoModuleFlag(t *testing.T) {
...
+	}
+
+	encrypted, err := IsEncrypted(envPath)
+	if err != nil {
+		t.Fatalf("IsEncrypted: %v", err)
+	}
+	if !encrypted {
+		t.Errorf("IsEncrypted = false after a successful real encrypt; content:\n%s", content)
+	}
+}
```
jainal09, line 417: Fixed in 6fce261 — extracted captureEncryptOutput (the failure-diagnostics retry) out of TestEncryptEndToEnd_RealCLI, dropping its complexity below the threshold.

envdrift-agent/internal/encrypt/encrypt_test.go
```diff
@@ -64,6 +66,64 @@ func TestIsEncrypted(t *testing.T) {
...
+			expected: true,
+		},
+		{
+			// A user variable that merely starts with sops_ is a real secret,
+			// not SOPS bookkeeping: it must keep the file in mixed state.
+			name: "sops_-prefixed user variable stays a plaintext secret",
+			content: "DATABASE_URL=\"ENC[AES256_GCM,data:abc,type:str]\"\n" +
+				"sops_token=AKIA-very-plaintext\n",
+			expected: false,
+		},
```
jainal09, line 201: Fixed in 6fce261 — TestIsEncrypted split into TestIsEncrypted (dotenvx/general) and TestIsEncryptedSOPS via a shared runIsEncryptedCases harness; each table is now well under the size threshold.

envdrift-agent/internal/project/config_test.go
```diff
@@ -211,6 +211,209 @@ env_file = "service.env"
...
+`)
+
+	cfg, err := LoadProjectConfig(project)
+	if err != nil {
+		t.Fatalf("LoadProjectConfig() error = %v", err)
+	}
+	if !cfg.Enabled {
+		t.Error("Enabled = false: walk stopped at a pyproject.toml without [tool.envdrift]")
+	}
+}
```
jainal09: Fixed in 6fce261 — the two pyproject-skipped tests merged into one table-driven TestLoadProjectConfig_PyprojectSkippedAndWalkContinues.

envdrift-agent/internal/project/config_test.go
```diff
@@ -250,3 +441,42 @@ enabled = false
...
+		t.Fatalf("LoadAllProjectConfigs() error = %v", err)
+	}
+	if len(configs) != 0 {
+		t.Errorf("expected the unreadable project to be dropped, got %d configs", len(configs))
+	}
+	out := logbuf.String()
+	if !strings.Contains(out, "Skipping project") || !strings.Contains(out, project) {
+		t.Errorf("dropped project was not logged; log output:\n%s", out)
+	}
+}
```
jainal09, line 496: Addressed in 946a0b5 — extracted projectUnderUnreadableAncestor and captureLog helpers from the test, dropping its method complexity below the threshold.

envdrift-agent/internal/config/config_test.go
```diff
@@ -5,8 +5,11 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
```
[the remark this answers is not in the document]
jainal09: Resolving as accepted-advisory: this is a module-level aggregate (Overall Code Complexity) on a TEST file that grew because this PR adds real regression coverage (empty-slice, SOPS SIV/CTR, dropped-project logging). The per-method flags were each fixed; further reducing the aggregate would mean splitting the test file purely to satisfy a metric, which would hurt readability. Consistent with treating CodeScene as advisory.

jainal09:
@coderabbitai full review

envdrift-agent/internal/encrypt/encrypt.go
```diff
@@ -26,30 +78,83 @@ func IsEncrypted(path string) (encrypted bool, err error) {
...
+	return sopsMetadataGroupKey.MatchString(key)
+}
+
+// unquoteValue strips surrounding whitespace and a single layer of matching
+// quotes, the two styles dotenv/SOPS emit (`"..."` and `'...'`).
+func unquoteValue(value string) string {
+	v := strings.TrimSpace(value)
+	if isWrappedInMatchingQuotes(v) {
+		v = strings.TrimSpace(v[1 : len(v)-1])
+	}
```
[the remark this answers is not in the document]
jainal09: Confirmed valid — fixed in 90c9eb5. `KEY="encrypted:x" # note` failed the matching-quotes check (first char `"`, last char of the comment), so the quoted ciphertext was classified plaintext and the guardian re-ran encrypt every idle cycle.

`unquoteValue` now routes through `stripCommentAfterQuotedValue`, which drops the trailing token only when nothing but a `# comment` follows the closing quote; any other trailing token leaves the value untouched, so a malformed line still counts as plaintext — the safe direction (worst case is a redundant, idempotent re-encrypt), unlike the broader suggestion which would classify `KEY="encrypted:x" MORE_JUNK` as ciphertext.

Regression tests (all proven failing pre-fix): quoted dotenvx and SOPS ciphertext with inline comments, quoted plaintext with a comment, a non-comment trailing token, and an empty quoted assignment with a comment.

The Python CLI's `_unquote_value` (src/envdrift/core/partial_encryption.py) has the identical limitation; filed as #578 to keep this PR agent-themed.

envdrift-agent/internal/encrypt/encrypt.go
```diff
@@ -26,30 +78,108 @@ func IsEncrypted(path string) (encrypted bool, err error) {
...
+// stripCommentAfterQuotedValue drops an inline `# comment` that follows a
+// closed quoted token: dotenv allows `KEY="value" # note`, and the trailing
+// comment defeated the matching-quotes check, so quoted ciphertext with an
+// inline comment was misclassified as plaintext and the guardian re-ran
+// encrypt on the already-encrypted file every idle cycle. Only a comment (or
+// nothing) may follow the closing quote — any other trailing token leaves the
+// value untouched, so a malformed line still counts as plaintext (the safe
+// direction: the worst case is a redundant, idempotent re-encrypt).
+func stripCommentAfterQuotedValue(v string) string {
+	if len(v) < 2 || (v[0] != '"' && v[0] != '\'') {
```
jainal09: Fixed in 8e528de: extracted `isQuoteByte` (also reused by `isWrappedInMatchingQuotes`), so the guard is now `len(v) < 2 || !isQuoteByte(v[0])` — one logical operator per conditional, below the threshold. No behavior change; the #504 inline-comment regression tests still pass.

envdrift-agent/internal/encrypt/encrypt_test.go
```diff
@@ -3,18 +3,47 @@ package encrypt
...
+	}
+}
+
 func TestIsEncrypted(t *testing.T) {
-	tests := []struct {
-		name     string
-		content  string
-		expected bool
-	}{
+	runIsEncryptedCases(t, []isEncryptedCase{
```
jainal09, line 46: Fixed in 8e528de: the five inline-comment regression cases (dotenvx + SOPS) moved into a focused `TestIsEncryptedInlineComments` on the shared `runIsEncryptedCases` runner, bringing `TestIsEncrypted` from 97 to 77 active lines — under the 80-line threshold.

docs/agent-spec.md
```diff
@@ -359,7 +362,7 @@ Instead of watching entire directories, the agent:
 | Package | File | Purpose |
 |---------|------|---------|
 | `registry` | `internal/registry/registry.go` | Loads and watches `~/.envdrift/projects.json` |
-| `project` | `internal/project/config.go` | Loads per-project `[guardian]` settings from `envdrift.toml` |
+| `project` | `internal/project/config.go` | Loads per-project `[guardian]` settings from `envdrift.toml` or `pyproject.toml` (CLI `find_config` contract) |
```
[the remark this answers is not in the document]
jainal09: Fixed — dropped the redundant `per-project` from the row (the config.go path + `find_config` contract already convey the scope); it's now 144 chars and `make lint-docs` passes clean.

Issue #22, opened 2026-01-01
Refactor: Extract duplicated config loading logic into shared helper


jainal09:
## Summary

The `sync` and `pull` commands in `cli.py` share approximately 125 lines of duplicated config loading and vault setup logic (lines 818-965 and 1115-1241). This creates maintenance burden and risk of inconsistency when one command is updated but the other is not.

## Suggested Solution

Extract the common logic into a shared helper function:

```python
def _load_config_and_create_vault_client(
    config_file: Path | None,
    provider: str | None,
    vault_url: str | None,
    region: str | None,
) -> tuple[SyncConfig, Any, str, str | None, str | None]:
    """
    Load configuration and create vault client.
    
    Returns:
        tuple of (sync_config, vault_client, effective_provider, effective_vault_url, effective_region)
    """
    # ... extracted config loading logic ...
    # ... extracted validation logic ...
    # ... extracted vault client creation ...
    return sync_config, vault_client, effective_provider, effective_vault_url, effective_region
```

Then both `sync` and `pull` commands can call this helper, reducing duplication and ensuring consistency.

## Affected Files

- `src/envdrift/cli.py`

## Related

- PR #21 - Added `pull` command with profile filtering
- CodeRabbit review suggestion: https://github.com/jainal09/envdrift/pull/21#discussion_r2656237070

## Labels

enhancement, tech-debt

PR #505, opened 2026-06-12
fix(scanner): close native detection gaps (encodings, env-file shapes, over-broad ignores) (#477)

### LINKED ISSUE #477: guard: native scanner detection gaps — UTF-16, .env.local/.env.test, trailing-.env names, over-broad ignores
## Summary

The native scanner's file-classification heuristics (default ignore lists, the `_is_env_file` naming gate, the
binary-detection ratio, and ignore-pattern path relativization) each silently exclude whole classes of real-secret
files from scanning — or apply inconsistently depending on how the scan path is spelled. Since native is the
always-on offline baseline (gitleaks is auto-install-only and frequently absent), these gaps produce clean exit-0
passes on files containing live AWS keys and GitHub PATs. Every item below was reproduced twice (original
reviewer + independent verifier) on origin/main commit cb68c97.

## Findings

- [ ] **Tracked `.env.local` / `.env.test` are hard-ignored, so real secrets in them pass guard with exit 0.**
  `src/envdrift/scanner/native.py:62`
  **Repro:** commit a `.env.local` containing AWS key+secret and a `ghp_` PAT;
  `envdrift guard --native-only --no-auto-install --json .` → `findings: []`, exit 0
  **Expected:** a tracked plaintext `.env.local`/`.env.test` with real secrets is flagged (unencrypted-env-file
  HIGH + pattern hits) — or at minimum the exclusion is documented.
  **Actual:** the whole file is skipped before any check; identical content in `.env.prod` yields 5 findings, exit 1.
  **Verifier note:** `.env.local` is the canonical real-secrets file for Next.js/Vite/CRA; docs/cli/lock.md:301
  even shows lock ENCRYPTING `.env.local`, so guard's skip is internally inconsistent. Distinct from #441
  (untracked non-.env names).

- [ ] **UTF-16-encoded env files are classified as binary and skipped wholesale — "No secrets detected" / exit 0
  on a file with real AWS credentials.** `src/envdrift/scanner/native.py:627`
  **Repro:** `printf 'AWS_ACCESS_KEY_ID=AKIA...\n...' > ctrl.env; iconv -f UTF-8 -t UTF-16 ctrl.env > u16.env;`
  `envdrift guard u16.env --native-only; echo $?` (ctrl.env: 2 critical, exit 1; u16.env: clean, exit 0)
  **Expected:** detect the BOM (or 2-byte stride) and decode UTF-16 text before pattern-matching, flagging the
  file exactly like its UTF-8 equivalent.
  **Actual:** UTF-16's ~50% NUL bytes trip `_looks_binary` (>5% non-text threshold); the file is skipped entirely,
  both with and without BOM.
  **Verifier note:** UTF-16 is Windows Notepad/PowerShell `Out-File` default "Unicode" output, and the project
  ships Windows CI (#442/#451). Distinct from #443's single-NUL item (fixed in 9780d24).

- [ ] **Trailing-.env names (`production.env`, `database.env`, `app.env`) are not recognized as env files — no
  unencrypted-env-file policy and no entropy scan.** `src/envdrift/scanner/native.py:724`
  **Repro:** commit `production.env` with `DB_CONN_VALUE=<32 random base62>`;
  `envdrift guard --native-only --no-auto-install --json .` → 0 findings, exit 0
  **Expected:** `_is_env_file` also recognizes the `<name>.env` convention (used by `docker --env-file`, direnv,
  CI systems) so it gets the same policy + entropy coverage as `.env.production`.
  **Actual:** the file is collected and pattern-scanned, but the env-file policy (native.py:673) and entropy gate
  (native.py:719) are both keyed on `_is_env_file`; a prefix-less high-entropy secret passes clean.
  **Verifier note:** renaming the identical file to `.env.production` yields unencrypted-env-file HIGH +
  high-entropy-string MEDIUM, exit 2.

- [ ] **`DEFAULT_GLOBAL_IGNORE_PATHS` unconditionally suppresses even CRITICAL distinctive-prefix secrets
  (ghp_/AKIA/PyPI) committed in `pyproject.toml` / lock files — across ALL scanners.** `src/envdrift/scanner/engine.py:230`
  **Repro:** commit a `ghp_` PAT + AKIA key inside `pyproject.toml` and `package-lock.json`, the same `ghp_` in
  `creds.config`; `envdrift guard --native-only --no-auto-install --json .`
  **Expected:** the global ignore suppresses only noisy entropy/keyword findings in these files (or is scoped to
  rule ids), not zero-false-positive CRITICAL tokens.
  **Actual:** the pyproject/lock findings are silently dropped post-scan for every scanner; only `creds.config`
  is flagged. A real committed token in pyproject.toml (e.g. private-index URL token) can never surface.
  **Verifier note:** the filter is applied post-merge at engine.py:535, so it also strips gitleaks/trufflehog
  findings.

- [ ] **Directory-scoped ignore patterns (`bin/**`, `dist/**`, `vendor/**`, `node_modules/**`, ...) are silently
  not applied when guard is invoked with a relative or symlinked path — `guard` and `guard .` give different
  results.** `src/envdrift/scanner/native.py:577`
  **Repro:** secrets in `bin/`, `env/`, `config/`; compare `envdrift guard --native-only --json` (1 finding)
  vs `envdrift guard --native-only --json .` (3 findings, bin/ and env/ scanned)
  **Expected:** path-scoped ignores apply consistently regardless of path spelling; `base_path` should be
  `resolve()`d before relativizing, matching `_collect_files`.
  **Actual:** `file_path.relative_to(base_path)` raises `ValueError` for relative/symlinked bases and falls back
  to the absolute path, so directory patterns never match — guard scans dirs it is configured to ignore.
  **Verifier note:** docs recommend the relative form (`guard ./apps ./services`, docs/cli/guard.md:316,361);
  macOS `/tmp` symlink alone triggers it. Scans more rather than less, but results are inconsistent and noisy.

## Suggested fix direction

- Drop `.env.local`/`.env.test` from `DEFAULT_IGNORE_PATTERNS` (native.py:62-69), or scan them when tracked and
  document any remaining exclusion in docs/cli/guard.md.
- In `_looks_binary`/`_scan_file` (native.py:256-274, 617-628), decode through UTF-16/UTF-8 BOMs and the 2-byte
  NUL stride before applying the binary ratio.
- Extend `_is_env_file` (native.py:724-734) to match the trailing `<name>.env` convention.
- Scope `DEFAULT_GLOBAL_IGNORE_PATHS` (engine.py:230-239) to noisy rule ids (entropy/generic-keyword) instead of
  dropping all findings for those filenames.
- In `_should_ignore` (native.py:577-581), resolve `base_path` before `relative_to` so it matches the resolved
  file paths from `_collect_files`.

_Part of the 2026-06 final deep-review (batch: GUARD-2). Each item was independently reproduced and re-verified before filing._

jainal09:
## Summary

The native scanner (guard's always-on offline baseline) silently excluded whole classes of
real-secret files, producing clean exit-0 passes on files containing live AWS keys and GitHub
PATs. This PR closes all five gaps from #477 with end-to-end regression tests that drive the
real CLI as a subprocess against scratch git repositories.

## Fixes

- **Tracked `.env.local` / `.env.test` scanned** — both names are dropped from
  `DEFAULT_IGNORE_PATTERNS` (`src/envdrift/scanner/native.py`); only the
  `.env.example`/`.env.sample`/`.env.template` templates stay excluded. The Go test-binary glob is
  narrowed to `[!.]*.test` so it cannot swallow `.env.test`.
- **UTF-16/UTF-32/UTF-8-BOM env files decoded, not skipped as binary** — new
  `_decode_unicode_text` decodes through BOMs (longest first, so UTF-32's `FF FE 00 00` is not
  misread as UTF-16) and a BOM-less UTF-16 NUL-stride sniff (`_sniff_utf16_stride`) before the
  binary-ratio heuristic can discard the file. Strict-decode failures fall back to the existing
  binary check + lenient UTF-8 path.
- **Trailing-`.env` names are env files** — `_is_env_file` (module + method, single source of
  truth) now matches `<name>.env` (`production.env`, `database.env`), so the
  unencrypted-env-file policy and entropy gate apply; the untracked-file git pathspecs gained
  `:(glob)**/*.env` to match.
- **Default global ignores can no longer swallow high-confidence findings** —
  `DEFAULT_GLOBAL_IGNORE_PATHS` is now scoped to noisy keyword/entropy rule ids (new
  `IgnoreConfig.noisy_rule_paths` + `is_noisy_rule`: ids containing `generic`/`entropy`/`keyword`,
  covering all scanner namespaces). A `ghp_`/`AKIA`/PyPI token in `pyproject.toml` or a lock file
  now surfaces from every scanner, while the noisy integrity-hash/keyword FPs stay suppressed.
  The native scanner's duplicate wholesale skips of those files are removed. Explicit user
  `ignore_paths` keep full-suppression semantics.
- **Directory-scoped ignores apply regardless of path spelling** — `_should_ignore` resolves both
  the file and the base before `relative_to`, so `guard`, `guard .`, and symlinked paths (macOS
  `/tmp`) produce the same result set and `bin/**`/`dist/**`/`vendor/**` ignores always apply.

Docs updated in the same PR: `docs/cli/guard.md` + `docs/guides/guard.md` (env-file shapes,
encoding handling, built-in noisy-only default ignores, ignore priority).

New `tests/integration/test_guard_native_gaps_e2e.py` (8 tests, real CLI subprocess + scratch git
repos, fixtures built only by string concatenation, UTF-16 fixtures written with explicit
`encoding="utf-16"` / `utf-16-le` / `utf-16-be`).

On unfixed `origin/main` (e93c954), all 8 fail:

```text
FAILED ...::test_tracked_env_local_and_test_with_secrets_are_flagged[.env.local]
FAILED ...::test_tracked_env_local_and_test_with_secrets_are_flagged[.env.test]
FAILED ...::test_utf16_env_file_flagged_like_utf8[utf-16]
FAILED ...::test_utf16_env_file_flagged_like_utf8[utf-16-le]
FAILED ...::test_utf16_env_file_flagged_like_utf8[utf-16-be]
FAILED ...::test_trailing_env_name_gets_policy_and_entropy_coverage
FAILED ...::test_distinctive_secret_in_pyproject_and_lockfile_not_suppressed
FAILED ...::test_relative_path_invocation_applies_directory_ignores
8 failed in 2.19s
```

With the fix: `8 passed in 2.52s`.

Additional verification:

- Unit lane: `uv run pytest -m "not integration"` → **2566 passed** (the two
  `test_guard_machine_output.py` fixtures named `clean.env`/`changed.env` were updated to
  non-env names, since a plaintext `*.env` is now correctly a HIGH finding).
- Focused unit tests added for every new helper/branch: `_decode_unicode_text`,
  `_sniff_utf16_stride` (both endians, odd-length, balanced-NUL rejection), env-file shapes,
  template exclusions, `[!.]*.test`, relative/symlinked `_should_ignore`, `is_noisy_rule`,
  noisy-path filtering, engine-level scoping, user-`ignore_paths` full suppression.
- Coverage on touched files: native.py 91%, engine.py 83%, ignores.py 97%.
- Integration subset: guard CLI e2e, guard setup, scanner git tools, external binaries (real
  talisman/trivy/infisical), partial-encryption e2e, git hooks — all green.
- Gates: ruff check, ruff format, pyrefly, bandit, `make lint-docs` — all green.

Fixes #477

<a href="https://cubic.dev/pr/jainal09/envdrift/pull/505?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Support for trailing-dot env files (e.g., `production.env`) and additional naming patterns
  * UTF-16 and UTF-32 encoded env file detection
  * Enhanced detection in config and lockfiles with refined ignore scoping

* **Bug Fixes**
  * Fixed directory-scoped ignore patterns to work correctly with relative paths and symlinks
  * Improved handling of `.env.local` and `.env.test` file detection
  * Enhanced BOM and text encoding detection

* **Documentation**
  * Expanded guard documentation with details on env-file patterns, encoding behavior, and built-in ignore defaults

<h3>Greptile Summary</h3>

This PR closes five native-scanner detection gaps from #477 (false-clean exits on `.env.local`/`.env.test`, UTF-16 env files skipped as binary, trailing `<name>.env` not recognized, `DEFAULT_GLOBAL_IGNORE_PATHS` unconditionally suppressing CRITICAL tokens, and directory-scoped ignores failing for relative/symlinked paths), backed by eight end-to-end regression tests against real CLI subprocesses.

- **`native.py`**: removes `.env.local`/`.env.test`/lock-file/config-file entries from `DEFAULT_IGNORE_PATTERNS`; adds `_decode_unicode_text` + `_sniff_utf16_stride` for BOM/BOM-less UTF-16 support; expands `_is_env_file` to include trailing `<name>.env` convention; refactors `_should_ignore` to resolve both sides before `relative_to` and splits pattern matching into `_pattern_matches` so basename-shaped patterns (e.g. `[!.]*.test`) are never applied to the full relative path.
- **`ignores.py` / `engine.py`**: introduces `noisy_rule_paths` on `IgnoreConfig` and `is_noisy_rule()`, wiring the engine's `DEFAULT_GLOBAL_IGNORE_PATHS` through the new field so only keyword/entropy rule IDs are suppressed in config/lock files while distinctive-prefix tokens (GitHub PAT, AWS key) always surface.
- **Tests**: 235 new unit-test lines across `test_native.py`/`test_ignores.py`/`test_engine.py` and a new 318-line integration file exercising all five gaps via real CLI subprocess.

Safe to merge; all five detection gaps are closed with targeted, well-tested changes and no regressions to existing suppression semantics.

The changes are focused and logically correct: BOM/stride UTF-16 decoding with errors=replace handles all realistic file corruption scenarios, _pattern_matches correctly scopes basename patterns so [!.]*.test can never swallow a nested .env.test, noisy-rule scoping preserves full suppression for user-configured paths while letting distinctive-prefix tokens surface from config/lock files, and path resolution in _should_ignore is now consistent regardless of how the scan path is spelled. All five gaps have fail-then-pass regression tests driving the real CLI subprocess. The only note is a minor fragility in the test helper _guard_json when progress output contains brace characters, which does not affect production code.

No files require special attention for correctness. The integration test helper _guard_json in tests/integration/test_guard_native_gaps_e2e.py has a minor robustness gap worth addressing before the test suite grows further.

| Filename | Overview |
|----------|----------|
| src/envdrift/scanner/native.py | Core scanner fixes: BOM/stride UTF-16 decoding, trailing-.env recognition, path-resolution fix in _should_ignore, and basename-vs-full-path split in _pattern_matches; logic is sound and well-covered by new unit tests. |
| src/envdrift/scanner/ignores.py | Adds noisy_rule_paths field and is_noisy_rule(), refactors _matches_global_ignore into _matches_path_patterns for reuse; from_dict now correctly populates the new field. |
| src/envdrift/scanner/engine.py | Wires DEFAULT_GLOBAL_IGNORE_PATHS through noisy_rule_paths instead of ignore_paths; user ignore_paths retain full-suppression semantics while built-in defaults are now scoped to noisy rules only. |
| tests/integration/test_guard_native_gaps_e2e.py | New integration tests covering all five gaps; _guard_json helper uses index/rindex to extract JSON from potentially mixed stdout, which is fragile if progress text contains braces. |
| tests/scanner/test_native.py | 235 new unit-test lines covering _decode_unicode_text, _sniff_utf16_stride, env-file naming shapes, _should_ignore path resolution, and _pattern_matches nested-.env.test fix. |
| tests/scanner/test_ignores.py | Tests for noisy-rule classification, noisy-path suppression, from_dict round-trip, and user ignore_paths full-suppression semantics — comprehensive coverage of new IgnoreConfig fields. |
| tests/scanner/test_engine.py | Adds TestDefaultGlobalIgnoreScoping verifying distinctive-prefix tokens surface from pyproject.toml/lock files while noisy keyword findings are suppressed. |
| tests/unit/test_guard_machine_output.py | Fixture files renamed from *.env to *.cfg to avoid triggering the new unencrypted-env-file finding; change is correctly motivated and low-risk. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[File path] --> B{_should_ignore?}
    B -- yes --> SKIP[Skip file]
    B -- no --> C[_read_scannable_content]
    C --> D{_decode_unicode_text\nBOM / stride sniff}
    D -- BOM match → decode with errors=replace --> E[text content]
    D -- BOM-less UTF-16 stride → decode with errors=replace --> E
    D -- None --> F{_looks_binary?}
    F -- yes --> SKIP2[Skip binary]
    F -- no --> G[UTF-8 decode with errors=ignore]
    G --> E
    E --> H[Strip NULs, check empty]
    H --> I{_is_env_file?}
    I -- yes --> J{Encrypted?}
    J -- no --> K[unencrypted-env-file finding HIGH]
    I -- no --> L[Pattern scan for secrets]
    K --> L
    L --> M[Raw findings list]
    M --> N[IgnoreFilter.filter]
    N --> O{inline ignore?}
    O -- yes --> DROP[Drop finding]
    O -- no --> P{rule+path ignore?}
    P -- yes --> DROP
    P -- no --> Q{is_noisy_rule AND\nmatches noisy_rule_paths?}
    Q -- yes --> DROP
    Q -- no --> R{matches ignore_paths?}
    R -- yes --> DROP
    R -- no --> KEEP[Keep finding]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[File path] --> B{_should_ignore?}
    B -- yes --> SKIP[Skip file]
    B -- no --> C[_read_scannable_content]
    C --> D{_decode_unicode_text\nBOM / stride sniff}
    D -- BOM match → decode with errors=replace --> E[text content]
    D -- BOM-less UTF-16 stride → decode with errors=replace --> E
    D -- None --> F{_looks_binary?}
    F -- yes --> SKIP2[Skip binary]
    F -- no --> G[UTF-8 decode with errors=ignore]
    G --> E
    E --> H[Strip NULs, check empty]
    H --> I{_is_env_file?}
    I -- yes --> J{Encrypted?}
    J -- no --> K[unencrypted-env-file finding HIGH]
    I -- no --> L[Pattern scan for secrets]
    K --> L
    L --> M[Raw findings list]
    M --> N[IgnoreFilter.filter]
    N --> O{inline ignore?}
    O -- yes --> DROP[Drop finding]
    O -- no --> P{rule+path ignore?}
    P -- yes --> DROP
    P -- no --> Q{is_noisy_rule AND\nmatches noisy_rule_paths?}
    Q -- yes --> DROP
    Q -- no --> R{matches ignore_paths?}
    R -- yes --> DROP
    R -- no --> KEEP[Keep finding]
```

</a>

1. `src/envdrift/scanner/ignores.py`, line 87-91 ([link](https://github.com/jainal09/envdrift/blob/a82158785491695197fade180b5110b7a7026788/src/envdrift/scanner/ignores.py#L87-L91)) 

   <a href="#">P2</a> **`from_dict` does not populate the new `noisy_rule_paths` field**

   `IgnoreConfig.from_dict` is a public classmethod used to build an `IgnoreFilter` from a TOML-sourced dict. After this PR, `IgnoreConfig` gains a `noisy_rule_paths` field that controls the built-in suppression of keyword/entropy rules in config and lock files. Because `from_dict` never sets `noisy_rule_paths`, any SDK caller who creates `IgnoreFilter(IgnoreConfig.from_dict(config))` directly — bypassing `ScanEngine` — will get an empty list and miss the built-in defaults entirely, while `ScanEngine.__init__` correctly populates it directly. Even if the intent is that `noisy_rule_paths` is engine-internal, leaving `from_dict` silently incomplete is a footgun for downstream callers.
</details>

<sub>Reviews (5): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/477-native-..."](https://github.com/jainal09/envdrift/commit/c637fadaa90ed06e9b5c2c766c935ad89d393485) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37198509)</sub>


src/envdrift/scanner/native.py
```diff
@@ -148,18 +149,18 @@
     # Go
     "vendor/**",
     "*.exe",
-    "*.test",
+    # Compiled Go test binaries are named "<pkg>.test" — never dotfiles. The
+    # leading [!.] keeps this from swallowing the real-secret ".env.test" (#477).
+    "[!.]*.test",
```
jainal09, line 154: 🤖 **envdrift review-bot** · high · bug


The PR narrows the Go test-binary glob from `*.test` to `[!.]*.test` so it "cannot swallow `.env.test`". That works only for a top-level `.env.test`, because `NativeScanner._should_ignore` matches every pattern against **three** candidates, the first being the full relative path:

```python
for pattern in self._ignore_patterns:
    # Match against full relative path
    if fnmatch.fnmatch(path_str, pattern):
        return True
```

`fnmatch`'s `*` matches across `/`, and the `[!.]` anchor only constrains the FIRST character of the whole string — which for a nested file is the first directory letter, not the filename's dot. Empirically (verified against the helpers extracted from this head):

```
fnmatch('.env.test',          '[!.]*.test')  -> False  (fixed case)
fnmatch('apps/web/.env.test', '[!.]*.test')  -> True   (still ignored!)
fnmatch('config/.env.test',   '[!.]*.test')  -> True   (still ignored!)
```

So a tracked `apps/web/.env.test` containing live AWS keys is skipped before `_scan_file` ever runs — findings: [], exit 0 — which is exactly the issue #477 gap-1 failure mode, in the *most common* layout for the "canonical Next.js/Vite/CRA real-secrets file" the PR's own docs cite (monorepo apps live in subdirectories). It also contradicts the new docs/guides/guard.md claim that the policy applies to "every file whose name matches an env-file shape ... including `.env.local` and `.env.test`".

**Test gap:** every new test places `.env.test` at the repo/tmp root (`tests/integration/test_guard_native_gaps_e2e.py::test_tracked_env_local_and_test_with_secrets_are_flagged` writes `tmp_path / env_name`; `tests/scanner/test_native.py::TestRealSecretEnvNamesScanned` likewise), so no test would catch this.

**Suggested direction:** the dot-exclusion belongs to the basename, not the joined path — e.g. keep plain `*.test` but special-case dotfile basenames in `_should_ignore`, or skip the full-path/parts checks for this pattern, or replace the glob with an explicit basename predicate (`name.endswith('.test') and not name.startswith('.')`). Add nested-path parametrizations (`apps/web/.env.test`) to both the unit and e2e regression tests — they fail on this head.

src/envdrift/scanner/native.py
```diff
@@ -274,6 +276,67 @@ def _looks_binary(raw: bytes, *, sample: int = 8192, threshold: float = 0.05) ->
...
+            try:
+                return raw.decode(encoding)
+            except UnicodeDecodeError:
+                return None
+    stride_encoding = _sniff_utf16_stride(raw)
+    if stride_encoding is not None:
+        try:
+            return raw.decode(stride_encoding)
+        except UnicodeDecodeError:
+            return None
```
jainal09: 🤖 **envdrift review-bot** · medium · bug


In `_decode_unicode_text`, the BOM-less stride path strict-decodes the **entire** raw buffer and returns `None` on any `UnicodeDecodeError`:

```python
stride_encoding = _sniff_utf16_stride(raw)
if stride_encoding is not None:
    try:
        return raw.decode(stride_encoding)
    except UnicodeDecodeError:
        return None
```

The caller (`_scan_file`, line 697-705) then runs `_looks_binary(raw)`, which is True for any UTF-16-ish bytes (~50% NULs), so the file is skipped wholesale with zero findings — the exact pre-fix gap-2 behavior.

The internal inconsistency makes this trivially reachable: `_sniff_utf16_stride` *deliberately* tolerates odd-length input (`if len(chunk) % 2: chunk = chunk[:-1]`, and `test_stride_sniff_handles_odd_length_chunk` pins that behavior), but `bytes.decode('utf-16-le')` on the odd-length `raw` raises "truncated data". So the code positively identifies the bytes as UTF-16 text, then discards the file as binary. Empirically verified against this head's helpers:

```
clean BOM-less LE          : sniff=utf-16-le  decode_ok=True
clean + 1 trailing byte    : sniff=utf-16-le  decode=None  _looks_binary=True  -> skipped, exit 0
lone surrogate mid-file    : sniff=utf-16-le  decode=None  _looks_binary=True  -> skipped, exit 0
```

Impact: a truncated/interrupted Windows write, a mixed-encoding concatenation, or an **attacker appending a single byte** to a BOM-less UTF-16 env file silently re-opens the "clean exit-0 over live AWS keys" hole — the same one-byte-evasion class the adjacent comment block says #22 closed for NULs.

Relation to existing review: greptile's inline comment covers the **BOM** path's strict-decode fail-open and explicitly proposes `errors="replace"` for the BOM branch "without removing the strict-decode gate for the non-BOM stride path" — i.e. its suggested fix leaves this stride-path hole open, so it needs a separate decision.

**Suggested direction:** once the stride sniff has positively identified UTF-16 text, don't fail back to the binary skip — decode the even-length prefix (`raw[:len(raw)//2*2]`) and/or use `errors="replace"`/`surrogatepass` (the stride sniff's dual-plane thresholds already make false positives on genuine binaries effectively impossible). At minimum, a decode failure after a positive sniff should bypass `_looks_binary` and take the lenient path rather than skipping the file silently. Add regression tests for the trailing-stray-byte and lone-surrogate cases — both skip the file on this head.

docs/guides/guard.md
```diff
@@ -4,11 +4,29 @@ Use `envdrift guard` as a last line of defense against plaintext secrets.
...
+  canonical real-secrets files for Next.js/Vite/CRA local development
+- `<name>.env` (e.g. `production.env`, `database.env`) — the `docker --env-file`,
+  direnv, and CI convention
+- Custom env files declared in `vault.sync` mappings
+
+Only the template names `.env.example`, `.env.sample`, and `.env.template` are skipped
+by default, since they hold placeholder values by convention.
+
+Env files are scanned regardless of text encoding: UTF-8, UTF-16 (the Windows Notepad
+and PowerShell "Unicode" default), and UTF-32 — with or without a BOM — are decoded
```
jainal09: 🤖 **envdrift review-bot** · medium · doc-mismatch


The new encoding paragraph in docs/guides/guard.md states:

> Env files are scanned regardless of text encoding: UTF-8, UTF-16 (the Windows Notepad and PowerShell "Unicode" default), and UTF-32 — **with or without a BOM** — are decoded before pattern matching instead of being misclassified as binary and skipped.

The code does not implement BOM-less UTF-32: `_decode_unicode_text` handles UTF-32 only via `_BOM_CODECS` (`codecs.BOM_UTF32_LE/BE`), and the only BOM-less sniff is `_sniff_utf16_stride`, whose thresholds reject the UTF-32 byte layout. ASCII UTF-32-LE is `X 00 00 00 ...`: the odd plane is 100% NUL but the even plane is ~50% NUL, failing both `even_nuls/half < 0.05` (LE branch) and `even_nuls/half > 0.7` (BE branch). Empirically verified against this head's helpers:

```
BOM-less UTF-32-LE: sniff=None  decode=None  _looks_binary=True  -> file skipped, exit 0
BOM-less UTF-32-BE: sniff=None  decode=None  _looks_binary=True  -> file skipped, exit 0
```

So a `iconv -t UTF-32LE` (or Python `encoding="utf-32-le"`) env file full of live credentials passes guard clean while the doc tells the user it is covered — a misleading security-coverage claim in a guard doc. Note the unit test only exercises `encoding="utf-32"`, which Python writes WITH a BOM, so the test suite never touches the BOM-less case. The sibling doc docs/cli/guard.md line 26 is worded correctly ("decoded through UTF-8/UTF-16/UTF-32 BOMs (and BOM-less UTF-16)").

**Suggested direction:** either fix the doc to match docs/cli/guard.md's accurate wording (BOM-less detection is UTF-16 only), or extend the sniff with a UTF-32 stride check (3-of-4 NUL planes) and add `utf-32-le`/`utf-32-be` parametrizations to `TestUnicodeEncodedFiles` (they fail on this head). The blanket opener "scanned regardless of text encoding" should be softened either way.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit a821587


The PR's intent is sound but both #477 gaps are only partially closed: the `.env.test` un-ignore works solely at the repo root, the BOM-less UTF-16 path fails open on a single stray byte, and one guard doc overclaims UTF-32 coverage.

**Findings (3):**
- [high] Nested .env.test files are still hard-ignored: '[!.]*.test' is fnmatched against the full relative path, so gap 1 is only fixed at the repo root — `src/envdrift/scanner/native.py:154` (inline)
- [medium] BOM-less UTF-16 stride path fails OPEN on strict-decode failure: the sniff tolerates odd-length input that the decode rejects, so one appended byte makes guard skip the whole file — `src/envdrift/scanner/native.py:336` (inline)
- [medium] docs/guides/guard.md claims BOM-less UTF-32 files are decoded — the code only sniffs BOM-less UTF-16, so a BOM-less UTF-32 env file is still skipped as binary — `docs/guides/guard.md:27` (inline)

src/envdrift/scanner/native.py
jainal09: Fixed in 77567a2 — great catch. _should_ignore now routes slash-free patterns (like [!.]*.test) to basename/path-part matching only, never the full relative path, so a nested apps/web/.env.test is no longer swallowed. New _pattern_matches helper. Regression: test_should_ignore_does_not_swallow_nested_env_test asserts pkg.test stays ignored at any depth while .env.test (top-level AND nested) is scanned — proven failing pre-fix.

src/envdrift/scanner/native.py
jainal09: Fixed in 77567a2 — _decode_unicode_text now decodes BOM/stride-identified text with errors="replace" instead of returning None on a UnicodeDecodeError, so an odd-length / truncated UTF-16 file is scanned rather than falling through to _looks_binary and being skipped. Regressions: test_truncated_utf16_bom_is_still_decoded_not_skipped (unit) + test_odd_length_utf16_env_file_secret_detected (end-to-end), both proven failing pre-fix.

docs/guides/guard.md
jainal09: Fixed in 77567a2 — corrected the guard.md paragraph: it now says UTF-8 and UTF-16 (with or without a BOM) are decoded, BOM-prefixed UTF-32 is decoded, and BOM-less UTF-32 (no reliable signature) is the one encoding still treated as binary. Matches the code.

src/envdrift/scanner/native.py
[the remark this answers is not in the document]
jainal09: Fixed in 77567a2 — applied your errors="replace" suggestion to the BOM decode path (and the stride path), so a clear-BOM UTF-16 file is always decoded and scanned regardless of an individual bad code unit. Thanks for the precise repro.

src/envdrift/scanner/native.py
```diff
@@ -221,30 +222,31 @@
...
-    """Return True if a path's filename is a ``.env`` / ``.env.*`` file."""
+    """Return True if a path's filename matches an env-file naming shape.
+
+    Recognizes the leading-dot convention (``.env``, ``.env.production``) and
+    the trailing-suffix convention (``production.env``, ``database.env``) used
+    by ``docker --env-file``, direnv and many CI systems (#477).
+    """
     file_name = Path(rel_path).name
-    return file_name == ".env" or file_name.startswith(".env.")
+    return file_name == ".env" or file_name.startswith(".env.") or file_name.endswith(".env")
```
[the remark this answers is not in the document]
jainal09: Fixed in 77567a2 — removed the redundant file_name == ".env" branch; _is_env_file is now startswith(".env.") or endswith(".env") (the latter already covers ".env" itself).

src/envdrift/scanner/native.py
```diff
@@ -619,16 +688,25 @@ def _scan_file(self, file_path: Path) -> list[ScanFinding]:
...
+        content = _decode_unicode_text(raw)
+        if content is None:
+            # Skip genuinely-binary files (compiled blobs etc.) to avoid noise —
+            # but a single stray NUL must NOT hide secrets in an otherwise-text
+            # file. The old "any NUL in the first 8 KiB -> discard ALL findings"
+            # rule let an attacker evade guard by injecting one NUL byte, leaving
+            # a real plaintext key undetected with a clean [OK] (#22).
+            if _looks_binary(raw):
+                return findings
+            content = raw.decode("utf-8", errors="ignore")
```
jainal09, line 711: Addressed in 77567a2 — extracted _read_scannable_content (read + unicode decode + binary check + empty guard) out of _scan_file, and the errors="replace" change removed two try/except blocks from the decode path; _scan_file's complexity is back down.

jainal09, line 711: Addressed in 77567a2 — same _read_scannable_content extraction removes the nested conditional blocks that made _scan_file bumpy.

src/envdrift/scanner/native.py
```diff
@@ -274,6 +276,67 @@ def _looks_binary(raw: bytes, *, sample: int = 8192, threshold: float = 0.05) ->
...
+                return raw.decode(encoding)
+            except UnicodeDecodeError:
+                return None
+    stride_encoding = _sniff_utf16_stride(raw)
+    if stride_encoding is not None:
+        try:
+            return raw.decode(stride_encoding)
+        except UnicodeDecodeError:
+            return None
+    return None
```
jainal09, line 339: Addressed in 77567a2 — _decode_unicode_text dropped its two try/except blocks (errors="replace"), flattening the function CodeScene flagged at this line.

src/envdrift/scanner/native.py
```diff
@@ -10,6 +10,7 @@
 
 from __future__ import annotations
 
+import codecs
```
jainal09, line 13: Resolving as accepted-advisory: 'Lines of Code in a Single File' is a whole-file aggregate. native.py already keeps its structure-aware filters in a sibling module (_native_filters) to stay within budget; this PR's net add is small (one extracted helper + comments) and further splitting purely to satisfy the aggregate would scatter the scanner's core logic and hurt readability. Consistent with treating CodeScene as advisory (the per-method flags were each fixed).

tests/scanner/test_ignores.py
```diff
@@ -653,3 +653,88 @@ def test_empty_file(self, tmp_path: Path):
 
         # Empty file, finding should pass through
         assert len(result) == 1
+
```
jainal09, line 673: Resolving as accepted-advisory: 'Low Cohesion' on a TEST module (test_ignores.py) — test files group many small independent cases by design, so cohesion metrics meant for production modules don't translate. No production code is affected.

jainal09:
@coderabbitai full review

tests/integration/test_guard_native_gaps_e2e.py
```diff
@@ -0,0 +1,309 @@
...
+        (tmp_path / sub).mkdir()
+        (tmp_path / sub / "secrets.conf").write_text(f"TOKEN={_GHP_TOKEN}\n", encoding="utf-8")
+    _commit_all(tmp_path)
+
+    no_arg = _guard_json(
+        _run_envdrift(["guard", "--native-only", "--no-auto-install", "--json"], cwd=tmp_path)
+    )
+    rel_arg = _guard_json(
+        _run_envdrift(["guard", "--native-only", "--no-auto-install", "--json", "."], cwd=tmp_path)
+    )
```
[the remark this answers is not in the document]
jainal09: Fixed in 7383607: the parity test now captures both `CompletedProcess` results and asserts `no_arg_result.returncode == rel_arg_result.returncode == 1` (the planted `config/secrets.conf` PAT is CRITICAL) before parsing JSON, so a failure-mode regression can no longer pass on parseable output alone. Also fixed the related outside-diff finding from the review body in the same commit: `IgnoreConfig.from_dict` now carries `noisy_rule_paths`, with a round-trip regression test in `tests/scanner/test_ignores.py`.

src/envdrift/scanner/native.py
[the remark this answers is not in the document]
jainal09: Declining as accepted advisory, consistent with the identical file-level finding earlier on this PR and the precedent in #522: the module is at 629 LoC because #477 mandates the added decoding/classification logic (`_decode_unicode_text`, `_pattern_matches`, resolved-path ignores), each already reviewed at method level with no per-method health flags outstanding. Splitting `native.py` purely to duck under the 600-line soft threshold would create churn that conflicts with the stacked PRs #526/#533 touching the same file. CodeScene is a non-required check on this repo.

tests/scanner/test_ignores.py
```diff
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import dataclasses
```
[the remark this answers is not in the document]
jainal09: Declining: test_ignores.py's added responsibilities are the review-mandated from_dict round-trip regression tests (noisy_rule_paths was silently dropped — a real data-loss bug fixed on this PR); the module mirrors scanner/ignores.py 1:1 per the suite convention. Precedent: same disposition on #511/#527 coverage-suite threads. CodeScene is non-required.

PR #506, opened 2026-06-12
fix(registry): lock writes, preserve corrupt files, fail cleanly on bad JSON (#492)

### LINKED ISSUE #492: registry: no file locking, corruption silently wipes registrations, malformed JSON crashes agent commands
## Summary

`ProjectRegistry` (the agent's `~/.envdrift/projects.json`) has three related robustness gaps: registration uses
an unlocked load→mutate→whole-file-replace cycle so concurrent registers silently lose entries, a corrupt
registry is silently reset to empty and then persisted (destroying all prior registrations behind a success
message), and valid-JSON-wrong-shape files escape the loader's catch tuple as raw tracebacks. In every case the
user sees either clean success or an unexplained crash — and projects silently stop being watched by the
guardian, so auto-encryption never fires for them. Every item below was reproduced twice (original reviewer +
independent verifier) on origin/main commit cb68c97.

## Findings

- [ ] **Concurrent `envdrift agent register` silently loses registrations: 10 parallel registers all print
  "Registered project" / exit 0 but only 5-7 survive in projects.json.** `src/envdrift/agent/registry.py:139`
  **Repro:** with a scratch `HOME`, launch
  `for i in $(seq 1 10); do envdrift agent register /tmp/race/proj$i & done; wait`, then count
  `projects.json` entries vs success messages.
  **Expected:** all 10 projects present (exclusive lock around load+save, or re-read-merge before replace), or at
  minimum a failure exit for the lost ones.
  **Actual:** every command reports success; only 5/10 (verifier, both runs) to 6-7/10 (original) entries survive —
  last writer wins, and the lost projects are never watched by the agent.
  **Verifier note:** `register()` is load → append → atomic whole-file replace; no `flock`/`lockf`/`fcntl`/
  `filelock` anywhere in the module.

- [ ] **A corrupt/truncated projects.json is silently WIPED on the next `agent register`: all previously
  registered projects are dropped with no warning while the command prints clean success.**
  `src/envdrift/agent/registry.py:94`
  **Repro:** register projA; truncate projects.json mid-string; `envdrift agent register /tmp/projB;`
  `cat ~/.envdrift/projects.json` → only projB.
  **Expected:** warn the user ("registry was corrupt, N project(s) discarded") and/or back up the corrupt file
  before starting fresh.
  **Actual:** the loader silently sets `self._projects = []` on `JSONDecodeError`; `register()` then saves,
  permanently replacing the corrupt file with a registry containing only the new project, exit 0.
  **Verifier note:** fully silent and unrecoverable; the guardian simply stops watching the lost projects.

- [ ] **Malformed-but-valid-JSON registry (top-level array, or string entries) crashes every agent command with a
  raw AttributeError/TypeError Rich traceback.** `src/envdrift/agent/registry.py:93`
  **Repro:** `echo '[]' > ~/.envdrift/projects.json; envdrift agent status` → AttributeError traceback;
  `echo '{"projects": ["/tmp/foo"]}' > ...; envdrift agent list` → TypeError traceback.
  **Expected:** treated like other corrupt registries — clean warning + recovery (or a clean "registry is corrupt
  at <path>" error), never an uncaught traceback.
  **Actual:** the `load()` catch tuple is only `(json.JSONDecodeError, OSError, KeyError)`, so
  `data.get(...)` on a list and `ProjectEntry.from_dict` on a string escape to the user, exit 1.
  **Verifier note:** reproduced on `agent status`, `agent list`, and `agent register`.

## Suggested fix direction

- Wrap `register()`/`unregister()`'s load+save cycle in an exclusive file lock (`fcntl.flock` / `msvcrt` on
  Windows, or the `filelock` package) — or re-read and merge under the lock before the atomic replace.
- On corrupt-registry recovery, print a warning naming how many entries were discarded and rename the corrupt
  file aside (e.g. `projects.json.corrupt-<ts>`) instead of silently persisting the wipe.
- Validate the loaded document's shape in `load()` (dict with a `projects` list of dicts) and extend the catch
  to convert `AttributeError`/`TypeError` into the same clean corrupt-registry path.

_Part of the 2026-06 final deep-review (batch: CFG-2). Each item was independently reproduced and re-verified before filing._

jainal09:
## Summary

`ProjectRegistry` (`~/.envdrift/projects.json`) had three related robustness gaps (#492): an unlocked
load->mutate->replace write cycle that silently lost concurrent registrations, silent wipe-and-persist of a
corrupt registry behind a clean success message, and raw AttributeError/TypeError tracebacks on
malformed-but-valid-JSON registries. All three are fixed, each with regression tests proven failing on the
unfixed code first.

## Fixes

| Issue checklist item | Change |
| --- | --- |
| Concurrent `agent register` silently loses registrations (`registry.py:139`) | `register()`/`unregister()`/`clear()` hold an exclusive inter-process lock on a sidecar `projects.json.lock` (`fcntl.flock` on POSIX, `msvcrt.locking` on Windows — stdlib only, no new deps) and **re-read the registry under the lock** before mutating, so writers merge instead of last-writer-wins. Lock timeout (10s) raises `RegistryLockError`; the CLI exits 1 with a clean message instead of hanging or corrupting state. |
| Corrupt/truncated projects.json silently WIPED on next register (`registry.py:94`) | `load()` records a `RegistryCorruption` notice instead of silently resetting. Read commands (`list`/`status`) warn and leave the file untouched; the next write first moves the corrupt original aside to `projects.json.corrupt-<timestamp>` and the CLI prints the backup path. A failed backup rename is reported truthfully ("could not be backed up") rather than inventing one. |
| Malformed-but-valid JSON crashes every agent command with a raw traceback (`registry.py:93`) | `load()` now shape-validates the document (dict with a `projects` list of dicts carrying a string `path`; non-string `added` normalized) and converts every corruption flavor — invalid JSON, wrong shape, unreadable file — into the same clean corrupt-registry warning path. `agent status` / `list` / `register` exit cleanly. |

Docs updated in the same PR: registry-safety section in `docs/guides/agent-setup.md` and a note in
`docs/agent-spec.md` (the Go agent's watcher already ignores the sidecar lock/backup files; no Go change
needed — it only reacts to `projects.json` itself).

## Test evidence

All regression tests were written first and run against unfixed `origin/main` (e93c954); every one failed for
the exact reason in the issue, then passed after the fix.

**Pre-fix failures (verbatim):**

- `tests/integration/test_cli_e2e.py::TestRegistryConcurrency::test_concurrent_registers_keep_all_projects`
  (10 parallel real `envdrift agent register` subprocesses, shared scratch HOME):
  `AssertionError: lost 2 of 10 concurrent registrations: missing=[...raceproj0, ...raceproj4]`
- `tests/unit/test_agent_registry_robustness.py::TestRegistryLocking::test_parallel_register_threads_lose_no_entries`:
  `AssertionError: lost 15 of 16 concurrent registrations`
- `...::TestRegistryLocking::test_interleaved_instances_do_not_lose_updates` (deterministic lost-update repro):
  projA missing from the saved registry
- `...::TestCorruptRegistryPreservation::test_register_after_truncation_backs_up_corrupt_file`:
  `AssertionError: corrupt registry must be backed up before overwrite — assert 0 == 1`
- `...::TestMalformedRegistryShapes::test_load_never_raises_and_flags_corruption[top-level-array]`:
  `AttributeError: 'list' object has no attribute 'get'`
- `tests/integration/test_cli_e2e.py::TestRegistryDurability::test_registry_string_entries_warn_cleanly_via_cli`
  (real CLI subprocess): exit 1 with `TypeError: string indices must be integers, not 'str'` Rich traceback

**Post-fix:** all of the above pass; concurrency tests ran 3x consecutively green. Cross-platform: subprocess
env is a copy of `os.environ` with `HOME`/`USERPROFILE` redirected, explicit `encoding=` everywhere, no fork
assumptions (plain `subprocess.Popen` + threads), no POSIX-only assertions; the integration tests run on the
macOS/Windows lanes via `cross-platform-integration.yml` and the unit tests on all three OSes via
`cross-platform.yml`.

**Gates:** `ruff check` / `ruff format --check` / `pyrefly` / `bandit` clean;
`pytest -m "not integration"`: 2550 passed; `tests/integration/test_cli_e2e.py`: 15 passed;
`make lint-docs`: 0 errors. Touched-file coverage: `agent/registry.py` 95%, `cli_commands/agent.py` 99%
(uncovered registry lines are the win32 lock branches, unreachable on the Linux coverage lane).

Fixes #492

<a href="https://cubic.dev/pr/jainal09/envdrift/pull/506?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

This PR fixes three robustness gaps in `ProjectRegistry` (#492): concurrent `register`/`unregister` calls now hold an exclusive inter-process lock (`fcntl.flock` on POSIX, `msvcrt.locking` on Windows) and re-read the registry under the lock so parallel writers merge instead of clobbering; a corrupt `projects.json` is now surfaced as a `RegistryCorruption` notice and backed up atomically before any write replaces it; and malformed-but-valid-JSON registries (top-level arrays, string entries) are shape-validated in `_parse_registry_document` so they never escape `load()` as raw `AttributeError`/`TypeError`.

- **Locking**: `_exclusive_lock()` polls a sidecar `.lock` file, distinguishing contention errnos (retry-until-timeout) from transient I/O blips (retry ×3 then re-raise as-is) from persistent errors (re-raise immediately), with `RegistryLockError` surfaced cleanly at the CLI exit-1 boundary.
- **Backup**: `_backup_corrupt_file()` moves the corrupt original to `projects.json.corrupt-<timestamp>` (with a same-tick numeric suffix to avoid clobbers) and records the path (or `backup_failed=True`) in `RegistryCorruption`; a re-entry guard blocks a second `save()` from moving the freshly-written valid registry aside.
- **Regression coverage**: 515-line `test_agent_registry_robustness.py` plus new CLI and install-agent tests verify each fix with deterministic lost-update reproductions, thread/process concurrency hammers, and failure-injection via `monkeypatch`.

Safe to merge — all three robustness gaps are closed with well-scoped changes, the previous review's re-entry and errno-filtering concerns have been addressed, and the test suite covers concurrent writes, backup failure injection, and edge-case JSON shapes deterministically.

The locking implementation correctly separates contention retries from transient I/O retries, the backup re-entry guard blocks a second save() from moving a freshly written valid registry aside, and shape validation is exhaustive across all corruption flavors. No logic errors or unsafe code paths were found in the changed files.

| Filename | Overview |
|----------|----------|
| src/envdrift/agent/registry.py | Core fix: adds RegistryLockError/RegistryCorruption, exclusive inter-process lock, shape validation, and atomic backup-before-overwrite. Logic is sound; re-entry guard, transient-error retry budget, and backup-path collision avoidance all implemented correctly. |
| src/envdrift/cli_commands/agent.py | Adds RegistryLockError catch (exit 1) and warn_registry_corruption calls after every mutating command; refactors post-register guidance into _report_post_register_guidance to keep complexity flat. |
| src/envdrift/cli_commands/agent_utils.py | New warn_registry_corruption() helper cleanly reports all three corruption outcomes (backed up, backup failed, still-in-place) without raising; shared by agent and install command modules. |
| src/envdrift/cli_commands/install.py | Extracts _register_current_project() with lock-error handling and corruption surfacing, matching the agent register command's hardening; no behavior change for the happy path. |
| tests/unit/test_agent_registry_robustness.py | New 515-line regression suite covering all three bug classes: shape validation, backup preservation, re-entry guard, inter-thread and inter-process concurrency hammers, and transient/persistent lock-error injection. |
| tests/integration/test_cli_e2e.py | Extended with CLI-level regression tests: corrupt-JSON backup verification, top-level-array and string-entry warning paths, and a 10-process concurrent-register stress test. |
| tests/unit/test_agent_cli.py | New TestAgentRegistryCorruptionCli class adds lock-held, backup-failure, unregister-miss, and corruption-warning CLI unit tests. |
| tests/unit/test_install_cli.py | New TestInstallAgentRegistryGuard tests verify the install-agent lock-held exit-1 path and corruption surfacing match the agent register command's behaviour. |
| src/envdrift/agent/__init__.py | Exports RegistryCorruption and RegistryLockError from the public package surface. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant CLI as CLI (register/unregister)
    participant Reg as ProjectRegistry
    participant Lock as projects.json.lock
    participant FS as projects.json

    CLI->>Reg: register(project_path)
    Reg->>Lock: os.open() + _acquire_lock()
    alt lock contention
        Lock-->>Reg: OSError(EWOULDBLOCK) — retry until timeout
        Reg-->>CLI: raise RegistryLockError
        CLI-->>CLI: print error, exit(1)
    end
    Reg->>FS: load() under lock
    alt corrupt / wrong shape
        FS-->>Reg: JSONDecodeError / ValueError / OSError
        Reg->>Reg: "_corruption = RegistryCorruption(detail=…)"
        Reg->>Reg: "_projects = []"
    else valid
        FS-->>Reg: list[ProjectEntry]
    end
    Reg->>Reg: append / remove entry
    Reg->>Reg: _backup_corrupt_file()
    alt corruption notice present and no prior attempt
        Reg->>FS: Path.replace(projects.json.corrupt-timestamp)
        Reg->>Reg: "notice.backup_path = backup_path"
    end
    Reg->>FS: _write_atomic() — tempfile + fsync + rename
    Reg->>Lock: _release_lock() + os.close(fd)
    Reg-->>CLI: True / False
    CLI->>Reg: warn_registry_corruption(registry, console)
    alt notice.backup_path set
        CLI-->>CLI: print backup path + re-register hint
    else notice.backup_failed
        CLI-->>CLI: print could not be backed up
    else notice with no write yet
        CLI-->>CLI: print treating as empty, will back up on next register
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant CLI as CLI (register/unregister)
    participant Reg as ProjectRegistry
    participant Lock as projects.json.lock
    participant FS as projects.json

    CLI->>Reg: register(project_path)
    Reg->>Lock: os.open() + _acquire_lock()
    alt lock contention
        Lock-->>Reg: OSError(EWOULDBLOCK) — retry until timeout
        Reg-->>CLI: raise RegistryLockError
        CLI-->>CLI: print error, exit(1)
    end
    Reg->>FS: load() under lock
    alt corrupt / wrong shape
        FS-->>Reg: JSONDecodeError / ValueError / OSError
        Reg->>Reg: "_corruption = RegistryCorruption(detail=…)"
        Reg->>Reg: "_projects = []"
    else valid
        FS-->>Reg: list[ProjectEntry]
    end
    Reg->>Reg: append / remove entry
    Reg->>Reg: _backup_corrupt_file()
    alt corruption notice present and no prior attempt
        Reg->>FS: Path.replace(projects.json.corrupt-timestamp)
        Reg->>Reg: "notice.backup_path = backup_path"
    end
    Reg->>FS: _write_atomic() — tempfile + fsync + rename
    Reg->>Lock: _release_lock() + os.close(fd)
    Reg-->>CLI: True / False
    CLI->>Reg: warn_registry_corruption(registry, console)
    alt notice.backup_path set
        CLI-->>CLI: print backup path + re-register hint
    else notice.backup_failed
        CLI-->>CLI: print could not be backed up
    else notice with no write yet
        CLI-->>CLI: print treating as empty, will back up on next register
    end
```

</a>

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/492-registr..."](https://github.com/jainal09/envdrift/commit/6b506247b246161cb4071289b78d65cad1313b9e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37198523)</sub>


docs/guides/agent-setup.md
```diff
@@ -167,6 +167,22 @@ Registry: /Users/you/.envdrift/projects.json
...
+The CLI guards the registry file against races and corruption:
+
+- **Concurrent registers are safe.** `register`/`unregister` hold an exclusive
+  lock on a sidecar `~/.envdrift/projects.json.lock` file and re-read the
+  registry under it, so parallel commands (e.g. several setup scripts at once)
+  never overwrite each other's entries. If the lock cannot be acquired within
+  10 seconds the command fails with exit code 1 instead of corrupting state.
+- **A corrupt registry is never silently wiped.** If `projects.json` is
+  truncated or malformed, read commands (`list`, `status`) warn and treat it as
+  empty without touching the file. The next `register`/`unregister` moves the
```
jainal09: 🤖 **envdrift review-bot** · low · doc-mismatch


Both the new docs and the new CLI warning state the corrupt file "will be backed up ... by the next register/**unregister**":

- `docs/guides/agent-setup.md:181-183`: "The next `register`/`unregister` moves the corrupt original aside to `projects.json.corrupt-<timestamp>` (and tells you where) before writing a fresh registry"
- `src/envdrift/cli_commands/agent.py:146-150` (`_warn_registry_corruption` else-branch): "The file is untouched and will be backed up to a .corrupt-<timestamp> file by the next register/unregister."

But `unregister` can **never** trigger the backup: a corrupt registry always loads as empty (`load()` sets `self._projects = []` plus the corruption notice), so the removal loop in `unregister()` always misses and returns `False` *without* calling `save()` — and `_backup_corrupt_file()` only runs inside `save()`. The PR's own tests prove this: `test_unregister_miss_on_corrupt_registry_preserves_file` asserts `_corrupt_backups(registry_path) == []` after an unregister on a corrupt file. Only `register` (or the un-exposed `clear()`) performs the backup.

Impact: a user with a corrupt registry who follows the doc/CLI hint and runs `envdrift agent unregister` will see the warning repeat forever with no backup ever made, and may believe recovery happened when it didn't.

Suggested direction: change both the doc sentence and the CLI message to name `register` only (or make a no-op `unregister` on a corrupt registry also perform the backup+rewrite, which would match the current wording).

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit fe5b34b


The registry locking/corruption hardening is solid for `envdrift agent register/unregister`, but the `envdrift install agent` path still hits the new `RegistryLockError` unguarded, and the recovery docs overpromise what `unregister` does.

**Findings (2):**
- [medium] `envdrift install agent` register path: uncaught RegistryLockError raw traceback + silent corruption recovery — `src/envdrift/cli_commands/install.py` (see below if not anchored)
- [low] Docs and CLI warning promise `unregister` will back up a corrupt registry — it never can — `docs/guides/agent-setup.md:181` (inline)

### Not anchored inline

🤖 **envdrift review-bot** · medium · bug


The fix hardens `envdrift agent register/unregister` but misses the **other** CLI caller of `register_project()`: the `envdrift install agent` flow in `src/envdrift/cli_commands/install.py:442`:

```python
            from envdrift.agent.registry import register_project

            success, message = register_project()
            if success:
                console.print(f"[green]✓[/green] {message}")
```

There is no `try/except RegistryLockError` here (I verified no enclosing `try` wraps this call), and `_warn_registry_corruption` is private to `cli_commands/agent.py` and never invoked. Two consequences:

1. **New raw-traceback path.** `register_project()` never raised before this PR; now it raises `RegistryLockError` after the 10s timeout. With the registry lock contended (the exact scenario this PR targets — e.g. several setup scripts running `install agent` / `agent register` at once, or a stuck process), `envdrift install agent` dies with an uncaught Rich traceback — the precise failure mode #492 set out to eliminate ("never an uncaught traceback"). `cli_commands/agent.py:217-221` shows the intended handling that this path lacks.
2. **Silent corruption recovery.** If `projects.json` is corrupt, `register()` inside this path backs it up and replaces it (good), but the user sees only "✓ Registered project: ..." — no corruption warning, no backup path, no "prior registrations were discarded". That contradicts the PR's own guarantee ("never silently", `docs/guides/agent-setup.md` registry-safety section) and the warning contract tested for the `agent register` command.

Suggested direction: wrap the `register_project()` call in `install.py` with the same `RegistryLockError` handler (warn + continue or exit 1 — installation shouldn't traceback), and surface `get_registry().corruption` there (e.g. by moving `_warn_registry_corruption` somewhere shared). Add a regression test mirroring `test_register_lock_held_exits_one_with_clean_error` for the `install agent` command.

src/envdrift/agent/registry.py
```diff
@@ -35,37 +83,71 @@ class ProjectEntry:
...
+    if not isinstance(raw_projects, list):
+        raise ValueError(f"'projects' must be a list, got {type(raw_projects).__name__}")
+    entries: list[ProjectEntry] = []
+    for item in raw_projects:
+        if not isinstance(item, dict):
+            raise ValueError(f"project entries must be objects, got {type(item).__name__}")
+        if not isinstance(item.get("path"), str):
+            raise ValueError("project entry is missing a string 'path'")
+        entries.append(ProjectEntry.from_dict(item))
+    return entries
```
jainal09: Fixed in 155c5bb: per-entry validation is extracted into `_parse_registry_entry`, so `_parse_registry_document` is now a flat shape check plus a comprehension (cyclomatic complexity 4, under the threshold). Behavior is unchanged — the full malformed-shape parametrized suite in `tests/unit/test_agent_registry_robustness.py` still passes.

src/envdrift/cli_commands/agent.py
```diff
@@ -186,7 +213,13 @@ def register(
...
 
     # Register the project
-    success, message = register_project(project_path)
+    try:
+        success, message = register_project(project_path)
+    except RegistryLockError as exc:
+        console.print(f"[red]✗[/red] Could not lock the agent registry: {exc}")
+        raise typer.Exit(1) from exc
+
+    _warn_registry_corruption(get_registry())
```
jainal09: Fixed in 155c5bb: the success-path follow-up (no-config hint / guardian config load + report) is extracted into `_report_post_register_guidance`, and the outcome handling is flattened to early returns. `register` is back under the complexity threshold and all existing CLI register tests pass unchanged.

tests/unit/test_agent_cli.py
```diff
@@ -429,3 +429,152 @@ def test_agent_help(self):
         assert "unregister" in result.stdout
         assert "list" in result.stdout
         assert "status" in result.stdout
+
```
jainal09, line 432: Fixed in 155c5bb: the two near-identical lock-held CLI tests are collapsed into one `@pytest.mark.parametrize("command", ["register", "unregister"])` test, keeping the stricter assertions (explicit exit code, no raw exception) for both commands.

src/envdrift/agent/registry.py
```diff
@@ -127,70 +282,105 @@ def _apply_permissions(self) -> None:
...
+        timestamp = datetime.now(UTC).strftime("%Y%m%dT%H%M%S%f")
+        backup_path = self._path.with_name(f"{self._path.name}.corrupt-{timestamp}")
+        try:
+            self._path.replace(backup_path)
+        except OSError:
+            # The save still proceeds; record that the corrupt bytes are gone
+            # so the CLI reports the truth instead of a phantom backup.
+            self._corruption.backup_failed = True
+            return
+        self._corruption.backup_path = backup_path
```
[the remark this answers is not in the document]
jainal09: Fixed in 155c5bb: the guard now also short-circuits on `backup_failed`, exactly as suggested. Regression test `test_second_save_after_failed_backup_keeps_valid_registry` drives the scenario end-to-end: a failed backup rename followed by a second `save()` without an intervening `load()` — pre-fix that second save renamed the freshly written valid `projects.json` to a `.corrupt-*` file (verified the new test fails on the old code); post-fix the valid registry stays put and no phantom backup appears. While in there, backup names also gained a same-tick numeric suffix (`_next_backup_path`) so repeated corruption events can never clobber an earlier backup.

src/envdrift/agent/registry.py
```diff
@@ -103,6 +204,60 @@ def _normalize_path(self, project_path: Path) -> Path:
...
+                    fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                return
+            except OSError:
+                if time.monotonic() >= deadline:
+                    raise RegistryLockError(
+                        f"timed out after {self._lock_timeout:g}s waiting for the "
+                        f"registry lock: {self._lock_path} (is another envdrift "
+                        "process stuck?)"
+                    ) from None
+                time.sleep(_LOCK_POLL_INTERVAL)
```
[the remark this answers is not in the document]
jainal09: Fixed in 155c5bb, with one deliberate refinement over the suggestion: contention errnos (`EWOULDBLOCK`/`EAGAIN`/`EACCES`, now in `_LOCK_CONTENTION_ERRNOS`) poll until the lock timeout as before, while any other OSError is retried a bounded 3 times (`_TRANSIENT_LOCK_ERROR_RETRIES`, ~0.15s) to absorb transient blips such as AV scanners touching the lock file, then re-raised **as-is** — so persistent `EBADF`/`EIO` surface with the real error in well under a second instead of a misleading 10s "timed out waiting for registry lock". Two regression tests cover both sides: `test_transient_lock_oserror_retries_then_succeeds` (injected flaky lock, two EIO failures then success) and `test_persistent_lock_io_error_raises_original_error_fast` (persistent EBADF raises the original OSError in <2.5s, verified to fail on the old code with the misreported RegistryLockError).

src/envdrift/agent/registry.py
```diff
@@ -127,70 +282,105 @@ def _apply_permissions(self) -> None:
         with suppress(OSError):
             self._path.chmod(0o600)
 
+    def _backup_corrupt_file(self) -> None:
+        """Move a corrupt registry aside before a save replaces it (#492)."""
+        if self._corruption is None or self._corruption.backup_path is not None:
+            return
```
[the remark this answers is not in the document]
jainal09: Fixed in 155c5bb — the guard now includes `or self._corruption.backup_failed`, with a comment explaining the re-entry invariant. See the sibling thread on line 300 for the full regression-test details (`test_second_save_after_failed_backup_keeps_valid_registry`).

docs/guides/agent-setup.md
jainal09: Fixed in 155c5bb by making the docs and CLI tell the truth rather than changing `unregister` semantics (a no-op unregister staying read-only is the safer behavior, and the PR's own preservation test depends on it):

- `docs/guides/agent-setup.md` now says the next **`register`** performs the backup and explicitly states an `unregister` never does (a corrupt registry loads as empty, nothing to remove, file left untouched).
- The `warn_registry_corruption` else-branch hint now reads "…by the next register." with a code comment explaining why `unregister` can never trigger it.
- `test_unregister_miss_on_corrupt_registry_warns_without_backup` now asserts the hint names only `register` (the assertion fails against the old "register/unregister" wording).

Also picked up the summary comment's medium finding in the same commit: the `envdrift install agent` register step at `install.py:442` now catches `RegistryLockError` (clean message + exit 1 + hint to re-run `envdrift agent register`) and surfaces corruption recovery via the shared `warn_registry_corruption` helper, with CLI regression tests for both paths.

src/envdrift/agent/registry.py
```diff
@@ -103,6 +221,73 @@ def _normalize_path(self, project_path: Path) -> Path:
...
+                    transient_errors += 1
+                    if transient_errors > _TRANSIENT_LOCK_ERROR_RETRIES:
+                        raise
+                elif time.monotonic() >= deadline:
+                    raise RegistryLockError(
+                        f"timed out after {self._lock_timeout:g}s waiting for the "
+                        f"registry lock: {self._lock_path} (is another envdrift "
+                        "process stuck?)"
+                    ) from None
+                time.sleep(_LOCK_POLL_INTERVAL)
```
jainal09, line 275: Fixed in 9bffd3d: the single lock attempt (`_try_lock`) and the retry/give-up decision (`_handle_lock_error`) are extracted, leaving `_acquire_lock` a flat try/except polling loop under the threshold. Behavior is pinned by the lock regression tests (contention timeout, transient-EIO retry, persistent-EBADF fast-fail).

src/envdrift/agent/registry.py
```diff
@@ -127,70 +312,127 @@ def _apply_permissions(self) -> None:
...
+    def _backup_corrupt_file(self) -> None:
+        """Move a corrupt registry aside before a save replaces it (#492)."""
+        if (
+            self._corruption is None
+            # Re-entry guard: once a backup was made (or its rename failed),
+            # never attempt another for the same corruption notice — a second
+            # save() without an intervening load() would otherwise move the
+            # freshly written *valid* registry aside as if it were corrupt.
+            or self._corruption.backup_path is not None
+            or self._corruption.backup_failed
```
jainal09: Fixed in 9bffd3d: the triple-clause guard is split into a plain `notice is None or not exists` early return plus a named `backup_already_attempted` boolean, so no conditional carries more than one logical operator. Semantics unchanged (re-entry regression test still green).

src/envdrift/agent/registry.py
jainal09, line 275: Fixed in 9bffd3d by the same extraction as the Complex Method finding on this function: `_acquire_lock` no longer has nested conditional blocks — the nesting moved into the single-purpose `_handle_lock_error` helper, which stays within thresholds.

src/envdrift/cli_commands/install.py
```diff
@@ -437,9 +437,18 @@ def install_agent(
         if find_config() is not None:
             console.print("\n[bold]Registering current project...[/bold]")
             # Use our own registry instead of calling the binary
-            from envdrift.agent.registry import register_project
+            from envdrift.agent.registry import RegistryLockError, get_registry, register_project
```
jainal09: Fixed in 9bffd3d: the whole post-install registration step (config check, lock-error handling, corruption warning) moved into `_register_current_project`, so `install_agent` is now below its pre-PR complexity (the +2 delta this thread flagged is gone, and then some). The 29-baseline monolith itself predates this PR and is untouched otherwise.

Issue #470, opened 2026-06-12
lock: encryption-state heuristics bless plaintext as 'ready to commit' and flag fully-encrypted files


jainal09:
## Summary

`envdrift lock` decides whether a file is "already encrypted" with weak inline heuristics (a >=90% ciphertext-line
ratio at `sync.py:1529-1542` and an any-ciphertext regex at `encryption_helpers.py:120-121`) instead of the correct
predicates that already exist in the codebase (`has_plaintext_secret_value`, `is_dotenvx_public_key_var`,
`_is_fully_encrypted`). The result cuts both ways: mixed files carrying a freshly added plaintext secret are skipped
and blessed "[OK] Lock complete! ... ready to commit" with exit 0 (a false security PASS in the documented CI/pre-commit
gate), while fully-encrypted files with fewer than 9 variables are forever flagged "partially encrypted" with exit 1.
Every item below was reproduced twice (original reviewer + independent verifier) on origin/main commit cb68c97.

## Findings

- [ ] **`lock --all` treats a mixed-state `.secret` (one encrypted value + one fresh plaintext) as "already
  encrypted", leaves the new secret plaintext, deletes the combined file, and exits 0.** `src/envdrift/cli_commands/sync.py:1804`
  **Repro:** envdrift.toml with `[vault]` hashicorp + one `[[vault.sync.mappings]]` + a `[partial_encryption]` combine
  env; `.env.production.secret` = `API_KEY=encrypted:...` plus appended `NEW_SECRET=plain`; then
  `VAULT_TOKEN=test-root-token envdrift lock --all --force; grep NEW_SECRET .env.production.secret`
  **Expected:** mixed `.secret` is re-encrypted (push's `_is_fully_encrypted` predicate); never exit 0 with plaintext.
  **Actual:** ".secret - skipped (already encrypted)", combined file deleted, exit 0, NEW_SECRET still plaintext.
  **Verifier note:** control run confirmed `push` on the same file re-encrypts NEW_SECRET; only `lock --all` leaks.

- [ ] **`lock` / `lock --check` bless a mixed dotenvx file as "already encrypted" (exit 0) once >=90% of value lines
  are ciphertext, leaving a fresh plaintext secret in cleartext.** `src/envdrift/cli_commands/sync.py:1530`
  **Repro:**
  ```bash
  for i in $(seq 1 18); do echo "SECRET_$i=value$i"; done > svc/.env.production
[...]

PR #510, opened 2026-06-12
fix(push): verify encryption before success and write merged files safely (#471)

### LINKED ISSUE #471: push: false success leaves plaintext, destroys combined files, and fabricates placeholder secrets
## Summary

The partial-encryption `push`/`pull --merge` seam bypasses the hardened guards the bare `envdrift encrypt` backend
already has (`has_plaintext_secret_value` read-back, `file_has_assignment`, `atomic_write` 0600): it trusts dotenvx's
exit 0, tolerates missing source files, and writes secret-bearing output with raw `write_text`. Users get
"[OK] Push complete! Source files are encrypted and ready to commit" while a secret survives in plaintext, the runtime
combined file is destroyed, or ~13 fabricated placeholder secrets are injected. Every item below was reproduced twice
(original reviewer + independent verifier) on origin/main commit cb68c97.

## Findings

- [ ] **`push` reports "[OK] Push complete! Source files are encrypted" (exit 0) but leaves the secret in plaintext
  when `.env.keys` is unwritable (read-only file or a directory).** `src/envdrift/integrations/dotenvx.py:983`
  **Repro:**
  ```bash
  # partial_encryption combine env in envdrift.toml; .clear + .secret populated
  printf '# placeholder\n' > .env.keys && chmod 444 .env.keys
  envdrift push; echo $?; grep JWT .env.production.secret .env.production   # exit 0, both still plaintext
  ```
  **Expected:** push fails non-zero when plaintext survives (the guarded `encrypt` backend read-back behavior).
  **Actual:** dotenvx prints "⚠ undefined" and exits 0; `ENCRYPT_ERROR_PATTERNS` misses it; plaintext ships into the
  committed `.secret` and the app-read combined file under a success banner.
  **Verifier note:** bare `envdrift encrypt` on the same setup errors "Encryption did not take effect" exit 1; the
  push seam never gained the #443-era post-check.

- [ ] **`push` with BOTH source files missing silently overwrites the existing combined file with an empty scaffold
  and reports success — destroying the runtime artifact.** `src/envdrift/core/partial_encryption.py:417`
  **Repro:** after a valid `push`, `rm .env.production.clear .env.production.secret; envdrift push; echo $?` →
  "Generated .env.production (0 clear + 0 encrypted)", "[OK] Push complete!", exit 0; combined file now header-only.
  **Expected:** refuse when neither source exists, exit non-zero, leave the combined file untouched (symmetric with
  `pull-partial`'s "Secret file not found" error).
  **Actual:** `combine_files` tolerates both files absent and unconditionally `write_text`s the scaffold; if the
  sources were deleted by mistake the combined file was the last copy and push destroys it.

- [ ] **`push` on an empty `.secret` fabricates ~13 placeholder secrets (HELLO, AWS_ACCESS_KEY_ID, OPENAI_API_KEY,
  ...) and reports clean success.** `src/envdrift/core/partial_encryption.py:707`
  **Repro:** combine mode: empty `.env.production.secret` → `envdrift push` prints
  "Generated .env.production (1 clear + 13 encrypted)", exit 0. Secrets-only mode (env with `secrets_only = true` +
  `secrets_dir`): empty `secrets/prod/.env.api` → "Encrypted 1 file(s)", same fabrication.
  **Expected:** refuse a file with no variable assignments — the `file_has_assignment` guard the bare `encrypt`
  backend already applies (`encryption/dotenvx.py:128-137`).
  **Actual:** `encrypt_secret_file` (:519) and `_encrypt_secrets_only_file` (:707) skip the guard; dotenvx scaffolds
  its placeholder template and push counts it as "13 encrypted".
  **Verifier note:** reproduced in both modes; bare `envdrift encrypt empty.env` correctly refuses (exit 1).

- [ ] **`pull --merge` (and push's combine) write the merged file containing DECRYPTED plaintext secrets
  world-readable at 0644 via raw `write_text`, not the hardened 0600 `atomic_write`.** `src/envdrift/cli_commands/sync.py:402`
  **Repro:** partial-encryption setup, `envdrift push && envdrift pull --merge;`
  `stat -f '%Sp' svc/.env.production.combined` → `-rw-r--r--` holding plaintext `API_KEY=supersecret-merged`.
  **Expected:** secret-bearing files are created 0600 via `sync/operations.py` `atomic_write` (as `.env.keys` is).
  **Actual:** `_write_merged_combined_file` (sync.py:402) and `combine_files` (partial_encryption.py:418) write at the
  process umask — the private key gets 0600 but a file full of plaintext secrets gets 0644.
  **Verifier note:** local-filesystem exposure only (file is gitignored) — medium.

## Suggested fix direction

- Add a post-encrypt read-back (`has_plaintext_secret_value`) at the push seam — either inside
  `DotenvxWrapper.encrypt` or in `encrypt_secret_file`/`_encrypt_secrets_only_file` — so surviving plaintext fails the
  command exactly like the guarded backend.
- In `combine_files`, error out when neither `clear_file` nor `secret_file` exists instead of writing the scaffold.
- Apply the existing `file_has_assignment` guard before invoking dotenvx in both partial-encryption encrypt paths.
- Route combined/merged-file writes through `sync/operations.py atomic_write` (default 0o600, fchmod-on-fd, atomic
  rename) already used for `.env.keys`.

_Part of the 2026-06 final deep-review (batch: CORE-2). Each item was independently reproduced and re-verified before filing._

### LINKED ISSUE #485: tests: add regression coverage for the encryption-state criticals — no existing test can catch them
## Summary

The three most severe defects of this review (push reporting "[OK] encrypted and ready to commit" on plaintext,
the `lock --check` N/(N+1) header-counting ratio, and lock blessing mixed-state files) all pass the full 3159-test
suite silently: the only tests near each code path use fixtures shaped to avoid the breaking input, mock the
component under test, or assert a tautology. This is a guard-rail gap distinct from the product bugs (filed
separately): until these regressions exist, the same secret-leak classes can regress again unnoticed. The item below
was reproduced twice (original reviewer + independent verifier) on origin/main commit cb68c97.

## Findings

- [ ] **None of the three encryption-state criticals (push plaintext under "[OK]", lock --check ratio, lock
  mixed-state) can be caught by any existing test.** `tests/integration/test_e2e_workflows.py:184`
  **Repro:** (1) `grep -rn 'undefined\|ENCRYPT_ERROR_PATTERNS' tests/` → only a schema-loader string; the push-path
  unit tests mock `DotenvxWrapper` wholesale (test_partial_encryption.py:189,204,613,631,646). (2) the only ratio
  tests (test_cov_cli_commands_sync.py:866-889) omit the `DOTENV_PUBLIC_KEY` header that breaks production AND patch
  the backend. (3) the sole real-binary `lock --check` e2e asserts `result.returncode in (0, 1)` — "Both are valid
  outcomes" — at test_e2e_workflows.py:185, an assertion that cannot fail on the wrong exit code.
  **Expected:** per repo convention, a real regression test per leak class, xfail(strict=False, see #NNN) until fixed.
  **Actual:** all three criticals pass the full suite green on a build that leaks plaintext under "[OK]".
  **Verifier note:** test_error_handling.py:517 does build a read-only .env.keys but only runs `lock --check`
  asserting no traceback; the #413 regressions are push-only; open #443 backlogs the encrypt-side
  undefined-exit-0 class but not the push path.

## Suggested fix direction

- Add a real-binary push regression: unwritable `.env.keys` (dotenvx "⚠ undefined", exit 0) must yield a non-zero
  envdrift exit with the plaintext flagged — covers the missing post-encrypt plaintext check in
  `DotenvxWrapper.encrypt`.
- Add a `lock --check` regression on a genuinely dotenvx-encrypted small file WITH its `DOTENV_PUBLIC_KEY` header
  (e.g. 3 vars), asserting exit 0; and a mixed-state `lock`/`lock --all` regression asserting re-encryption.
- Mark all three `@pytest.mark.xfail(strict=False, reason="... (see #NNN)")` referencing the product issues, and flip
  them in the fixing PRs.
- Replace the vacuous `returncode in (0, 1)` at test_e2e_workflows.py:185 with the definite expected exit code for
  its known input.

_Part of the 2026-06 final deep-review (batch: TEST-2). Each item was independently reproduced and re-verified before filing._

jainal09:
## Summary

The partial-encryption `push` / `pull --merge` seam bypassed the hardened guards the bare
`envdrift encrypt` backend already has: it trusted dotenvx's exit 0, tolerated missing source
files, handed empty files to the encryptor, and wrote secret-bearing output with a bare
`write_text` at the process umask. Users got `[OK] Push complete! Source files are encrypted and
ready to commit` (exit 0) while a secret survived in plaintext, the runtime combined file was
destroyed, or ~13 fabricated placeholder secrets were injected.

All four checklist items of #471 are fixed, each behind a regression test that was **proven to
fail on the unfixed code first** (real `envdrift` CLI subprocess + real `dotenvx` binary — no
mocking of the behavior under test).

## Fixes (per checklist item)

1. **Push reports success while plaintext survives** (`integrations/dotenvx.py` error-pattern gap)
   — `encrypt_secret_file` / `_encrypt_secrets_only_file` now re-read the secret file after
   `dotenvx encrypt` and raise `Encryption did not take effect: ... still contains plaintext
   values` when any plaintext secret value survived (dotenvx prints `⚠ undefined` and exits 0
   when `.env.keys` is read-only, a directory, or malformed). The read-back runs **before**
   lifting the git skip-worktree protection, so a still-plaintext file stays protected. This is
   the same `has_plaintext_secret_value` post-check the guarded backend gained in #443.
2. **Push with BOTH source files missing destroyed the combined file**
   (`core/partial_encryption.py:417`) — `combine_files` now raises
   `Neither clear file (...) nor secret file (...) exists; refusing to overwrite ...` and leaves
   the existing combined file byte-for-byte untouched (symmetric with `pull-partial`'s
   "Secret file not found"). A single missing source is still tolerated.
3. **Push on an empty `.secret` fabricated ~13 placeholder secrets**
   (`core/partial_encryption.py:707`) — both encrypt seams now apply the bare backend's
   `file_has_assignment` guard and refuse with `Nothing to encrypt: ...` before dotenvx can
   scaffold its placeholder template. Covers combine mode and secrets-only mode (including
   `--check`).
4. **`pull --merge` / push's combine wrote secret-bearing files world-readable at 0644**
   (`cli_commands/sync.py:402`) — `_write_merged_combined_file` and `combine_files` now route
   through `sync/operations.py atomic_write` (0600 for fresh files, fchmod-on-fd, fsync + atomic
   rename), exactly like `.env.keys`. `atomic_write` now writes with explicit `encoding="utf-8"`
   so merged content never depends on the platform locale.

`sync.py` edits are scoped to the `pull --merge` writer (~line 402) to avoid overlap with #507
(the #470 lock/check PR).

17 new regression tests, all run against the unfixed code first:

| Test (real CLI + real dotenvx unless noted) | Pre-fix failure (exact issue symptom) | Post-fix |
| --- | --- | --- |
| e2e: push with `.env.keys` as a directory | `exit=0`, `[OK] Push complete!`, `(1 clear + 1 encrypted)`, `JWT_SECRET` plaintext in `.secret` AND combined | exit 1, `did not take effect`, source intact, no combined |
| e2e: push with read-only `.env.keys` (POSIX, non-root) | same false success | exit 1, no banner |
| e2e: push with both sources deleted | `exit=0`, `(0 clear + 0 encrypted)`, banner, combined replaced by header-only scaffold | exit 1, combined byte-identical |
| e2e: push on empty `.secret` (combine) | `exit=0`, `(1 clear + 13 encrypted)`, banner | exit 1, `Nothing to encrypt`, file still 0 bytes |
| e2e: push on empty file (secrets-only) | `Encrypted 1 file(s)`, exit 0, AWS/OpenAI placeholders fabricated | exit 1, no fabrication |
| e2e: `pull --merge` merged-file permissions | `mode 0o644 != 0o600` (holding decrypted `API_KEY`) | 0600, atomic, no temp residue |
| unit: `combine_files` both-missing (write=True/False), no-combined variant | `DID NOT RAISE` | raises, combined preserved |
| unit: `combine_files` 0600+atomic; `_write_merged_combined_file` 0600+atomic | `mode 0o644 != 0o600` | pass |
| unit: empty/comment-only refusal (combine + secrets-only + `--check`); silent-failure read-back (both seams) | `DID NOT RAISE` | pass |

Permission asserts are gated `sys.platform != "win32"`; the read-only-file e2e also skip-gates
root. Fixtures carry the exact production-shaped `DOTENV_PUBLIC_KEY` header block, and
credential-looking values are built by concatenation. Nine pre-existing unit tests that mocked
`DotenvxWrapper.encrypt` as a no-op were updated to *honest* mocks that actually encrypt — the
new read-back rightly rejects a "successful" encrypt that leaves plaintext.

Gates: `ruff check` / `ruff format --check` / `pyrefly` / `bandit` clean; full suite
`uv run pytest` green (2860 passed, 15 skipped); touched files ≥ 90% coverage
(`core/partial_encryption.py` 97%, `cli_commands/sync.py` 90%, `sync/operations.py` 97%);
`make lint-docs` clean.

Docs updated in the same PR: `docs/cli/push.md` (new Safety guarantees section),
`docs/cli/pull.md` (`--merge` permissions note), `docs/guides/partial-encryption.md`.

Fixes #471
Together with #507 (the #470 lock/check regression coverage), this completes the
regression-coverage gap tracked in #485. Closes #485

<a href="https://cubic.dev/pr/jainal09/envdrift/pull/510?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

This PR hardens the `push` / `pull --merge` partial-encryption path against four failure modes where the old code reported `[OK] Push complete!` while leaving secrets plaintext, destroying the combined file, fabricating placeholder secrets, or writing decrypted data world-readable.

- **Encryption verification**: `_encrypt_and_verify` re-reads each secret file after `dotenvx encrypt` and raises if any plaintext value survives, closing the dotenvx exit-0-without-encrypting gap; the guard fires before lifting git skip-worktree protection so a still-plaintext file stays protected.
- **Empty-file guard**: Both `encrypt_secret_file` and `_encrypt_secrets_only_file` (including `--check` mode) now call `file_has_assignment` before delegating to dotenvx, preventing the ~13-placeholder-secret fabrication.
- **Both-sources-missing guard**: `combine_files` raises immediately when neither clear nor secret file exists, leaving the existing combined file byte-for-byte intact.
- **Secure combined-file writes**: `combine_files` and `_write_merged_combined_file` now route through `atomic_write(max_permissions=0o600)` instead of bare `write_text`; the new `max_permissions` cap also tightens pre-existing world-readable files created by earlier runs.

Safe to merge — all four security gaps are closed with narrow, well-scoped changes and each fix is covered by a regression test proven to fail on the unfixed code first.

The changes are surgical and focused: each of the four fixed paths has a clear pre-condition guard or post-condition check, the existing public API of `atomic_write` is unchanged (new parameter is keyword-only and optional), `secret_lines` is still consumed for the secret-var count, and the `_honest_encrypt_in_place` test helper correctly simulates a successful dotenvx run so the new read-back check doesn't falsely trip on all existing mocks. No unrelated code is touched.

No files require special attention. The integration e2e test for `pull --merge` permissions remains gated on the Azure SDK (a pre-existing limitation noted in the previous review), but the unit twin in `test_cli.py` covers the same `_write_merged_combined_file` code path unconditionally.

| Filename | Overview |
|----------|----------|
| src/envdrift/core/partial_encryption.py | Core logic for all four fixes: new error factory methods, `_build_combined_content` helper, `_encrypt_and_verify` with post-encrypt plaintext read-back, `file_has_assignment` guard in both encrypt seams, and both-sources-missing guard in `combine_files`; `secret_lines` is preserved for the secret-var count at line 490. |
| src/envdrift/sync/operations.py | Adds `max_permissions` to `atomic_write`: mode is intersected with the cap so pre-existing world-readable files are tightened on the next write; modes are only ever narrowed, never widened; encoding and newline are already pinned to UTF-8/LF. |
| src/envdrift/cli_commands/sync.py | Single targeted change: `_write_merged_combined_file` replaces bare `write_text` with `atomic_write(max_permissions=0o600)`; import moved to module level (as noted in previous review thread). |
| tests/unit/test_partial_encryption_push_guards.py | New unit test module covering all four guards: both-sources-missing (write=True/False, no-existing-combined), atomic/0600 combined write, empty/comment-only refusal for both seams plus --check, plaintext-survives detection, and pre-existing 0o644 tightening. |
| tests/integration/test_push_false_success_e2e.py | End-to-end regression tests running the real CLI against the real dotenvx binary; covers all four failure modes; POSIX-only permission assertions are correctly gated; pull --merge e2e remains gated on Azure SDK availability (pre-existing concern already noted in a previous review comment). |
| tests/unit/test_partial_encryption.py | Nine pre-existing mocks updated to use `_honest_encrypt_in_place` so they remain valid after the post-encrypt read-back; new `_write_merged_combined_file` 0600+atomic unit test added; `test_push_secrets_only_raises_on_encrypt_failure` moved to the guards module. |
| tests/unit/test_sync_operations.py | Three new `atomic_write` tests: tightens pre-existing 0o644, preserves stricter 0o400, and confirms default behavior is unchanged when no cap is set. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[envdrift push] --> B{Both source files missing?}
    B -- Yes --> ERR1[PartialEncryptionError: sources_missing\nCombined file preserved byte-for-byte]
    B -- No --> C{Secret file has assignments?}
    C -- No --> ERR2[PartialEncryptionError: nothing_to_encrypt\nDotenvx never called]
    C -- Yes --> D[dotenvx.encrypt file]
    D --> E{has_plaintext_secret_value after encrypt?}
    E -- Yes --> ERR3[PartialEncryptionError: encryption_did_not_take_effect\nSkip-worktree protection kept on]
    E -- No --> F[git unskip-worktree]
    F --> G[combine_files]
    G --> H[atomic_write combined_file\nmax_permissions=0o600\nfchmod on fd, atomic rename]
    H --> I[OK: Push complete!]

    J[pull --merge] --> K[_write_merged_combined_file]
    K --> H
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[envdrift push] --> B{Both source files missing?}
    B -- Yes --> ERR1[PartialEncryptionError: sources_missing\nCombined file preserved byte-for-byte]
    B -- No --> C{Secret file has assignments?}
    C -- No --> ERR2[PartialEncryptionError: nothing_to_encrypt\nDotenvx never called]
    C -- Yes --> D[dotenvx.encrypt file]
    D --> E{has_plaintext_secret_value after encrypt?}
    E -- Yes --> ERR3[PartialEncryptionError: encryption_did_not_take_effect\nSkip-worktree protection kept on]
    E -- No --> F[git unskip-worktree]
    F --> G[combine_files]
    G --> H[atomic_write combined_file\nmax_permissions=0o600\nfchmod on fd, atomic rename]
    H --> I[OK: Push complete!]

    J[pull --merge] --> K[_write_merged_combined_file]
    K --> H
```

</a>

1. `src/envdrift/core/partial_encryption.py`, line 136-145 ([link](https://github.com/jainal09/envdrift/blob/56e946b66963a72eb7fa38e2a7b2785de4828465/src/envdrift/core/partial_encryption.py#L136-L145)) 

   <a href="#">P2</a> **Error message says "overwrite" when the combined file may not exist yet**

   When both source files are missing due to a config typo (all three paths are wrong), `combined_file` itself may not exist on disk. The phrase "refusing to overwrite {combined_file} with an empty scaffold" is slightly misleading — there is nothing to overwrite. The test `test_combine_files_refuses_missing_sources_even_without_existing_combined` exercises exactly this path and passes, so it's cosmetic rather than a logic bug, but the message could say "refusing to create or overwrite" for accuracy.

2. `tests/integration/test_partial_encryption_e2e.py`, line 521-522 ([link](https://github.com/jainal09/envdrift/blob/56e946b66963a72eb7fa38e2a7b2785de4828465/tests/integration/test_partial_encryption_e2e.py#L521-L522)) 

   <a href="#">P2</a> **E2e permissions test always skipped without Azure SDK installed**

   `pytest.importorskip("azure.identity")` / `pytest.importorskip("azure.keyvault.secrets")` means this test is silently skipped in any CI/dev environment without the Azure SDK — which is most standard Python pipelines. The guard is likely needed because parsing the `envdrift.toml` with `provider = "azure"` causes an eager Azure import. The unit twin `test_write_merged_combined_file_owner_only_and_atomic` does cover the 0600 fix, but the end-to-end `pull --merge` subprocess path for permissions goes untested in those environments. Consider whether the vault section can be omitted from the TOML while still exercising `pull --merge`.
</details>

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/471-push-fa..."](https://github.com/jainal09/envdrift/commit/2a6d62f19b6f63e91aa98e2ae818d021c14bc38b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37199593)</sub>


docs/cli/pull.md
```diff
@@ -113,7 +113,9 @@ This creates a single usable `.env` file for local development.
 > **Note:** The merged `.env.{env}` file contains decrypted secrets, so it is
 > added to `.gitignore` (just like `envdrift push`) to keep a routine
 > `git add .` from staging plaintext secrets. Commit only the `.clear` and
-> `.secret` files, never the combined file.
+> `.secret` files, never the combined file. The merged file is written
+> atomically with owner-only permissions (`0600` on POSIX, like `.env.keys`),
+> never world-readable at the process umask.
```
jainal09, line 118: 🤖 **envdrift review-bot** · medium · security


**What is wrong.** `atomic_write` deliberately preserves the destination's existing mode when the target already exists as a real file (`src/envdrift/sync/operations.py`):

```python
mode = permissions
with contextlib.suppress(FileNotFoundError):
    dest_stat = path.lstat()
    if not stat.S_ISLNK(dest_stat.st_mode):
        mode = dest_stat.st_mode & 0o777
```

I verified this empirically against the head-SHA file in a /tmp scratch: a fresh write yields `0o600`, but overwriting a pre-existing `0o644` file yields `0o644`.

The population affected by #471(4) is precisely users whose combined/merged files were already created world-readable (0644) by the pre-fix `write_text`. After upgrading, every `envdrift push` (`combine_files` → `atomic_write`) and `envdrift pull --merge` (`_write_merged_combined_file` → `atomic_write`) re-writes that file but **preserves 0644 forever** — including the `pull --merge` output that holds *decrypted* plaintext secrets. The exposure the issue reported is therefore only fixed for fresh files, never remediated for existing ones.

**Doc mismatch.** The docs added in this same PR state the opposite:
- `docs/cli/pull.md:117-118`: "written atomically with owner-only permissions (`0600` on POSIX, like `.env.keys`), **never world-readable at the process umask**" — false for the common overwrite case.
- `docs/cli/push.md:57` and `docs/guides/partial-encryption.md:128` make the same unqualified "owner-only 0600" claim.

**Test gap (acknowledged by the test itself).** Both permission tests only cover the fresh-create path: the unit test `test_write_merged_combined_file_owner_only_and_atomic` writes a new file, and the e2e `test_pull_merge_combined_file_is_owner_only_and_atomic` explicitly does `combined.unlink()` first with the comment "a fresh create is where the 0644-at-umask leak appeared" — sidestepping the overwrite path. A test asserting an existing 0644 merged file is tightened to 0600 after `pull --merge` would FAIL on the fixed code.

**Suggested direction.** For these secret-bearing writers, enforce a permission ceiling rather than preserving a more-permissive pre-existing mode — e.g. an `atomic_write(..., enforce=True)` / clamp (`mode = min(mode, permissions)`-style masking) used by `combine_files` and `_write_merged_combined_file` (the symlink-protection rationale for preservation still holds; only widening beyond `permissions` needs to be refused), plus a regression test that pre-creates the destination at 0644 and asserts 0600 after the write. At minimum, correct the three doc claims and document the upgrade caveat (`chmod 600` existing combined files).

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 56e946b


The fix is correct for freshly created files, but `atomic_write`'s mode-preservation means pre-existing 0644 combined/merged files are never tightened to 0600, contradicting the new docs' "never world-readable" claim — and both new permission tests only exercise the fresh-create path.

**Findings (1):**
- [medium] 0644 combined/merged files created by the pre-fix bug are never remediated; new docs over-promise "never world-readable" (overwrite path also untested) — `docs/cli/pull.md:118` (inline)

docs/cli/pull.md
jainal09: Fixed in 45b64dd — excellent catch on the preserved-mode hole. `atomic_write` now takes `max_permissions`; the two secret-bearing call sites (`combine_files` and `pull --merge`'s `_write_merged_combined_file`) pass `max_permissions=0o600`, so a preserved pre-existing mode is intersected with the cap: combined files left 0644 by the pre-fix `write_text` are tightened to 0600 on the next write. Modes are only narrowed (a stricter 0400 survives), and the no-cap default is unchanged for non-secret writers. Regression trio in test_sync_operations.py (tightens / never-widens / default-preserves, fail-proven pre-fix) plus a combine_files-level tightening test.

src/envdrift/cli_commands/sync.py
```diff
@@ -399,7 +399,12 @@ def _write_merged_combined_file(clear_file: Path, secret_file: Path, combined_fi
             and not line.strip().startswith("DOTENV_PUBLIC_KEY")
         )
 
-    combined_file.write_text("\n".join(combined_lines) + "\n", encoding="utf-8")
+    # The merged file holds DECRYPTED secret values, so write it exactly like
+    # the .env.keys private key: 0600 for a fresh file, fchmod on the fd,
+    # atomic rename — never a bare write_text at the process umask (#471).
+    from envdrift.sync.operations import atomic_write
+
+    atomic_write(combined_file, "\n".join(combined_lines) + "\n")
```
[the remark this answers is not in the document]
jainal09: Done in 45b64dd — moved to a module-level import in cli_commands/sync.py now rather than deferring: you're right that there's no circular-import constraint, and #507's diff doesn't touch the import block, so conflict risk is nil.

src/envdrift/core/partial_encryption.py
```diff
@@ -374,6 +416,16 @@ def combine_files(
...
+    # Refuse to fabricate an empty scaffold when BOTH source files are missing
+    # (deleted by mistake, or a clear_file/secret_file typo in envdrift.toml).
+    # The combined file may be the only remaining copy of the runtime env;
+    # pre-#471 push silently replaced it with a header-only scaffold under the
+    # "[OK] Push complete!" banner. Erroring here is symmetric with
+    # pull-partial's "Secret file not found". A single missing source is still
+    # tolerated (clear-only and secret-only environments are valid).
+    if not clear_file.exists() and not secret_file.exists():
+        raise PartialEncryptionError.sources_missing(clear_file, secret_file, combined_file)
+
```
jainal09, line 464: Addressed in 45b64dd — extracted `_build_combined_content` (warning header + clear + encrypted sections); combine_files drops from 11 to below the threshold.

src/envdrift/core/partial_encryption.py
```diff
@@ -700,6 +773,12 @@ def _encrypt_secrets_only_file(dotenvx: DotenvxWrapper, file: Path, *, check: bo
             # so the encrypted diff is visible to git.
             _git_unskip_worktree(file)
         return False
+    # Refuse a file with no variable assignments BEFORE invoking dotenvx (and
+    # before counting it as would-encrypt in check mode): dotenvx scaffolds
+    # placeholder secrets into an empty file and exits 0 (#443, #471). The same
+    # guard as encrypt_secret_file / the bare-encrypt backend.
+    if not file_has_assignment(file):
+        raise PartialEncryptionError.nothing_to_encrypt(file)
```
jainal09, line 797: Addressed in 45b64dd — extracted the shared `_encrypt_and_verify` helper (dotenvx encrypt + read-back verification); _encrypt_secrets_only_file drops below the threshold.

src/envdrift/core/partial_encryption.py
```diff
@@ -512,6 +568,14 @@ def encrypt_secret_file(env_config: PartialEncryptionEnvironmentConfig) -> None:
...
         return
 
+    # Refuse a file with no variable assignments BEFORE invoking dotenvx:
+    # handed an empty/comment-only file, dotenvx scaffolds its placeholder
+    # template (HELLO, AWS_ACCESS_KEY_ID, OPENAI_API_KEY, ...) and exits 0, so
+    # push would fabricate ~13 secrets the user never wrote and count them as
+    # encrypted. Same guard as the bare-encrypt backend (#443, #471).
+    if not file_has_assignment(secret_file):
+        raise PartialEncryptionError.nothing_to_encrypt(secret_file)
+
```
jainal09, line 608: Addressed in 45b64dd — encrypt_secret_file now delegates its encrypt+verify tail to the shared `_encrypt_and_verify` helper, dropping below the threshold.

tests/integration/test_partial_encryption_e2e.py
```diff
@@ -22,7 +22,9 @@
 
 import os
 import shutil
+import stat
```
jainal09: Addressed in 45b64dd — this PR's four e2e classes moved to tests/integration/test_push_false_success_e2e.py (with private copies of the runner helpers; tests/integration is not a package). The host module returns to its pre-PR size; the stacked #525 diff at the top of the file is untouched.

tests/unit/test_partial_encryption.py
```diff
@@ -187,9 +187,13 @@ def test_encrypt_secret_file_reencrypts_mixed_state(tmp_path: Path):
         name="test", clear_file="", secret_file=str(secret_file), combined_file=""
     )
     with patch("envdrift.core.partial_encryption.DotenvxWrapper") as mock_cls:
+        # Honest mock (#471): a successful encrypt actually encrypts the file,
```
jainal09, line 190: Addressed in 45b64dd — this PR's added unit tests (plus the thematically-matching encrypt-failure sibling) moved to tests/unit/test_partial_encryption_push_guards.py with self-contained fixtures, bringing the host module back to the threshold.

tests/unit/test_partial_encryption.py
```diff
@@ -1432,3 +1454,215 @@ def test_combine_files_handles_non_ascii_under_ascii_locale(tmp_path):
...
+    for raw in path.read_text(encoding="utf-8").splitlines():
+        stripped = raw.strip()
+        key, sep, value = stripped.partition("=")
+        bare = value.strip().strip('"').strip("'")
+        if (
+            sep
+            and key.strip()
+            and not stripped.startswith("#")
+            and bare
+            and not bare.startswith("encrypted:")
```
jainal09: Addressed in 45b64dd — `_honest_encrypt_in_place` rewritten with early-continue guards (comment / non-assignment / already-encrypted), no conditional carries more than two operands now.

jainal09:
@coderabbitai full review

docs/cli/pull.md
jainal09: 🤖 **envdrift review-bot**


Verified resolved at `45b64dd`. The `max_permissions` cap is the right shape: `mode &= max_permissions` after the preserve-existing-mode step means a legacy `0644` combined/merged file is tightened to `0600` on its next write (my exact repro — pinned by `test_atomic_write_max_permissions_tightens_preexisting_world_readable`), a stricter `0400` survives (`never_widens`), and uncapped callers keep preserve semantics (`without_cap_preserves`). Both secret-bearing sites carry the cap — `combine_files` (partial_encryption.py:479) and the `pull --merge` writer (sync.py:407, which holds *decrypted* values). That also makes the docs' "never world-readable at the process umask" claim true for the overwrite case, closing the doc-mismatch half of this finding. Nice fix — closing as resolved.

tests/integration/test_push_false_success_e2e.py
```diff
@@ -0,0 +1,412 @@
...
+    env = os.environ.copy()
+    env["PYTHONPATH"] = integration_pythonpath
+    return subprocess.run(
+        [*envdrift_cmd, *args],
+        cwd=cwd,
+        env=env,
+        capture_output=True,
+        text=True,
+        timeout=timeout,
+    )
```
jainal09, line 51: Fixed in 6aa093e — _run_envdrift dropped the never-overridden timeout param (hardcoded 60), so it is down to 4 args.

tests/integration/test_push_false_success_e2e.py
```diff
@@ -0,0 +1,412 @@
...
+    """Write a combine-mode envdrift.toml for one environment."""
+    (work_dir / "envdrift.toml").write_text(
+        "[partial_encryption]\n"
+        "enabled = true\n\n"
+        "[[partial_encryption.environments]]\n"
+        f'name = "{name}"\n'
+        f'clear_file = "{clear_file}"\n'
+        f'secret_file = "{secret_file}"\n'
+        f'combined_file = "{combined_file}"\n'
+    )
```
jainal09: Fixed in 6aa093e — _write_combine_config now derives the conventional .env.<name>.clear/.secret/combined layout from name alone (2 args); all four call sites simplified to _write_combine_config(work_dir).

jainal09:
@coderabbitai full review

jainal09:
@coderabbitai full review

Issue #486, opened 2026-06-12
parser: quoting/comment/escape/interpolation edge cases diverge from python-dotenv and corrupt results


jainal09:
## Summary

`EnvParser` (the engine behind `validate` and `diff`) diverges from python-dotenv/pydantic-settings semantics in five
edge cases: `${VAR}` interpolation, quote-state tracking in inline-comment stripping, escape decoding in double-quoted
values, Unicode line boundaries, and UTF-8 BOM handling. Each divergence makes `validate --ci` fail configs the real
loader accepts (or pass ones it rejects) and makes `diff` report phantom drift between files that parse identically.
Every item below was reproduced twice (original reviewer + independent verifier) on origin/main commit cb68c97.

## Findings

- [ ] **`validate` fails valid configs that use `${VAR}` interpolation — the real loaders expand it; envdrift
  type-checks the literal.** `src/envdrift/core/parser.py:232`
  **Repro:** `.env` with `OFFSET=1` / `PORT=${OFFSET}234` against `PORT: int`; pydantic-settings loads PORT=1234;
  `envdrift validate .env --schema sc:Settings --ci` → "Expected integer, got '${OFFSET}234'", exit 1.
  **Expected:** expand `${VAR}` per python-dotenv's default interpolation, or recognize the reference and skip/flag it
  distinctly instead of failing the type check on the unexpanded literal.
  **Actual:** false CI FAIL on a config the real Settings class loads; `diff` likewise compares unexpanded literals.
  **Verifier note:** dotenvx also expands `${VAR}` at run time — both ecosystems envdrift bridges disagree with it.

- [ ] **An apostrophe in an unquoted value disables inline-comment stripping for the rest of the line — false CHANGED
  drift and false validate FAILs.** `src/envdrift/core/parser.py:288`
  **Repro:** `MSG=user's data # comment` vs `MSG=user's data`: `dotenv_values` parses both identically;
  `envdrift diff a.env b.env` → CHANGED / "Drift detected"; validate vs `Literal["user's data"]` fails exit 1.
  **Expected:** a quote character mid-token in an unquoted value does not open a quote context (python-dotenv
  semantics); the comment is stripped.
  **Actual:** `_strip_inline_comment` toggles `in_single` on the apostrophe and absorbs ` # comment` into the value;
  any natural-language value (O'Brien, don't) plus an inline comment triggers it.

[...]

PR #529, opened 2026-06-12
fix(parser): parse quoted multiline values like python-dotenv (#458)

### LINKED ISSUE #458: validate/parser: multiline quoted values are split per physical line → truncation + phantom vars + false PASS

jainal09:
## Summary

`EnvParser.parse_string` matched `KEY=value` per **physical** line, so a quoted multiline value
(e.g. a PEM certificate) was **truncated** to its first line — keeping the stray opening quote —
and its interior lines were **re-parsed as phantom variables**. python-dotenv (and therefore
pydantic-settings, validate's ground truth) treats the value as continuing until the matching
close quote.

Consequences fixed here:

- **validate**: a phantom variable could satisfy a required field → **false PASS** (exit 0 even
  with `--ci`); phantom extras also produced false FAILs under `extra="forbid"`.
- **diff**: drift hidden past a value's first line compared equal → **false no-drift**; real drift
  was attributed to phantom variable names.
- **encrypt --check**: phantom variables with sensitive-looking names (e.g. `DB_PASSWORD`) were
  classified as plaintext secrets and **blocked commits** on variables the user never wrote.

## Fix

- `parse_string` now lexes quoted values like python-dotenv: a value opened with `"`/`'` continues
  across physical lines to the matching **unescaped** close quote (escape-aware scan — `\"`/`\'`
  never close), with newlines preserved and CRLF input normalizing to LF exactly like dotenv.
- python-dotenv's quoted-value **escape sets are decoded** (`\\ \' \" \a \b \f \n \r \t \v` in
  double quotes; `\\ \'` in single quotes) so parsed values match pydantic-settings byte-for-byte.
- Unterminated quotes (and a close quote followed by non-comment junk) keep the **legacy per-line
  fallback**, so malformed files parse exactly as before.
- Continuation applies to both strict and lenient (`validate`) parsing; `raw_line` keeps all
  physical lines and line numbers resume after the value.
- Docs updated: `docs/cli/validate.md`, `docs/concepts/how-it-works.md`.

Regression tests were written first and run on the base branch
(`origin/fix/472-validate-pydantic-parity`, 2e573b8):

```text
$ uv run pytest tests/unit/test_parser_multiline.py tests/integration/test_multiline_parsing_cli.py
26 failed, 5 passed   # truncation, phantom vars, false PASS (exit 0 with --ci),
                      # false no-drift, encrypt --check exit 1 naming phantom DB_PASSWORD
```

With the fix (490ac70):

```text
$ uv run pytest tests/unit/test_parser_multiline.py tests/integration/test_multiline_parsing_cli.py
31 passed
$ uv run pytest -m "not integration" -q     # full unit lane
2634 passed, 2 skipped
$ uv run pytest -m integration -q           # full integration lane
333 passed, 13 skipped
```

Coverage: `src/envdrift/core/parser.py` 96% (≥ 80% gate). Tests include in-process
**byte-for-byte parity** with real `dotenv.dotenv_values` and a real pydantic-settings class
loading the same multiline fixture, plus a CRLF (Windows) fixture; PEM-like fixture literals are
built by concatenation. ruff / ruff format / pyrefly / bandit / `make lint-docs` all green.

Fixes #458

**Stacked on #517 — merge that first** (this PR targets `fix/472-validate-pydantic-parity` and
retargets to `main` when #517 merges).

<a href="https://cubic.dev/pr/jainal09/envdrift/pull/529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

This PR fixes `EnvParser.parse_string` so that quoted multiline values (e.g. PEM certificates) are lexed the same way python-dotenv / pydantic-settings lex them, eliminating false PASS/FAIL verdicts in `validate`, hidden drift in `diff`, and phantom plaintext-secret blocks in `encrypt --check`.

- `_continue_quoted_value` extends the raw RHS across physical lines until the matching unescaped close quote is found, with an unterminated-quote or trailing-junk fallback that preserves legacy per-line behaviour for malformed files.
- `_match_quoted_value` + `_decode_escapes` now handle escape decoding for both single-line and multiline quoted values, matching python-dotenv's escape sets exactly; tests include byte-for-byte parity checks against `dotenv.dotenv_values` and a live pydantic-settings class.
- The `raw_line` field is widened to cover all physical lines consumed by a multiline value, and line numbers resume correctly after the value ends.

Safe to merge; the multiline lexing logic is correct and the test suite provides strong byte-for-byte parity coverage against python-dotenv and pydantic-settings.

The core parsing algorithm is logically sound across all edge cases examined — escaped quotes, CRLF normalisation, unterminated-quote fallback, trailing-junk rejection, lenient mode. The three findings are quality concerns rather than correctness defects: _decode_escapes passes a str to a bytes codec via undocumented CPython behaviour, _continue_quoted_value rescans the entire accumulated string on each iteration containing the quote character, and the early-exit-on-trailing-junk invariant is load-bearing but undocumented. None affect results for any input covered by the test suite.

src/envdrift/core/parser.py — specifically _decode_escapes and _continue_quoted_value

| Filename | Overview |
|----------|----------|
| src/envdrift/core/parser.py | Core parser rewritten to lex quoted multiline values like python-dotenv; logic is correct but _decode_escapes relies on undocumented codecs.decode(str, unicode-escape) behaviour, and _continue_quoted_value rescans the entire joined string on each iteration (O(n²)). |
| tests/unit/test_parser_multiline.py | Comprehensive regression suite covering double/single quotes, escape sequences, CRLF normalisation, line-number resumption, raw_line preservation, unterminated-quote fallback, and byte-for-byte parity with python-dotenv and pydantic-settings; no issues found. |
| tests/integration/test_multiline_parsing_cli.py | End-to-end CLI tests for validate, diff, and encrypt --check commands against multiline fixtures; uses subprocess correctly with a deterministic environment fixture; no issues found. |
| docs/cli/validate.md | Documents the new multiline quoting behaviour and escape decoding rules accurately. |
| docs/concepts/how-it-works.md | Minor copy update to step 2 of the validate description; accurate. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant C as Caller
    participant PS as parse_string
    participant CQV as _continue_quoted_value
    participant FCQ as _find_close_quote
    participant VFR as value_from_raw
    participant MQV as _match_quoted_value
    participant DE as _decode_escapes

    C->>PS: parse_string(content)
    loop each physical line
        PS->>PS: regex match → key, raw_value
        PS->>CQV: (raw_value, lines, index)
        CQV->>FCQ: raw_value stripped
        FCQ-->>CQV: None — no close quote on this line
        loop each continuation line
            CQV->>CQV: "joined += newline + next_line"
            alt quote char not in next_line
                CQV->>CQV: continue — skip rescan
            else quote char present
                CQV->>FCQ: lstripped — entire joined string
                FCQ-->>CQV: close position or None
                alt trailing is whitespace or comment
                    CQV-->>PS: joined, consumed
                else junk after close quote
                    CQV-->>PS: raw_value, 0 — legacy fallback
                end
            end
        end
        PS->>VFR: joined_raw
        VFR->>MQV: raw_value.strip()
        MQV->>FCQ: text
        FCQ-->>MQV: close position
        MQV->>DE: inner text between quotes
        DE-->>MQV: decoded value
        MQV-->>VFR: decoded value
        VFR-->>PS: final string value
        PS->>PS: detect encryption, build EnvVar
    end
    PS-->>C: EnvFile
```

<sub>Reviews (1): Last reviewed commit: ["fix(parser): parse quoted multiline valu..."](https://github.com/jainal09/envdrift/commit/490ac70f2677343e15bf77b377544e2b3e5122ce) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37247078)</sub>

> Greptile also left **3 inline comments** on this PR.

### DIFF
--- a/docs/cli/validate.md
+++ b/docs/cli/validate.md
@@ -33,6 +33,17 @@ Type validation mirrors what the real app does at startup:
   and the sensitive-name warning, so encrypted files validate cleanly against an
   `extra="forbid"` schema.
 
+The file itself is parsed with python-dotenv's quoting rules — the parser
+pydantic-settings uses — so the variables envdrift judges are exactly the ones
+the real app loads:
+
+- A value opened with `"` or `'` continues across physical lines until the
+  matching close quote (multiline PEM certificates and keys parse as one
+  value); interior `KEY=value`-looking lines are part of the value, never
+  separate variables.
+- Inside quoted values the python-dotenv escape sequences are decoded (`\n`,
+  `\t`, `\"`, `\\`, ... in double quotes; `\\` and `\'` in single quotes).
+
 ## Arguments
 
 | Argument   | Description                       | Default |
--- a/src/envdrift/core/parser.py
+++ b/src/envdrift/core/parser.py
@@ -7,6 +7,7 @@ Supports:
 
 from __future__ import annotations
 
+import codecs
 import re
 from dataclasses import dataclass, field
 from enum import Enum
@@ -124,11 +125,14 @@ class EnvParser:
     - dotenvx encrypted: KEY="encrypted:xxxx"
     - SOPS encrypted: KEY="ENC[AES256_GCM,data:...,iv:...,tag:...,type:str]"
     - Comments and blank lines (skipped)
-
-    Note:
-        Multiline values are not currently supported. Each line is parsed
-        independently. For multiline secrets (e.g., PEM keys), consider
-        base64 encoding or using a single-line escaped format.
+    - Quoted multiline values (#458): a value opened with ``"`` or ``'``
+      continues across physical lines until the matching unescaped close
+      quote, exactly like python-dotenv (the parser pydantic-settings uses).
+      Newlines inside the value are preserved (normalized to ``\\n``), and the
+      python-dotenv escape sets are decoded inside quoted values
+      (``\\\\ \\' \\" \\a \\b \\f \\n \\r \\t \\v`` in double quotes,
+      ``\\\\ \\'`` in single quotes). An unterminated quote falls back to the
+      legacy per-line treatment so malformed files still parse line by line.
     """
 
     # dotenvx encrypted value pattern
@@ -153,6 +157,21 @@ class EnvParser:
     # .env), keeping the init→validate round-trip intact. Other callers keep the
     # strict pattern so their behaviour is unchanged.
     LENIENT_LINE_PATTERN = re.compile(r"^(?:export\s+)?([^\s=#]+)\s*=(.*)$")
+
+    # python-dotenv's quoted-value escape sets (dotenv/parser.py): decoded
+    # inside double-/single-quoted values so `validate`/`diff` see exactly the
+    # values pydantic-settings loads (#458). The value itself runs to the first
+    # unescaped close quote — across physical lines — found by
+    # ``_find_close_quote`` (an escape-aware scan, equivalent to dotenv's
+    # greedy `(?:\\"|[^"])*` lexing without its backtracking pathologies).
+    DOUBLE_QUOTE_ESCAPES_PATTERN = re.compile(r"\\[\\'\"abfnrtv]")
+    SINGLE_QUOTE_ESCAPES_PATTERN = re.compile(r"\\[\\']")
+
+    # What may legally follow a closing quote on its physical line: optional
+    # whitespace and an optional `#...` comment (python-dotenv's `_comment` +
+    # `_end_of_line`). Anything else means the quote did not cleanly terminate
+    # the assignment, and the legacy raw treatment applies.
+    TRAILING_AFTER_QUOTE_PATTERN = re.compile(r"[^\S\r\n]*(?:#.*)?")
 
     def parse(self, path: Path | str, *, lenient: bool = False) -> EnvFile:
         """
@@ -210,9 +229,12 @@ class EnvParser:
         lines = content.splitlines()
         pattern = self.LENIENT_LINE_PATTERN if lenient else self.LINE_PATTERN
 
-        for line_num, line in enumerate(lines, start=1):
-            original_line = line
-            line = line.strip()
+        index = 0
+        while index < len(lines):
+            line_num = index + 1
+            original_line = lines[index]
+            index += 1
+            line = original_line.strip()
 
             # Skip empty lines
             if not line:
@@ -229,7 +251,19 @@ class EnvParser:
                 continue
 
             key = match.group(1)
-            value = self.value_from_raw(match.group(2))
+            raw_value = match.group(2)
+            raw_lines = [original_line]
+
+            # Quoted multiline continuation (#458): when the RHS opens a quote
+            # that closes on a later physical line, the continuation lines are
+            # part of THIS value — not new assignments or comments.
+            joined_raw, consumed = self._continue_quoted_value(raw_value, lines, index)
+            if consumed:
+                raw_value = joined_raw
+                raw_lines.extend(lines[index : index + consumed])
+                index += consumed
+
+            value = self.value_from_raw(raw_value)
 
             # Determine encryption status and backend
             encryption_status, encryption_backend = self._detect_encryption_status(value)
@@ -239,7 +273,7 @@ class EnvParser:
                 value=value,
                 line_number=line_num,
                 encryption_status=encryption_status,
-                raw_line=original_line,
+                raw_line="\n".join(raw_lines),
                 encryption_backend=encryption_backend,
             )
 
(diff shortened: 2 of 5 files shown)


src/envdrift/core/parser.py
```diff
@@ -248,18 +282,119 @@ def parse_string(self, content: str, *, lenient: bool = False) -> EnvFile:
...
+                continue  # close quote can't be here; skip the re-scan
+            lstripped = joined.lstrip()
+            close = self._find_close_quote(lstripped)
+            if close is None:
+                continue  # the quote(s) on this line were escaped
+            if self.TRAILING_AFTER_QUOTE_PATTERN.fullmatch(lstripped, close + 1):
+                return joined, offset
+            # Closes, but non-comment content trails the quote — not a clean
+            # quoted value; keep the legacy per-line treatment.
+            return raw_value, 0
```
jainal09, line 387: 🤖 **envdrift review-bot** · medium · bug


When a multiline quoted value closes but non-comment content trails the close quote, `_continue_quoted_value` returns `(raw_value, 0)` and the interior physical lines are re-parsed as independent assignments — the exact phantom-var fabrication #458 fixes. python-dotenv behaves differently: its error path consumes the input **through the close-quote line** (the value regex already matched across the lines before `_end_of_line` fails), so the interior lines never become bindings.

Empirically verified against real python-dotenv (repo venv, head parser copied to /tmp):

```text
CERT="-----BEGIN-----
DB_PASSWORD=oops
-----END-----" junk

dotenv_values → {}                               (whole binding dropped)
envdrift head → {'CERT': '"-----BEGIN-----', 'DB_PASSWORD': 'oops'}   (truncated CERT + phantom)
```

Impact: for this malformed shape, every consequence listed in #458 persists — `validate --ci` false-PASSes (the phantom `DB_PASSWORD` satisfies a required field pydantic-settings considers missing), phantom extras false-FAIL under `extra="forbid"`, and `encrypt --check` blocks on a `*_PASSWORD` variable the user never wrote. The new test `test_close_quote_with_trailing_junk_falls_back_to_legacy` even asserts the fallback while its comment claims "either way no multiline value is fabricated" — but phantom *variables* are fabricated; the test only checks `B` and `NEXT`, never an interior assignment-looking line, so this divergence is untested (test gap).

Suggested direction: in the trailing-junk branch, keep whatever value policy you want for the malformed binding (drop it like dotenv, or keep the legacy truncated value), but still **consume the physical lines through the close-quote line** so interior lines can never re-parse as assignments; add a regression test with an interior `DB_PASSWORD=` line plus `" junk` asserting no phantom var (compare against `dotenv_values`). A lesser sibling divergence in the same fallback family: an unterminated quote at EOF keeps `BROKEN='"no closing quote'` while dotenv drops the binding entirely — so a required `BROKEN` also false-PASSes (dotenv does parse subsequent lines as bindings in that case, so per-line fallback for the *interior* lines is correct there; only the kept truncated var diverges).

docs/cli/validate.md
```diff
@@ -33,6 +33,17 @@ Type validation mirrors what the real app does at startup:
   and the sensitive-name warning, so encrypted files validate cleanly against an
   `extra="forbid"` schema.
 
+The file itself is parsed with python-dotenv's quoting rules — the parser
+pydantic-settings uses — so the variables envdrift judges are exactly the ones
+the real app loads:
```
jainal09, line 38: 🤖 **envdrift review-bot** · low · doc-mismatch


The new docs paragraph claims: "The file itself is parsed with python-dotenv's quoting rules — the parser pydantic-settings uses — so the variables envdrift judges are **exactly** the ones the real app loads". That is not true for the legacy fallback paths the PR deliberately keeps (and the parser docstring itself admits): an unterminated quote keeps a truncated `BROKEN='"no closing quote'` variable that python-dotenv drops entirely, and a close quote followed by non-comment junk fabricates variables (truncated value + interior phantom assignments) where python-dotenv produces none — both empirically confirmed against real `dotenv_values`. A user reading this doc would trust `validate --ci` to be byte-equivalent to pydantic-settings; for these malformed shapes it can still false-PASS.

Suggested direction: qualify the claim — e.g. "for well-formed quoted values; malformed quoting (unterminated or junk-trailed quotes) falls back to legacy per-line parsing, which may diverge from python-dotenv" — or fix the divergence (see the parser finding) and keep the strong claim. Keep docs/concepts/how-it-works.md consistent with whichever wording is chosen.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit 490ac70


The multiline-quote parity fix is solid for well-formed values, but the trailing-junk fallback still fabricates phantom variables (the exact #458 bug class) and the new docs overstate python-dotenv parity — both verified empirically against real `dotenv_values`.

**Findings (2):**
- [medium] Trailing-junk-after-close-quote fallback resurrects #458's phantom vars — python-dotenv drops the whole multiline binding — `src/envdrift/core/parser.py:387` (inline)
- [low] docs/cli/validate.md overstates parity: "exactly the ones the real app loads" is false for malformed quote shapes — `docs/cli/validate.md:38` (inline)

src/envdrift/core/parser.py
```diff
@@ -248,18 +282,119 @@ def parse_string(self, content: str, *, lenient: bool = False) -> EnvFile:
...
+        return raw_value, 0  # unterminated: legacy per-line fallback
+
+    @staticmethod
+    def _decode_escapes(escapes: re.Pattern[str], value: str) -> str:
+        """Decode a quoted value's escape sequences like python-dotenv."""
+        return escapes.sub(
+            lambda match: codecs.decode(match.group(0), "unicode-escape"),
+            value,
+        )
```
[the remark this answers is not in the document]
jainal09: Fixed in 357a6ef — `_decode_escapes` now uses an explicit `ESCAPE_DECODE_TABLE` lookup (the table is total for both escape patterns), so there is no `codecs.decode(str, ...)` bytes-codec coercion at all and non-matching text (including non-ASCII) is never re-encoded. Output is byte-identical to python-dotenv's `decode_escapes`: the parity suite plus a 1000-file differential fuzz against python-dotenv 1.2.2 pass with zero mismatches.

src/envdrift/core/parser.py
```diff
@@ -248,18 +282,119 @@ def parse_string(self, content: str, *, lenient: bool = False) -> EnvFile:
...
+            lstripped = joined.lstrip()
+            close = self._find_close_quote(lstripped)
+            if close is None:
+                continue  # the quote(s) on this line were escaped
+            if self.TRAILING_AFTER_QUOTE_PATTERN.fullmatch(lstripped, close + 1):
+                return joined, offset
+            # Closes, but non-comment content trails the quote — not a clean
+            # quoted value; keep the legacy per-line treatment.
+            return raw_value, 0
+        return raw_value, 0  # unterminated: legacy per-line fallback
```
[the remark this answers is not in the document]
jainal09: Fixed in 357a6ef — `_continue_quoted_value` now scans each continuation line exactly once with `_scan_chunk` (a cursor never re-visits earlier lines; the joined string is assembled once at the end), making the lex linear in the input. Per-line scanning is equivalent to the full-string scan because a quote at line start can never be escaped: its predecessor in the joined text is the `\n` joiner (or the opening quote), never a backslash — documented in `_scan_chunk`.

src/envdrift/core/parser.py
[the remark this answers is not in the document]
jainal09: Addressed in 357a6ef by restructuring: the trailing check now runs per physical line inside `_continue_quoted_value` (`chunks[close_chunk]`, offset `close_idx + 1`), so it structurally cannot span lines — no implicit early-return invariant left to relax. For the re-lex in `_match_quoted_value`, the docstring now documents why the fullmatch is newline-safe even on joined multiline text: `[^\S\r\n]` excludes newlines and `#.*` stops at one, and `parse_string` only builds joined values that end on the close-quote line.

jainal09: Fixed in 357a6ef by going all the way to python-dotenv's behavior rather than keeping any legacy value policy: the close quote is now found with dotenv's greedy-regex semantics (the first quote not immediately preceded by a backslash — else the last backslash-preceded one — scanning the remaining input across lines), and a junk-trailed binding is **dropped with input consumed through the close-quote line**, so interior lines can never re-parse as assignments. Unterminated quotes are likewise dropped, consuming only the opening line (dotenv's rest-of-line error recovery), so following lines re-parse exactly like dotenv does.

Your exact repro is a regression test now — `test_trailing_junk_interior_assignment_is_not_a_phantom_var` (asserts no `DB_PASSWORD`, no truncated `CERT`) — plus `dotenv_values` byte-parity params for `trailing-junk-interior-assignment`, `single-line-trailing-junk-dropped`, `multiline-trailing-junk-dropped`, `unterminated-dropped`, and `unterminated-following-lines-reparse`. The formerly false test (`test_close_quote_with_trailing_junk_falls_back_to_legacy`) is flipped to assert the dotenv behavior with the ground truth cited in its docstring.

Battle-tested with a 1000-file differential fuzz against real python-dotenv 1.2.2 (quotes/escapes/multiline/trailing junk/unterminated/non-ASCII/CRLF): zero mismatches in the quoted-value semantics this PR owns; the only residual classes are the pre-existing unquoted-comment rule (owned by #486/#536) and dotenv's quoted-key lexing (filed as #573).

docs/cli/validate.md
jainal09: Fixed in 357a6ef (and the main-merge commit carrying the conflict resolution): both directions of your suggestion were taken. The divergence itself is fixed — malformed quoted bindings (unterminated or junk-trailed) now drop exactly like python-dotenv, verified against `dotenv_values` — so the quoting claim can stay strong, but it is now scoped honestly: the paragraph says "quoted values (well-formed or malformed) are judged exactly as the real app loads them", adds a bullet documenting the malformed-binding drop, and a new paragraph explicitly lists what is NOT yet mirrored (`${VAR}` interpolation, the unquoted-value inline-comment rule, `'quoted keys'` — the first two are #486's scope, the last is tracked in #573). `docs/concepts/how-it-works.md` is updated to the same qualified wording, and `make lint-docs` passes.

Issue #357, opened 2026-06-05
Parser does not strip inline comments → false type errors, false CHANGED diffs, and encrypted values misclassified as plaintext


jainal09:
## Summary
The `.env` parser never strips trailing inline comments, and `_unquote` only removes quotes when the **entire** trimmed value is quote-wrapped. As a result, values with inline comments are mis-parsed, breaking type validation, diffing, and encryption-status detection.

## Location
`src/envdrift/core/parser.py:202-205` (value capture) and `:223-235` (`_unquote`)

## Details
```python
value = match.group(2).strip()   # keeps " # comment"
value = self._unquote(value)     # only strips if the whole value is quote-wrapped
```
Consequences:
- `PORT=8080 # prod port` → value `"8080 # prod port"` → `Validator._check_type` raises a spurious "Expected integer", and a false `CHANGED` row vs `PORT=8080`.
- `API_KEY="encrypted:abc" # rotated` → value stays `"encrypted:abc" # rotated` (not quote-wrapped end-to-end), so `_detect_encryption_status`'s `^encrypted:` check fails and a correctly-encrypted value is classified **PLAINTEXT** → false "unencrypted secret" warning.

Standard `.env` tooling (dotenvx, python-dotenv) strips trailing unquoted `# comment` segments.

## Impact
False validation errors, false drift in `diff`, and false unencrypted-secret warnings for the very common pattern of inline comments.

## Suggested fix
Strip an unquoted trailing `# ...` comment before `_unquote`, mirroring python-dotenv semantics (don't strip a `#` that is inside quotes).

---
*Read-only deep review of `main`.*

PR #536, opened 2026-06-12
fix(parser): match python-dotenv on interpolation, comments, escapes, BOM, line splits (#486)

### LINKED ISSUE #486: parser: quoting/comment/escape/interpolation edge cases diverge from python-dotenv and corrupt results

jainal09:
## Summary

`EnvParser` — the engine behind `validate`, `diff`, and `encrypt --check` — diverged from
python-dotenv / pydantic-settings (the ground truth for what an application actually loads) in five
edge cases, producing false `--ci` verdicts and phantom drift. This PR ships the regression tests
(written first, proven failing on the base branch) plus the fix.

**Stacked on #529 (which stacks on #517) — merge #517, then #529, then this.** Base branch:
`fix/458-parser-multiline`.

## Fixes

- **`${VAR}` interpolation (finding 51):** `${NAME}` / `${NAME:-default}` now expand exactly like
  `dotenv_values` (the loader pydantic-settings wraps): a name defined earlier in the same file wins
  over `os.environ`, an unset name falls back to the default or `""`, resolution is sequential and
  replacement text is never re-expanded. `PORT=${OFFSET}234` validates as `1234`, not the literal.
- **Inline-comment quote state (52):** unquoted values use python-dotenv's whitespace-then-`#` rule;
  a mid-token apostrophe (`MSG=user's data # comment`) no longer opens a quote context and swallows
  the comment. The quote-aware scan is kept for the malformed-quote legacy fallback (#458 semantics),
  now pinned by dedicated tests.
- **Double-quoted escapes (54):** already fixed on the base branch by #529's quoted-value lexer —
  verified and pinned here with the issue's exact repros (`MSG="tab\there"` decodes;
  `TOKEN="abc\ndef"` is 7 chars so `min_length=8` rejects it) at unit and CLI level.
- **Unicode line boundaries (57):** content splits on `\n` / `\r\n` / `\r` only (python-dotenv's
  `_end_of_line`); U+2028, form feed, NEL, vertical tab, ... stay inside the value instead of
  truncating it and fabricating phantom variables.
- **UTF-8 BOM (96):** files are read with `utf-8-sig` and `parse_string` strips a leading BOM, so
  the first key is never an invisible `﻿NAME` phantom (validate listed the same variable as
  both MISSING and EXTRA; diff reported phantom drift against the BOM-less twin). Note: python-dotenv
  itself keeps the BOM in the key — stripping it is the issue's chosen direction, and the deliberate
  divergence is documented and pinned in `test_bom_divergence_from_python_dotenv_is_deliberate`.

Docs updated in the same PR: `docs/concepts/how-it-works.md` and `docs/cli/validate.md` parsing
semantics (markdownlint clean).

On the **base branch** (`origin/fix/458-parser-multiline` @ 490ac70):

- `tests/unit/test_parser_edge_cases.py`: **35 failed, 10 passed** (the passes are the item-54
  escape pins plus preserved behaviors).
- `tests/integration/test_parser_edge_cases_cli.py` (real CLI subprocess): **9 failed, 2 passed**
  (the 2 passes are the item-54 escape repros, already fixed by #529's lexer).
- Base failures reproduce the issue verbatim: `Expected integer, got '${OFFSET}234'`,
  `EXTRA VARIABLES: SECRET_PASSWORD` from the U+2028 line, `API_KEY` listed under both
  MISSING REQUIRED and EXTRA (the BOM renders invisibly), CHANGED drift between dotenv-identical
  twins.

After the fix (this branch):

- 49 unit + 11 CLI regression tests pass, including byte-for-byte parity with the real
  `dotenv.dotenv_values` (13 parametrized fixtures) and real pydantic-settings classes loading the
  same files in-process; BOM and U+2028 fixtures are written as exact bytes; subprocess tests assert
  exit codes on normalized output.
- Full lanes green: 2683 non-integration + 659 scanner + parsing-relevant integration files
  (45 + 60 tests, incl. all 31 #458 multiline tests). `ruff check` / `ruff format --check` /
  `pyrefly check` (0 errors) / `bandit` (0 findings) clean. `src/envdrift/core/parser.py` at 98%
  line coverage on the parser suites.

Fixes #486

<a href="https://cubic.dev/pr/jainal09/envdrift/pull/536?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

This PR fixes five parser edge cases where `EnvParser` diverged from python-dotenv / pydantic-settings, causing false `--ci` verdicts and phantom drift. The core parser (`parser.py`) gains `${VAR}`/`${VAR:-default}` interpolation via `_interpolate`, a new `_parse_unquoted_value` that matches python-dotenv's whitespace-then-`#` comment rule, Unicode line-boundary splitting restricted to `\n`/`\r\n`/`\r` only, and UTF-8 BOM stripping with a recorded flag for downstream warnings.

- **`parser.py`**: Adds `_interpolate` (sequential `${VAR}` expansion with `os.environ` fallback), `_parse_unquoted_value` (replaces the old quote-aware scan for unquoted RHS), `_strip_leading_bom`, `LINE_BOUNDARY_PATTERN`, `POSIX_VARIABLE_PATTERN`, and `UNQUOTED_COMMENT_PATTERN`; `value_from_raw` is updated to route unquoted values through the new method.
- **`validator.py`**: Emits a warning when `env_file.leading_bom` is set, surfacing the pydantic-settings runtime mismatch instead of silently passing.
- **Tests**: 49 unit + 11 CLI regression tests covering all five edge cases, byte-for-byte parity assertions against real `dotenv_values`, and in-process pydantic-settings class loading.

The changes are narrowly scoped to the parser value-extraction path and validated by byte-exact parity tests against the real python-dotenv and pydantic-settings loaders.

Each of the five bug fixes has a dedicated regression test proven failing on the base branch and passing after the fix. The interpolation, unquoted-value lexer, BOM stripping, and line-boundary changes are all correct and well-tested.

| Filename | Overview |
|----------|----------|
| src/envdrift/core/parser.py | Core change: adds interpolation, unquoted-value lexer, BOM stripping, and unicode-safe line splitting. Logic is correct and well-guarded. |
| src/envdrift/core/validator.py | Adds a BOM warning block that fires when `env_file.leading_bom` is set; no logic changes to the existing validation flow. |
| tests/unit/test_parser_edge_cases.py | 552-line regression suite covering all five edge cases with parametrized parity tests against real `dotenv_values`. |
| tests/integration/test_parser_edge_cases_cli.py | End-to-end CLI subprocess tests for each edge case with precise exit-code and JSON output assertions. |
| tests/unit/test_parser.py | Existing tests updated to reflect corrected `K= # c` semantics and PLAINTEXT status for leading-hash values. |

<sub>Reviews (4): Last reviewed commit: ["Merge origin/main into fix/486-parser-ed..."](https://github.com/jainal09/envdrift/commit/11b45654dc0c2108eb367fd819d54ff4f0bd163f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37315011)</sub>


src/envdrift/core/parser.py
```diff
@@ -395,21 +430,52 @@ def _decode_escapes(escapes: re.Pattern[str], value: str) -> str:
...
         # Fast path: the overwhelming majority of values contain no `#`, so skip
-        # the per-character quote-tracking scan entirely for them.
+        # the scan entirely for them.
         if "#" not in value:
             return value
 
+        stripped = value.lstrip()
+        if not stripped or stripped[0] not in "\"'":
+            # Unquoted value: python-dotenv's `parse_unquoted_value` rule.
+            return self.UNQUOTED_COMMENT_PATTERN.sub("", value)
```
jainal09: 🤖 **envdrift review-bot** · medium · bug


The new unquoted path applies `UNQUOTED_COMMENT_PATTERN = re.compile(r"\s+#.*")` to the RAW value, which (per the `LINE_PATTERN` design comment, parser.py:156-158) deliberately includes the whitespace right after `=`. python-dotenv does the opposite: its binding parser consumes inline whitespace after `=` (`_whitespace = [^\S\r\n]*`) BEFORE lexing the value, then applies `re.sub(r"\s+#.*", "", part)` inside `parse_unquoted_value` — so a `#` that begins the value after leading spaces is value content, never a comment.

Empirical proof (python-dotenv 1.2.1, the repo venv):

```
DIFF 'K= # c\n'
     envdrift={'K': ''}
     dotenv  ={'K': '# c'}
```

pydantic-settings (which wraps `dotenv_values`) therefore loads `K="# c"` while envdrift head reports `K` as `EMPTY` — exactly the false-verdict class this PR claims to eliminate (validate can fail/required-empty a config the app loads non-empty; diff reports drift between `K= #x` and `K=#x` twins that the real loader sees as different... actually loads as '#x' vs ' #x'-stripped — verdicts diverge either way; encrypt --check classifies it EMPTY and skips it).

Three compounding problems:
1. The new docstring (parser.py:456-458, "python-dotenv strips from the first whitespace-preceded `#` unconditionally") and the pattern comment (parser.py:194-196, "python-dotenv's `parse_unquoted_value` rule") overstate parity — the implementation does NOT match `parse_unquoted_value` for this shape.
2. docs/cli/validate.md line 13 ("An inline `# comment` (preceded by whitespace) is stripped from unquoted values...") and docs/concepts/how-it-works.md ("envdrift sees exactly the variables pydantic-settings would load") are now inaccurate for `K= # c`.
3. The 13-fixture `TestPythonDotenvParity` corpus omits this shape, so the byte-for-byte parity suite cannot catch it.

Note the conflict: existing test tests/unit/test_parser.py:347 (`EMPTY=  # just a comment` → `""`, from #357) pins the current divergent behavior, so this needs an explicit decision: either align with python-dotenv (flip the #357 pin, apply the comment rule only after consuming post-`=` whitespace like python-dotenv does) or carve out the documented exception in the docstrings/docs and add the case to the parity corpus as a pinned, deliberate divergence (like the BOM one). Right now the code, the docstring, and the docs disagree with each other.

src/envdrift/core/parser.py
```diff
@@ -202,8 +225,11 @@ def parse(self, path: Path | str, *, lenient: bool = False) -> EnvFile:
 
         # A binary / non-UTF-8 file raised a raw UnicodeDecodeError traceback;
         # convert it to a clean ValueError with an actionable message (#24).
+        # utf-8-sig: a UTF-8 BOM is an encoding artifact (Notepad/PowerShell),
+        # not part of the first key — keeping it produced an invisible phantom
+        # `\ufeffNAME` variable and self-contradictory MISSING+EXTRA output (#486).
         try:
-            content = path.read_text(encoding="utf-8")
+            content = path.read_text(encoding="utf-8-sig")
```
jainal09: 🤖 **envdrift review-bot** · medium · bug


The PR reads files with `utf-8-sig` and strips `﻿` in `parse_string`, so a BOM-prefixed `API_KEY=abc123` parses as `API_KEY`. But the primary ground truth this PR defines ("the value the app receives") does the opposite: pydantic-settings reads env files with `dotenv_values(file_path, encoding=encoding or 'utf8')` — NOT utf-8-sig — so the real loader sees key `﻿API_KEY` and a required `API_KEY` field FAILS at app startup. Empirically (pydantic-settings 2.12.0):

```
pydantic-settings FAILED: ValidationError 1 validation error for Settings
API_KEY  Field required [type=missing, input_value={'﻿api_key': 'abc123'}, ...]
envdrift parses: {'API_KEY': 'abc123'}
```

Yet the new integration test `test_validate_bom_file_passes_and_is_intelligible` asserts exit 0 + "Validation PASSED" on exactly this file. That is a false GREEN on a config that crashes the app — the highest-impact failure class for this tool, and the one verdict change issue #486 explicitly warned against: the verifier note says "pydantic-settings also chokes on the BOM key, so the FAILED verdict matches ground truth — the defect is the unintelligible output and phantom drift, not the verdict." The PR fixed the unintelligible MISSING+EXTRA output (good) but ALSO flipped the verdict (beyond what the issue asked), and docs/concepts/how-it-works.md still claims "envdrift sees exactly the variables pydantic-settings would load."

The divergence is at least documented and pinned (`test_bom_divergence_from_python_dotenv_is_deliberate`), so this is a deliberate trade-off — but it trades a cosmetic defect for a silent false PASS. Suggested direction: keep the BOM-stripped parse for intelligible naming and drift comparison, but have `validate` emit a warning/failure when a leading BOM is detected (the file will not load correctly under pydantic-settings until the BOM is removed or env_file_encoding='utf-8-sig' is set), or at minimum correct the how-it-works.md claim and document that BOM files validate-pass while pydantic-settings rejects them.

src/envdrift/core/parser.py
```diff
@@ -395,21 +430,52 @@ def _decode_escapes(escapes: re.Pattern[str], value: str) -> str:
...
+        default or ``""``, and replacement text is never re-expanded.
+        """
+
+        def resolve(match: re.Match[str]) -> str:
+            name = match.group("name")
+            var = parsed.get(name)
+            if var is not None:
+                return var.value
+            default = match.group("default")
+            return os.environ.get(name, default if default is not None else "")
```
jainal09: 🤖 **envdrift review-bot** · low · bug


`_interpolate`'s fallback is `os.environ.get(name, ...)`. On Windows, `os._Environ` upper-cases keys on both store and lookup, so `${path}` / `${Path}` resolve to the PATH value. python-dotenv instead materializes a plain dict (`env.update(os.environ)` in `resolve_variables`) whose keys are the upper-cased names, then does a case-SENSITIVE `env.get(self.name, default)` — so `${path}` misses and resolves to the default/empty string. Result: on Windows (which this repo supports — see the `_utf8_subprocess_on_windows` conftest fixture), a lowercase/mixed-case `${name}` reference expands in envdrift but not in pydantic-settings, re-introducing a verdict divergence in the exact category this PR fixes. The parity test corpus only uses uppercase names, so it cannot catch this, and CI parity tests on POSIX runners behave identically either way.

Suggested direction: snapshot `os.environ` into a plain dict for the lookup (e.g. `dict(os.environ).get(name, ...)` or an equivalent cached copy) to match python-dotenv's resolution exactly, and/or add a mixed-case `${name}` fixture to TestPythonDotenvParity.

jainal09:
🤖 **envdrift review-bot** — automated re-review of commit db348b3


The parser-parity work is solid overall, but three verified divergences remain between what envdrift parses and what pydantic-settings/python-dotenv actually load — two of them produce false verdicts in exactly the class this PR sets out to eliminate.

**Findings (3):**
- [medium] Leading-whitespace-then-# after '=' still diverges from python-dotenv: envdrift parses EMPTY where pydantic-settings loads the '#...' text as the value — `src/envdrift/core/parser.py:477` (inline)
- [medium] BOM strip flips validate --ci to a false PASS: pydantic-settings raises ValidationError on the same file envdrift now reports 'Validation PASSED' — `src/envdrift/core/parser.py:232` (inline)
- [low] _interpolate os.environ fallback diverges from python-dotenv on Windows: case-insensitive lookup vs python-dotenv's case-sensitive dict — `src/envdrift/core/parser.py:449` (inline)

All findings were anchored inline on the diff; see the individual comments for evidence and suggested directions.

src/envdrift/core/parser.py
jainal09: Fixed in 0fb26fd — the unquoted path now mirrors python-dotenv exactly. The post-`=` whitespace is consumed first (dotenv's `_equal_sign` lexer, `=[^\S\r\n]*`), then the `\s+#.*` comment rule + `rstrip` run, in the new `_parse_unquoted_value`.

Ground truth (python-dotenv 1.2.2, repo venv):

```python
>>> dotenv_values(path_to("K= # c\nEMPTY=  # just a comment\n"))
{'K': '# c', 'EMPTY': '# just a comment'}
```

envdrift now parses these byte-identically (and classifies them PLAINTEXT, not EMPTY). The #357 pins in `tests/unit/test_parser.py` are flipped with the dotenv evidence in their docstrings, `TestUnquotedLeadingCommentValue` pins the shapes (`K= # c`, `EMPTY=  # just a comment`, `K= \t # c # d` → `'# c'`, `K= v # c` → `'v'`), and the `TestPythonDotenvParity` corpus gained `leading-hash-after-equals-is-the-value` and `post-equals-whitespace-comment` fixtures. Docs (`docs/cli/validate.md`, `docs/concepts/how-it-works.md`) now describe the real rule, closing the code/docstring/docs disagreement. Issue #537, which tracked this divergence, is fixed by this commit. An 80k-case differential fuzz vs real `dotenv_values` over interpolation/comment/escape/BOM/line-split shapes reports 0 mismatches.

src/envdrift/core/parser.py
jainal09: Fixed in 0fb26fd, taking the suggested direction: the BOM-stripped parse is kept for intelligible naming and drift comparison, but the divergence is no longer silent — `parse_string` records the stripped BOM on a new `EnvFile.leading_bom` field and `validate` emits a warning ("File starts with a UTF-8 BOM: pydantic-settings reads .env files as plain UTF-8 … Remove the BOM or set env_file_encoding='utf-8-sig' on the model config").

The verdict deliberately stays PASS-with-warning rather than FAIL, because a blanket FAIL would itself diverge from ground truth: the BOM only corrupts the *first* binding, so a BOM file whose first line is a comment or blank loads fine in pydantic-settings (python-dotenv 1.2.2: `"﻿# header\nAPI_KEY=x\n"` → `{'﻿': None, 'API_KEY': 'x'}` — API_KEY intact). The warning names the exact risk without inventing a startup failure the app may not have.

Tests: `test_validate_bom_file_passes_with_a_bom_warning` (CLI, asserts exit code and the warning text), `test_leading_bom_is_recorded_on_the_env_file`, `test_validator_warns_on_leading_bom`, and the deliberate-divergence pin now also asserts the flag. Both doc claims are corrected — `how-it-works.md` names this as the one deliberate divergence from "exactly what pydantic-settings loads".

src/envdrift/core/parser.py
jainal09: Fixed in 0fb26fd: `_interpolate` now resolves the fallback through a `dict(os.environ)` snapshot with a case-sensitive `.get`, exactly like python-dotenv's `resolve_variables` (`env.update(os.environ)` then case-sensitive lookup).

Verified on real Windows (PowerShell host, Python 3.14.4, python-dotenv 1.2.2):

- `os.environ.get('path')` hits (case-insensitive `os._Environ`), while `dict(os.environ).get('path')` misses — confirming the divergence you predicted;
- `dotenv_values` resolves `A=${path:-fell-back}` to `'fell-back'` and `B=${PATH:-fell-back}` to the real PATH value;
- the fixed parser matches dotenv byte-for-byte on both, whereas the previous `os.environ.get` fallback expanded `${path}`.

Added `test_environ_lookup_is_case_sensitive` (meaningful on the Windows CI runners: the lowercase ref must fall back) and the `interpolation-mixed-case-name` fixture to the parity corpus.

src/envdrift/core/parser.py
```diff
@@ -265,6 +294,12 @@ def parse_string(self, content: str, *, lenient: bool = False) -> EnvFile:
...
             value = self.value_from_raw(raw_value)
 
+            # ${VAR} interpolation (#486): expand exactly like dotenv_values —
+            # the loader pydantic-settings uses — so validate/diff judge the
+            # value the application receives, not the unexpanded literal.
+            if "${" in value:
+                value = self._interpolate(value, env_file.variables)
+
             # Determine encryption status and backend
             encryption_status, encryption_backend = self._detect_encryption_status(value)
```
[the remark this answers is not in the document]
jainal09: Documented and pinned in 0fb26fd, keeping the current classification deliberately. `B=${A}` with `A` encrypted resolves to the ciphertext string and stays ENCRYPTED: the file stores only the reference, so there is no plaintext secret on disk for `encrypt --check` to protect, and classifying the alias PLAINTEXT would flip `is_fully_encrypted` to false and make `encrypt`/`lock`/the guardian agent re-encrypt the literal `${A}` on every idle cycle — an encrypt-churn loop worse than the skip. The trade-off is now spelled out in a comment at the `_detect_encryption_status` call site, and `test_alias_of_encrypted_value_inherits_encrypted_status` pins the semantics (value, status and backend) so they cannot drift silently.

src/envdrift/core/parser.py
```diff
@@ -173,6 +181,21 @@ class EnvParser:
...
 
+    # .env line boundaries are `\n` / `\r\n` / `\r` ONLY (python-dotenv's
+    # `_end_of_line`). `str.splitlines()` also severed values at the wider
+    # Unicode line-boundary set (U+2028, form feed, NEL, ...), truncating the
+    # value and fabricating phantom variables from its tail (#486).
+    LINE_BOUNDARY_PATTERN = re.compile(r"\r\n|\r|\n")
+
+    # python-dotenv's `${name}` / `${name:-default}` interpolation pattern
+    # (dotenv/variables.py `_posix_variable`), expanded by `_interpolate` (#486).
+    POSIX_VARIABLE_PATTERN = re.compile(r"\$\{(?P<name>[^\}:]*)(?::-(?P<default>[^\}]*))?\}")
```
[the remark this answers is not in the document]
jainal09: Fixed in 0fb26fd: `POSIX_VARIABLE_PATTERN` now uses the plain character classes (`[^}:]`, `[^}]`).

src/envdrift/core/parser.py
```diff
@@ -190,6 +210,23 @@ class EnvParser:
     # the assignment, and the legacy raw treatment applies.
     TRAILING_AFTER_QUOTE_PATTERN = re.compile(r"[^\S\r\n]*(?:#.*)?")
 
+    # .env line boundaries are `\n` / `\r\n` / `\r` ONLY (python-dotenv's
+    # `_end_of_line`). `str.splitlines()` also severed values at the wider
+    # Unicode line-boundary set (U+2028, form feed, NEL, ...), truncating the
+    # value and fabricating phantom variables from its tail (#486).
+    LINE_BOUNDARY_PATTERN = re.compile(r"\r\n|\r|\n")
```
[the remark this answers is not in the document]
jainal09: Verified — not reproducible for `encrypt --check`. That command parses through `EnvParser.parse(..., lenient=True)` + `EncryptionDetector.analyze` (`src/envdrift/cli_commands/encryption.py`, `_analyze_file`), which uses the fixed `LINE_BOUNDARY_PATTERN` split — no `splitlines()` on that path. The PR's real-CLI regression test `test_encrypt_check_sees_no_phantom_secret` drives `envdrift encrypt --check` as a subprocess on exactly this input (`NOTE=part1<U+2028>SECRET_PASSWORD=hunter2`) and passes: exit 0, no `SECRET_PASSWORD` in the output.

The remaining `splitlines()` sites are in other subsystems (partial-encryption push/lock file reconstruction, sync ops, scanners) with deliberately line-based semantics of their own. I checked the closest one — `partial_encryption.has_plaintext_secret_value` — with this input: it returns the same file-level boolean with or without the U+2028 split (`_line_has_plaintext_secret` is True for the whole unsplit line too), and it never fabricates variable names. No behavior change is attributable to the parser's line-boundary fix there.

src/envdrift/cli_commands/encryption.py
```diff
@@ -195,6 +229,13 @@ def encrypt_cmd(
             print_error(f"ENV file not found: {env_file}")
         raise typer.Exit(code=1)
 
+    # Refuse companion targets (.keys/.example/.sample/.template) across ALL
+    # requested files — the default single [.env] and every multi-file target —
+    # before touching any of them, so one companion in a batch aborts the whole
+    # invocation with nothing encrypted (#474).
+    for env_file in files:
+        _refuse_companion_target(env_file, "encrypt")
```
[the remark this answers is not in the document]
jainal09: Valid finding, wrong PR: the flagged hunk (`for env_file in files: _refuse_companion_target(env_file, "encrypt")`) is from #509 (`c6b9f11`, already merged to `main`) and only appears in this PR's diff because the branch merged `origin/main` while the base branch hasn't caught up yet — this parser PR does not touch `encryption.py`.

I reproduced it against the real CLI (`encrypt --check .env .env.example` → exit 1, refusal error, no report for `.env`) and filed #579 with the repro, impact analysis (the generated default hook regex `^\.env\.(production|staging)$` doesn't match `.env.example`, so only widened patterns/manual batches are affected), and the suggested `if not check:` scoping — so it gets fixed in its own subsystem rather than smuggled into a parser PR.

src/envdrift/core/parser.py
```diff
@@ -243,7 +282,15 @@ def parse_string(self, content: str, *, lenient: bool = False) -> EnvFile:
...
-        lines = content.splitlines()
+        # A leading UTF-8 BOM is an encoding artifact (Notepad/PowerShell), not
+        # part of the first key — keeping it produced an invisible phantom
+        # ``U+FEFF``-prefixed ``NAME`` variable and self-contradictory MISSING+EXTRA output.
+        # Strip it, but record it: pydantic-settings reads plain UTF-8 and
+        # WOULD see the ``U+FEFF``-prefixed key, so `validate` warns (#486).
+        if content.startswith("\ufeff"):
+            content = content.removeprefix("\ufeff")
+            env_file.leading_bom = True
+        lines = self.LINE_BOUNDARY_PATTERN.split(content)
```
jainal09: Fixed in 19c2feb: the BOM-recording branch is extracted into `_strip_leading_bom` (documented helper, behavior identical), bringing `parse_string` back under the threshold.

src/envdrift/core/parser.py
```diff
@@ -474,15 +555,45 @@ def _decode_escapes(self, escapes: re.Pattern[str], value: str) -> str:
...
-        whitespace context is intact.
+        Legacy quote-aware scan, reachable only through ``value_from_raw``'s
+        malformed-quoted-binding fallback (direct callers): ``parse_string``
+        drops those bindings like python-dotenv, and unquoted values go
+        through ``_parse_unquoted_value`` instead. A `#` starts a comment only
+        when it is outside quotes AND preceded by whitespace; a `#` inside
+        matching quotes, glued to a token, or escaped (`\\#`, and an escaped
+        quote `\\"` that must not toggle quote state) is preserved. Call this
+        on the raw `value` (before stripping) so the whitespace context is
+        intact.
```
jainal09, line 608: Declining this one with rationale: `_strip_inline_comment`'s complexity is the pre-existing quote-state machine kept verbatim from the base branch — this PR only rerouted `parse_string`'s unquoted values away from it (to the simpler dotenv-exact `_parse_unquoted_value`), so the method is now a legacy fallback reachable only through the public `value_from_raw` for direct callers. Its per-character scan (escape handling, single/double-quote state, whitespace-preceded `#`) is irreducible without splitting the state machine across helpers that each carry the state — more obscure, not simpler. The diff makes it less load-bearing, not more complex.

src/envdrift/cli_commands/encryption.py
```diff
@@ -62,6 +62,40 @@ def _resolve_config_path(config_path: Path | None, value: Path | str | None) ->
     return path
 
 
+def _refuse_companion_target(env_file: Path, action: str) -> None:
```
jainal09, line 65: This delta is not from this PR: the file is untouched by the parser changes here and only entered the diff because the branch merged `origin/main` (which contains #509, `c6b9f11`) while the base branch hasn't caught up. The flag will drop out once the base syncs with main / the PR retargets. The underlying #509 behavior nit that cubic found in the same hunk is tracked in #579.

src/envdrift/cli_commands/encryption.py
```diff
@@ -195,6 +229,13 @@ def encrypt_cmd(
             print_error(f"ENV file not found: {env_file}")
         raise typer.Exit(code=1)
 
+    # Refuse companion targets (.keys/.example/.sample/.template) across ALL
+    # requested files — the default single [.env] and every multi-file target —
+    # before touching any of them, so one companion in a batch aborts the whole
+    # invocation with nothing encrypted (#474).
+    for env_file in files:
+        _refuse_companion_target(env_file, "encrypt")
+
```
jainal09, line 252: This delta is not from this PR: the file is untouched by the parser changes here and only entered the diff because the branch merged `origin/main` (which contains #509, `c6b9f11`) while the base branch hasn't caught up. The flag will drop out once the base syncs with main / the PR retargets. The underlying #509 behavior nit that cubic found in the same hunk is tracked in #579.

src/envdrift/cli_commands/encryption.py
```diff
@@ -631,6 +672,8 @@ def decrypt_cmd(
         print_error(f"ENV file not found: {env_file}")
         raise typer.Exit(code=1)
 
+    _refuse_companion_target(env_file, "decrypt")
+
```
jainal09, line 711: This delta is not from this PR: the file is untouched by the parser changes here and only entered the diff because the branch merged `origin/main` (which contains #509, `c6b9f11`) while the base branch hasn't caught up. The flag will drop out once the base syncs with main / the PR retargets. The underlying #509 behavior nit that cubic found in the same hunk is tracked in #579.

src/envdrift/encryption/dotenvx.py
```diff
@@ -112,6 +112,27 @@ def encrypt(
...
+                success=False,
+                message=(
+                    f"Refusing to encrypt {env_file}: it is a companion file "
+                    "(.keys/.example/.sample/.template), not an env file. "
+                    "Encrypting the dotenvx private-key store would permanently "
+                    "lock out every file encrypted with its keys."
+                ),
+                file_path=env_file,
+            )
+
```
jainal09, line 135: This delta is not from this PR: the file is untouched by the parser changes here and only entered the diff because the branch merged `origin/main` (which contains #509, `c6b9f11`) while the base branch hasn't caught up. The flag will drop out once the base syncs with main / the PR retargets. The underlying #509 behavior nit that cubic found in the same hunk is tracked in #579.

src/envdrift/core/validator.py
```diff
@@ -125,6 +125,20 @@ def validate(
...
+        # Surface that loudly instead of a silent false PASS (#486 review).
+        if env_file.leading_bom:
+            result.warnings.append(
+                "File starts with a UTF-8 BOM: pydantic-settings reads .env files "
+                "as plain UTF-8, so the app will see the first key with an "
+                "invisible '\\ufeff' prefix and a required field backed by it "
+                "will come up missing at startup. Remove the BOM or set "
+                "env_file_encoding='utf-8-sig' on the model config."
+            )
+
```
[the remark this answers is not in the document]
jainal09: Declining with rationale: the +1 is a single guard clause (`if env_file.leading_bom: result.warnings.append(...)`) implementing the warning requested by the review-bot BOM thread on this PR. `Validator.validate` was already at complexity 65; a real decomposition of that method is a standalone refactor far outside this stacked parser PR's scope, and per campaign precedent oversized "brain methods" are tracked separately rather than opportunistically refactored inside unrelated fix PRs.

Issue #441, opened 2026-06-09
Dogfood sweep (phase 3): 32 multi-agent-verified usability/correctness findings


jainal09:
## Dogfood findings — multi-agent sweep (phase 3)

A parallel multi-agent dogfood (42 agents, 9 surfaces, each finding adversarially verified by an independent agent) of envdrift surfaces not covered by the manual first-run/collaboration passes. **32 confirmed findings**: 3 high, 15 medium, 13 low, 1 polish. Pick off via small themed PRs.

### HIGH

- **encrypt on an empty .env silently injects ~15 fake template secrets into the user's file with a bare [OK]/exit 0**
  - _surface:_ Error/edge paths across core commands · _command:_ `envdrift encrypt empty.env`
  - _fix:_ /Users/jainal09/envdrift-df/src/envdrift/cli_commands/encryption.py:305 (guard the `else` encrypt branch, between the parse at line 242 and the encrypt call at line 305): when `not env.variables` / `len(env) == 0`, warn "empty.env has no variables, nothing to encrypt" and return 
- **lock --verify-vault --force reports "Encrypted: 1" but leaves the file in PLAINTEXT (data-loss / commit-secrets footgun)**
  - _surface:_ HashiCorp Vault backend · _command:_ `envdrift lock --verify-vault --force   (with a local .env.keys whose DOTENV_PRIVATE_KEY_<ENV> does not match the public `
  - _fix:_ src/envdrift/cli_commands/sync.py:1752-1759 (verify is_encrypted_content() on the file after encrypt before incrementing encrypted_count / printing the green line); root cause also in src/envdrift/integrations/dotenvx.py:701-707 (add "undefined" / "MISPAIRED_PRIVATE_KEY" to ENCRY
- **Native scanner silently skips untracked files not named .env/.env.* — secrets pass with exit 0**
  - _surface:_ secret scanner edge cases · _command:_ `envdrift guard --native-only   (in a git repo with an untracked 'secrets.env' or 'config.txt' containing a real GitHub t`
  - _fix:_ /Users/jainal09/envdrift-df/src/envdrift/scanner/native.py:406-419 (Method 2 in _collect_files): the untracked-file pathspec `_ENV_FILE_PATHSPEC` (native.py:241) + `_is_env_file` gate (native.py:418) name-restricts untracked discovery. Untracked, non-.env-named text files contain

### MEDIUM

[...]

PR #591, opened 2026-07-09
fix(hook): make encryption checks complete

### LINKED ISSUE #579: encrypt --check aborts the whole batch when a companion file (.env.example) is among the arguments
## Summary

Since #509, `encrypt_cmd` refuses companion targets (`.keys`/`.example`/`.sample`/`.template`)
unconditionally, BEFORE the `--check` branch is reached
(`src/envdrift/cli_commands/encryption.py:233-236`, the `for env_file in files:
_refuse_companion_target(env_file, "encrypt")` loop). `--check` is non-destructive, so the
lockout-prevention rationale of #509 does not apply to it.

## Repro (main, real CLI)

```console
$ printf 'API_KEY=plain\n' > .env
$ printf 'API_KEY=\n' > .env.example
$ envdrift encrypt --check .env .env.example
[ERROR] Refusing to encrypt .env.example: it is a plaintext companion file
(.example/.sample/.template), not a secret store.
$ echo $?
1
```

The invocation exits 1 with **no encryption report for `.env`** — one companion in the batch
aborts the whole check.

## Impact

- A pre-commit hook using `envdrift encrypt --check` with `pass_filenames: true` and a `files`
  pattern wider than the generated default (`^\.env\.(production|staging)$`) fails the commit as
  soon as `.env.example` is staged, without checking the real env files in the batch.
- Manual `envdrift encrypt --check .env*` globs hit the same abort.

The default generated hook regex does not match `.env.example`, so out-of-the-box hooks are not
affected — this bites widened patterns and manual batches.

## Suggested fix

Scope the refusal to the destructive path (`if not check: _refuse_companion_target(...)`), or in
check mode report the companion as skipped instead of aborting. Found by cubic-dev-ai on PR #536
(the hunk is #509 code that entered that PR's diff via a base-sync merge of main).

### LINKED ISSUE #580: install_hooks creates a missing explicit config_path even when create_if_missing=False
## Bug

`install_hooks()` violates its documented contract when the caller passes an **explicit** `config_path` that does not exist together with `create_if_missing=False`: instead of raising `FileNotFoundError`, it silently creates the file and returns `True`.

## Evidence

`src/envdrift/integrations/precommit.py` (main @ a7bf222):

- The docstring promises: `Raises: FileNotFoundError: If config not found and create_if_missing=False` (~line 342).
- The `create_if_missing` guard is only applied on the **auto-detect** path (`config_path is None`, lines ~351-358).
- The explicit-path branch creates unconditionally (lines ~360-362):

```python
if not config_path.exists():
    config_path.write_text(_fresh_config_text(_active_hook_ids()), encoding="utf-8")
    return True
```

So a flow that explicitly requested "do not create" gets a file created behind its back.

`tests/unit/test_precommit.py::test_raises_when_config_not_found_and_no_create` only covers the auto-detect path (`install_hooks(create_if_missing=False)` with no `config_path`), which is why this slipped through.

Introduced with the #512 hook rework (`cb4344c`). Found via a cubic review thread on PR #530 (raised in the retarget window; PR #530 does not touch `precommit.py`).

## Suggested fix

Guard the missing-file branch:

```python
if not config_path.exists():
    if not create_if_missing:
        raise FileNotFoundError(
            f"Pre-commit config not found: {config_path}. "
            "Run from repository root or specify --config path."
        )
    config_path.write_text(_fresh_config_text(_active_hook_ids()), encoding="utf-8")
    return True
```

An `xfail(strict=False)` regression test asserting the correct behavior is being added on PR #530 per the repo convention; flip it to a plain assertion in the fixing PR.

### LINKED ISSUE #583: encrypt --check misses plaintext secret shadowed by a later duplicate encrypted assignment
## Bug

`envdrift encrypt --check` decides `should_block_commit` from the parsed `EnvFile`
(`src/envdrift/cli_commands/encryption.py:319` — `parser.parse(env_file, lenient=True)` then
`detector.analyze(env, ...)` at `:326`). `EnvParser` collapses duplicate assignments
last-wins (python-dotenv parity, #536), so a file like:

```dotenv
SECRET_KEY=plaintext-value-123
SECRET_KEY="encrypted:BDqDBJ24bUq3x"
```

parses to only `{'SECRET_KEY': 'encrypted:BDqDBJ24bUq3x'}` (verified on main @ c42a357 with
`EnvParser().parse(...)`). The earlier **plaintext** line is invisible to
`EncryptionDetector.should_block_commit`, so `encrypt --check` (and the pre-commit hook built on
it) passes a file that still contains a plaintext secret on disk — dotenvx-style append-encrypt
workflows produce exactly this duplicate shape.

## Fix direction

The parser's last-wins semantics are correct (python-dotenv parity); the gap is that `--check`
should scan raw assignments (or all duplicate occurrences), not the collapsed map — e.g. have the
detector consider every `KEY=value` line, or have the parser expose duplicate assignments for the
check path.

## Provenance

Flagged by cubic P1 on PR #526 (retarget-window review; the code is main's, not #526's delta —
`git diff main...fix/478-guard-exit-truthfulness` does not touch `encryption.py`). Filed per the
fix-or-file convention.

jainal09:
## Summary

Make encryption checks safe and complete when they run from a pre-commit filename batch.

## Changes

- Check every accepted assignment, so a later encrypted duplicate cannot hide an earlier plaintext secret.
- Skip plaintext companion files during `encrypt --check`, while continuing to check other requested secret stores.
- Honor `create_if_missing=False` for an explicit missing pre-commit config path.
- Keep two confirmed scanner-output regressions (#441) and the git-secrets clean-repo regression (#581) runnable as non-strict expected failures until their dedicated fixes land.

## Related issues

Closes #579
Closes #580
Closes #583

Tracks #441 and #581.

## Testing

- [x] `uv run pytest -m "not integration" -q` — 3313 passed, 5 skipped, 2 xfailed, 1 xpassed
- [x] `uv run pytest -m integration -q` with pinned dotenvx 2.1.5 and LocalStack/Vault/Lowkey — 496 passed, 37 skipped, 3 xfailed
- [x] `make lint-docs`
- [x] `uv run ruff check src tests`, `uv run ruff format --check .`, `uv run pyrefly check`, and `uv run bandit -r src -c pyproject.toml`
- [x] CLI dogfood for mixed `.env.production`/`.env.example` input and duplicate plaintext/ciphertext assignments

## Checklist

- [x] New/changed behavior has a **real** regression test (containers/binaries/CLI, not mocks)
- [x] Confirmed-but-unfixed bugs are `xfail(strict=False)` with an issue ref
- [x] Touched files keep **≥ 80%** coverage (CodeCov CI green)
- [x] No hardcoded binary versions — read from `src/envdrift/constants.json` (Renovate-managed)
- [x] `ruff check` / `ruff format` / `pyrefly` / `bandit` clean
- [x] Docs (`docs/`) updated for behavior change
- [x] Any pre-existing bug found was fixed or filed as an issue
- [x] No `FORCE_COLOR`-dependent integration output; no `importlib.reload()` of package modules

<h3>Greptile Summary</h3>

This PR tightens encryption checks for pre-commit batches and duplicate assignments. The main changes are:

- Skip plaintext companion files during `encrypt --check` while still checking secret stores.
- Track accepted duplicate assignments so hidden plaintext secrets are checked.
- Respect `create_if_missing=False` for explicit missing pre-commit config paths.
- Keep known scanner-output regressions runnable as non-strict expected failures.
- Update docs and tests for the new check behavior.

This looks safe to merge.

| Filename | Overview |
|----------|----------|
| src/envdrift/cli_commands/encryption.py | Adds companion-file skipping for check mode and enables duplicate-assignment analysis for encryption reports. |
| src/envdrift/core/encryption.py | Adds an opt-in path for checking every accepted assignment instead of only last-assignment-wins values. |
| src/envdrift/core/parser.py | Stores accepted assignments in parse order while preserving the existing variable lookup behavior. |
| src/envdrift/integrations/precommit.py | Routes missing config creation through the `create_if_missing` flag. |

<sub>Reviews (4): Last reviewed commit: ["fix(hook): make encryption checks comple..."](https://github.com/jainal09/envdrift/commit/efcb44e05f72983f66274393085cb1903874ceaf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43115601)</sub>

### DIFF
--- a/src/envdrift/cli_commands/encryption.py
+++ b/src/envdrift/cli_commands/encryption.py
@@ -8,12 +8,13 @@ Supports multiple encryption backends:
 from __future__ import annotations
 
 import shlex
+from collections.abc import Callable
 from pathlib import Path
 from typing import Annotated
 
 import typer
 
-from envdrift.core.encryption import EncryptionDetector
+from envdrift.core.encryption import EncryptionDetector, EncryptionReport
 from envdrift.core.parser import EnvParser
 from envdrift.core.schema import SchemaLoader, SchemaLoadError
 from envdrift.encryption import EncryptionProvider, get_encryption_backend
@@ -99,6 +100,46 @@ def _refuse_companion_target(env_file: Path, action: str) -> None:
             "file (.example/.sample/.template), not a secret store."
         )
     raise typer.Exit(code=1)
+
+
+def _is_companion_target(env_file: Path) -> bool:
+    """Return whether ``env_file`` is a plaintext companion, by filename."""
+    from envdrift.env_files import _is_excluded_env_file
+
+    return _is_excluded_env_file(env_file.name)
+
+
+def _refuse_companion_targets(
+    files: list[Path],
+    action: str,
+    *,
+    allow_plaintext_companions: bool,
+) -> None:
+    """Refuse companion files unless this is a non-destructive check batch."""
+    if allow_plaintext_companions:
+        return
+    for env_file in files:
+        _refuse_companion_target(env_file, action)
+
+
+def _check_files_and_report(
+    files: list[Path],
+    analyze_file: Callable[[Path], EncryptionReport],
+    detector: EncryptionDetector,
+) -> bool:
+    """Print check reports and return whether any non-companion file blocks a commit."""
+    should_block = False
+    for env_file in files:
+        if _is_companion_target(env_file):
+            print_warning(
+                f"Skipping {env_file}: it is a plaintext companion file, not a secret store."
+            )
+            continue
+        report = analyze_file(env_file)
+        print_encryption_report(report)
+        if detector.should_block_commit(report):
+            should_block = True
+    return should_block
 
 
 def _protect_private_keys(env_file: Path) -> None:
@@ -248,12 +289,14 @@ def encrypt_cmd(
             print_error(f"ENV file not found: {env_file}")
         raise typer.Exit(code=1)
 
-    # Refuse companion targets (.keys/.example/.sample/.template) across ALL
-    # requested files — the default single [.env] and every multi-file target —
-    # before touching any of them, so one companion in a batch aborts the whole
-    # invocation with nothing encrypted (#474).
-    for env_file in files:
-        _refuse_companion_target(env_file, "encrypt")
+    # A check invocation can receive a broad pre-commit filename batch. It
+    # permits plaintext companions while the check helper skips them, whereas
+    # a destructive invocation refuses every companion target (#579).
+    _refuse_companion_targets(
+        files,
+        "encrypt",
+        allow_plaintext_companions=check,
+    )
 
     if verify_vault or vault_provider or vault_url or vault_region or vault_secret:
         print_error("Vault verification moved to `envdrift decrypt --verify-vault ...`")
@@ -315,7 +358,7 @@ def encrypt_cmd(
     parser = EnvParser()
     detector = EncryptionDetector()
 
-    def _analyze_file(env_file: Path):
+    def _analyze_file(env_file: Path) -> EncryptionReport:
         """Parse and analyze one env file, exiting cleanly on unreadable input."""
         try:
             # lenient=True so --check analyzes every KEY=value assignment dotenvx
@@ -328,7 +371,11 @@ def encrypt_cmd(
             print_error(str(e))
             raise typer.Exit(code=1) from None
 
-        report = detector.analyze(env, schema_meta)
+        report = detector.analyze(
+            env,
+            schema_meta,
+            include_overridden_assignments=check,
+        )
         detected_backend = detector.detect_backend_for_file(env_file)
         if detected_backend:
             report.detected_backend = detected_backend
@@ -337,15 +384,7 @@ def encrypt_cmd(
         return report
 
     if check:
-        # Report status per file; block if any file would block a commit.
-        should_block = False
-        for env_file in files:
-            report = _analyze_file(env_file)
-            print_encryption_report(report)
-            if detector.should_block_commit(report):
-                should_block = True
-
-        if should_block:
+        if _check_files_and_report(files, _analyze_file, detector):
             raise typer.Exit(code=1)
     else:
         # Analyze every target before touching any backend: unreadable input and
--- a/tests/integration/test_scanner_git_tools.py
+++ b/tests/integration/test_scanner_git_tools.py
@@ -258,6 +258,10 @@ class TestGitSecretsRealBinary:
         assert len(result.findings) >= 1, "git-secrets finding (emitted on stderr) was not captured"
         assert any("aws" in f.rule_id for f in result.findings)
 
+    @pytest.mark.xfail(
+        strict=False,
+        reason="git-secrets scans its registered Bedrock pattern in .git/config (see #581)",
+    )
     def test_scan_clean_repo_has_no_findings(self, scanner_test_env):
         """scan() of a repo without secrets yields no findings and no error."""
         work_dir = scanner_test_env["work_dir"]
--- a/tests/scanner/test_output.py
+++ b/tests/scanner/test_output.py
@@ -445,6 +445,10 @@ class TestRichOutput:
         assert "aws-access-key-id" in out  # Rule kept for CI triage
         assert "AKIA1234" in out  # Preview kept
 
+    @pytest.mark.xfail(
+        strict=False,
+        reason="format_rich ignores the supplied wide Console under pytest capture (see #441)",
+    )
     def test_format_rich_wide_terminal_shows_all_columns(self):
         """Wide terminal shows the Rule and Preview columns too."""
         console = Console(record=True, force_terminal=True, width=140)
@@ -474,6 +478,10 @@ class TestRichOutput:
         format_rich(self._aws_finding_result(), console)
         assert "AKIA1234" not in console.export_text()
 
+    @pytest.mark.xfail(
+        strict=False,
+        reason="format_rich ignores the supplied wide Console under pytest capture (see #441)",
+    )
     def test_format_rich_threshold_100_is_wide(self):
         """Width 100 (exactly the threshold) shows the secondary Preview column."""
         console = Console(record=True, force_terminal=True, width=100)
--- a/docs/cli/encrypt.md
+++ b/docs/cli/encrypt.md
@@ -29,7 +29,7 @@ exit code (dotenvx can exit `0` without encrypting):
 - **Refuses content-free files.** An empty, blank-line-only, or comment-only file
   has no variables to encrypt, so the command declines with a non-zero exit
   instead of letting dotenvx scaffold a placeholder-secrets template into it.
-- **Refuses the key store and companion files.** `envdrift encrypt .env.keys`
+- **Refuses key stores and companion files when encrypting.** `envdrift encrypt .env.keys`
   would encrypt the dotenvx private-key store itself — the keys become
   ciphertext under a brand-new keypair whose private half is never saved,
   permanently locking out every encrypted file in the project — so the command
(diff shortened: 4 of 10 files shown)


src/envdrift/cli_commands/encryption.py
```diff
@@ -340,6 +351,11 @@ def _analyze_file(env_file: Path):
         # Report status per file; block if any file would block a commit.
         should_block = False
         for env_file in files:
+            if _is_companion_target(env_file):
+                print_warning(
+                    f"Skipping {env_file}: it is a plaintext companion file, not a secret store."
+                )
+                continue
```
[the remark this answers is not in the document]
jainal09: Thanks — this behavior is intentional and is the requested scope of #579. The project-wide canonical classification treats `.example`, `.sample`, and `.template` as non-secret plaintext companions; the existing scanner defaults also exclude them. `encrypt --check` enforces encryption for actual secret stores, not a general secret-scanning policy. Blocking an all-companion batch would restore #579s custom `pass_filenames` lockout even for conventional placeholders. This PR adds a real CLI regression test for the explicit skip contract and documents it. A real secret in a file deliberately named as a plaintext template violates that companion-file contract and needs a dedicated scanner/policy, rather than encryption enforcement on a file that must remain plaintext.

src/envdrift/cli_commands/encryption.py
```diff
@@ -101,6 +102,46 @@ def _refuse_companion_target(env_file: Path, action: str) -> None:
...
+    analyze_file: Callable[[Path], EncryptionReport],
+    detector: EncryptionDetector,
+) -> bool:
+    """Print check reports and return whether any non-companion file blocks a commit."""
+    should_block = False
+    for env_file in files:
+        if _is_companion_target(env_file):
+            print_warning(
+                f"Skipping {env_file}: it is a plaintext companion file, not a secret store."
+            )
```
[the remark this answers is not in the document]
jainal09: Thanks — this behavior is intentional and is the requested scope of #579. The project-wide canonical classification treats `.example`, `.sample`, and `.template` as non-secret plaintext companions; the existing scanner defaults also exclude them. `encrypt --check` enforces encryption for actual secret stores, not a general secret-scanning policy. Blocking an all-companion batch would restore #579s custom `pass_filenames` lockout even for conventional placeholders. This PR adds a real CLI regression test for the explicit skip contract and documents it. A real secret in a file deliberately named as a plaintext template violates that companion-file contract and needs a dedicated scanner/policy, rather than encryption enforcement on a file that must remain plaintext.

tests/scanner/test_output.py
```diff
@@ -445,6 +445,10 @@ def test_format_rich_non_interactive_keeps_all_columns(self):
         assert "aws-access-key-id" in out  # Rule kept for CI triage
         assert "AKIA1234" in out  # Preview kept
 
+    @pytest.mark.xfail(
```
jainal09, line 448: **Blocking:** this test passes reliably — I ran it 3× isolated on `main`, plus `test_output.py` (48 passed) and the whole `tests/scanner/` dir (745 passed), and it **xpasses** on this branch. It's not failing, so `xfail(strict=False)` here doesn't tolerate a known failure — it *disables a working regression test*: if `format_rich` ever regresses and stops emitting the wide columns, this would fail-as-expected and CI would stay green, hiding it. Please remove the xfail. If it genuinely fails in your local terminal, prefer a targeted `pytest.mark.skipif(<that condition>)` (or fix #441) so CI keeps guarding the real behavior.

tests/scanner/test_output.py
```diff
@@ -474,6 +478,10 @@ def test_format_rich_threshold_99_is_narrow(self):
         format_rich(self._aws_finding_result(), console)
         assert "AKIA1234" not in console.export_text()
 
+    @pytest.mark.xfail(
```
jainal09, line 481: **Blocking (same as above):** `test_format_rich_threshold_100_is_wide` also passes reliably on `main` and xpasses here. Marking it `xfail(strict=False)` masks a future regression at exactly the width-100 threshold boundary this test exists to guard. Please drop the xfail.

tests/integration/test_scanner_git_tools.py
```diff
@@ -258,6 +258,10 @@ def test_scan_detects_aws_key_via_stderr(self, scanner_test_env):
         assert len(result.findings) >= 1, "git-secrets finding (emitted on stderr) was not captured"
         assert any("aws" in f.rule_id for f in result.findings)
 
+    @pytest.mark.xfail(
+        strict=False,
+        reason="git-secrets scans its registered Bedrock pattern in .git/config (see #581)",
```
jainal09, line 263: Non-blocking note: this `#581` xfail is more defensible than the #441 ones — the clean-repo scan genuinely fails in some environments (git-secrets matching its own registered Bedrock pattern in `.git/config`), so there's a real failing manifestation being tracked. Just flagging that `strict=False` here still swallows a *CI* regression if the scanner truly breaks; if the failure is environment-specific (it passes in CI), a `skipif` scoped to that environment would keep CI coverage live while tolerating the local artifact. Your call — leaving it as-is is acceptable since #581 is real and filed.

jainal09 (commented):
Thorough review — I checked this out and dogfooded it, not just read the diff. **The three fixes are genuinely good and verified:**

- **#583 (duplicate hidden by ciphertext):** dogfooded end-to-end — `encrypt --check` on `SECRET_KEY=plaintext` shadowed by `SECRET_KEY="encrypted:..."` now reports `Plaintext: 1` / `PLAINTEXT SECRETS DETECTED: SECRET_KEY` and exits 1. Confirmed the regression test is load-bearing: reverting the `analyze()`/parser change makes it fail. Nice — this closes a real silent-secret hole.
- **#579 (companion skip in `--check`):** correct; a companion no longer aborts a pre-commit filename batch while real stores are still checked.
- **#580 (`create_if_missing=False`):** correct and fails loud — the default stays `True` (CLI unchanged) and the new `FileNotFoundError` is caught by the CLI's `except (PrecommitConfigError, OSError)` → clean `print_error` + exit 1, not a traceback. Good that the previously-xfailed test is flipped to a real assertion.

Gates are clean (ruff, pyrefly 0, 3313 unit passed).

**Blocking:** the two `#441` xfails on `tests/scanner/test_output.py` disable *passing* regression tests (details inline). I verified these two tests pass reliably on `main` — 3/3 isolated runs, `test_output.py` 48 passed, the whole `tests/scanner/` dir 745 passed — and they **xpass** on this branch. Marking a green test `xfail(strict=False)` silently swallows a future real regression in `format_rich`'s wide-column output. Please drop these two xfails (the tests already pass). If they fail in a specific local terminal, that's worth an inline `skipif` on that condition or a real `#441` fix — not disabling CI coverage that currently works.

One softer note on the `#581` xfail (non-blocking) inline.

tests/scanner/test_output.py
jainal09: Still blocking on the new head `a70e9bc`: the refactor is a nice cleanup but this xfail is untouched, and the test still **XPASS**es here (`-rX` confirms `XPASS test_format_rich_wide_terminal_shows_all_columns`). Please delete this `@pytest.mark.xfail` — the test passes, so keeping it as an expected-failure disables real coverage. Happy to unblock the moment it's an active assertion again.

tests/scanner/test_output.py
jainal09: Same on the new head — `XPASS test_format_rich_threshold_100_is_wide`. Drop the xfail and this thread resolves.

tests/scanner/test_output.py
jainal09: Reproduced on rebased head `efcb44e`: `uv run pytest tests/scanner/test_output.py --runxfail -q` fails both tests. The captured failure shows the supplied width-140 Console becomes `<console width=80 None>`, so the Rule/Preview columns are absent. With normal pytest they are genuine xfails, not xpasses. Retaining the non-strict xfail is required by the repository convention while #441 remains open.

tests/scanner/test_output.py
jainal09: Reproduced on rebased head `efcb44e`: `uv run pytest tests/scanner/test_output.py --runxfail -q` fails both tests. The captured failure shows the supplied width-100 Console becomes `<console width=80 None>`, so the Preview column is absent. With normal pytest they are genuine xfails, not xpasses. Retaining the non-strict xfail is required by the repository convention while #441 remains open.

PR #594, opened 2026-07-10
test(scanner): isolate Rich width tests from host terminals

jainal09:
## Summary

Make the active Rich wide-terminal layout assertions independent of the host terminals capture state.

## Changes

- Keep `format_rich` on Richs public `console.width` behavior.
- Construct fixed-width interactive test consoles with Richs controlled environment, so their configured width remains deterministic.
- Preserve the active wide, narrow, and threshold-100 assertions; no test is skipped or xfailed.

## Related issues

Part of #441; follows #593s re-enabling of the active width assertions.

## Testing

- [x] `uv run pytest tests/scanner/test_output.py --no-cov -q` — 48 passed
- [x] `uv run pytest -m "not integration" -q` — 3315 passed, 5 skipped, 1 xpassed
- [x] `uv run ruff check src tests`, `uv run ruff format --check .`, `uv run pyrefly check`, and `uv run bandit -r src -c pyproject.toml`

## Checklist

- [x] Active regression assertions cover the fixed-width behavior
- [x] No passing test was skipped or marked xfail
- [ ] Touched-file coverage confirmation pending CodeCov CI
- [x] No production behavior or binary version changed
- [x] Static checks are clean
- [x] No documentation change is needed for test-environment isolation

<h3>Greptile Summary</h3>

This PR updates Rich output tests for configured console width handling.

- Adds a fixed-width interactive console helper for scanner output tests.
- Uses the helper in narrow, wide, and threshold layout assertions.
- Keeps coverage for fallback and non-interactive table behavior.

This looks safe to merge.

| Filename | Overview |
|----------|----------|
| tests/scanner/test_output.py | Adds a fixed-width interactive Rich console helper and applies it to width-sensitive scanner output tests. |

<sub>Reviews (2): Last reviewed commit: ["test(scanner): isolate Rich width tests ..."](https://github.com/jainal09/envdrift/commit/fe9c512c80ca80ebc8cd77b0f1d358facee5a174) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43209951)</sub>

### DIFF
--- a/tests/scanner/test_output.py
+++ b/tests/scanner/test_output.py
@@ -22,6 +22,11 @@ from envdrift.scanner.output import (
 )
 
 
+def _interactive_console(width: int) -> Console:
+    """Build a fixed-width interactive Console independent of host terminal state."""
+    return Console(record=True, force_terminal=True, width=width, _environ={})
+
+
 class TestJsonOutput:
     """Tests for JSON output formatter."""
 
@@ -421,7 +426,7 @@ class TestRichOutput:
         """Interactive narrow terminal: keep Sev/Location/Description (one line
         each, ellipsized) and drop the Rule/Preview columns.
         """
-        console = Console(record=True, force_terminal=True, width=80)
+        console = _interactive_console(80)
         format_rich(self._aws_finding_result(), console)
         out = console.export_text()
 
@@ -447,7 +452,7 @@ class TestRichOutput:
 
     def test_format_rich_wide_terminal_shows_all_columns(self):
         """Wide terminal shows the Rule and Preview columns too."""
-        console = Console(record=True, force_terminal=True, width=140)
+        console = _interactive_console(140)
         format_rich(self._aws_finding_result(), console)
         out = console.export_text()
 
@@ -470,13 +475,13 @@ class TestRichOutput:
 
     def test_format_rich_threshold_99_is_narrow(self):
         """Width 99 (one below the threshold) drops the secondary Preview column."""
-        console = Console(record=True, force_terminal=True, width=99)
+        console = _interactive_console(99)
         format_rich(self._aws_finding_result(), console)
         assert "AKIA1234" not in console.export_text()
 
     def test_format_rich_threshold_100_is_wide(self):
         """Width 100 (exactly the threshold) shows the secondary Preview column."""
-        console = Console(record=True, force_terminal=True, width=100)
+        console = _interactive_console(100)
         format_rich(self._aws_finding_result(), console)
         assert "AKIA1234" in console.export_text()
 


src/envdrift/scanner/output.py
```diff
@@ -221,7 +235,7 @@ def format_rich(result: AggregatedScanResult, console: Console | None = None) ->
         _build_findings_table(
             result,
             interactive=console.is_terminal,
-            wide=console.width >= _WIDE_TERMINAL_WIDTH,
+            wide=_configured_console_width(console) >= _WIDE_TERMINAL_WIDTH,
```
jainal09: **This is a no-op — `console.width` already honors a configured width.**

I dogfooded on the branch (Rich 14.2.0). `_configured_console_width(console)` returns the identical value to `console.width` in *every* scenario:

| construction | `console.width` | `_configured_console_width` | differ? |
|---|---|---|---|
| `width=100, force_terminal, record` | 100 | 100 | no |
| `width=120, force_terminal` | 120 | 120 | no |
| `width=99, force_terminal, record` | 99 | 99 | no |
| no width (probe) | 80 | 80 | no |
| `width=100` **with `COLUMNS=40`/`LINES=10` env set** | 100 | 100 | no |

Rich's `Console.size` returns `self._width` whenever it's set, regardless of the terminal probe or `COLUMNS`/`LINES` — so `console.width` never "resolves from the terminal even if the caller configured `width`." That premise doesn't hold on this Rich version.

**Revert experiment (proof it's not load-bearing):** I changed this line back to `wide=console.width >= _WIDE_TERMINAL_WIDTH` and re-ran the width suite — `4 passed`. Reverting the core change breaks nothing.

**The new test doesn't guard the claim.** `test_configured_console_width_falls_back_to_terminal_when_unset` only asserts the fallback branch (`_configured_console_width(console) == console.width` for an *unconfigured* console). That passes even if the helper were literally `return console.width`. Nothing exercises the actual claimed behavior — "prefer the configured `_width` over the terminal probe" — because that behavior is indistinguishable from `console.width`.

**And it's fragile:** `getattr(console, "_width", None)` reaches into a Rich private attribute. If Rich renames it, the helper silently falls back to `console.width` — the same value — so no test would ever catch the "regression."

**#441 is already resolved.** The two width tests were un-xfailed in #593 and pass reliably at width 100/99/120. There's no latent width bug in `format_rich` to fix.

**To resolve, pick one:**
1. Add a test that **fails on `main`** (without this change) — a concrete Console construction where `console.width` does *not* honor the configured width. If you can produce that, the fix is justified and *that* test is the real regression guard. I'll happily merge.
2. Otherwise close #594 as "no code change needed — #441 was resolved by re-enabling the tests in #593," so we don't ship a no-op that depends on Rich internals.

jainal09 (commented):
Reviewer block — dogfooded this change against the branch at e25f818a (Rich 14.2.0, the pinned version). See the inline comment on the call site for the evidence. Short version: this change is a **provable no-op**, its new test doesn't guard the claimed behavior, and it adds a private-attribute dependency for zero behavioral gain. Please either show a test that fails *without* this change, or close it as "#441 already resolved by #593". I'm the merge authority here — I'll merge once this is resolved.

src/envdrift/scanner/output.py
jainal09: Agreed. Revision `fe9c512` removes the production helper and restores `format_rich` to `console.width`. The only remaining delta gives the fixed-width interactive test Consoles a controlled Rich environment (`_environ={}`), which removes host capture state from the test fixture while preserving the active wide/narrow/threshold assertions. `tests/scanner/test_output.py` passes 48/48 and the full non-integration suite passes 3315 tests. Please verify this revised, test-only change and resolve the thread if satisfied.

jainal09: **Resolved — verified on `fe9c512c`.** You took option 2 and then some:

- The no-op `_configured_console_width` helper and the trivial fallback test are **gone**; `src/envdrift/scanner/output.py` is now byte-identical to `main` (call site back to `console.width`). ✅
- PR retyped `fix:` → `test:`, so the squash won't cut a spurious release. ✅
- Remaining change is a clean test-hygiene refactor: `_interactive_console(width)` with `_environ={}` to pin the width tests independent of host terminal env.

Dogfooded on the branch (Rich 14.2.0): `tests/scanner/test_output.py` → **48 passed**; ruff/format/pyrefly all clean.

One honest note for the record: I probed `format_rich` with `COLUMNS=40`/`LINES=10` in the host env and the output is identical with or without `_environ={}` (`_width` already wins over `COLUMNS`), so the isolation is **belt-and-suspenders, not strictly load-bearing today**. That's fine for a `test:` change — it's the idiomatic Rich test seam, DRYs up 5 tests, and guards future tests that might not pin `width=` explicitly. Accepting it as hygiene, not a bug fix. Merging on green CI.

Issue #573, opened 2026-07-06
Parser: python-dotenv quoted-key ('KEY'=v) and bare-key (KEY without =) bindings are not mirrored


jainal09:
Found during the differential-fuzz hardening of #458/#529 (1000 generated .env files vs real python-dotenv 1.2.2): the only residual variable-set divergences outside #486's scope trace to python-dotenv's KEY lexing, which `EnvParser` does not implement.

## Evidence (python-dotenv 1.2.2 ground truth)

1. **Bare keys**: a line with no `=` is a binding with value `None` in dotenv, skipped entirely by envdrift.

   ```text
   plain            ->  dotenv_values: {'plain': None}
                        EnvParser:     {}
   ```

2. **Quoted keys**: dotenv's `_single_quoted_key = r"'([^']+)'"` accepts `'MY KEY'=v`:

   ```text
   'MY KEY'=v       ->  dotenv_values: {'MY KEY': 'v'}
                        EnvParser:     {} (strict) / {"'MY": ...} (lenient)
   ```

3. **Worst case — the key regex spans newlines**, so a stray line starting with `'` swallows the following lines into a phantom key and drops bindings the app would otherwise (not) see:

   ```text
   '
   K=1
   '
   B=2              ->  dotenv_values: {'\nK=1\n': None, 'B': '2'}
                        EnvParser:     {'K': '1', 'B': '2'}
   ```

   In fuzzing this shape arose naturally when the close-quote line of a dropped malformed binding began with `'` — 3/1000 files diverged this way.

## Where

- `src/envdrift/core/parser.py` — `LINE_PATTERN` / `LENIENT_LINE_PATTERN` require `KEY=...` on one physical line; there is no quoted-key or bare-key path.

## Impact

`validate`/`diff` can disagree with what pydantic-settings actually loads for files using these (rare) shapes: a bare key is a real `None`-valued env entry to pydantic-settings, and a quoted key is a real variable. Low severity, but it bounds the parser-parity claim (docs already qualify it as of #529's branch).

Discovered while working on PR #529 (fix/458-parser-multiline, commit 357a6ef).

jainal09:
[...]

PR #622, opened 2026-07-11
fix(parser): address key lexer review findings

jainal09:
## Summary

Follow up the review findings that were still open when #620 was reviewer-merged. This refactors the key lexer below the repository's complexity threshold, proves malformed bare-looking lines do not shadow the process environment, and restores ordinary `KEY=value` parsing to its pre-#620 performance.

## Changes

- split quoted/unquoted key tokenization, tail classification, and quoted-value close location into small helpers
- reduce every parser method to McCabe complexity 9 or lower without suppressions
- restore the established regex hot path for ordinary assignments; the richer lexer now runs only for quoted/bare fallback cases
- replace the per-binding interpolation dictionary with a bare-key shadow set
- add the exact `KEY junk` plus process-environment regression raised by Greptile

## Related issues

- Follow-up to #620 / #573

## Testing

- [x] `uv run pytest -m "not integration"` — 3351 passed, 6 skipped, 1 xpassed
- [x] targeted parser/parity suite — 230 passed
- [x] `ruff --select C901` with max complexity 9 — clean
- [x] deterministic 1,000-case differential check against python-dotenv 1.2.2 — zero mismatches
- [x] local 1,000-loop large-file comparison: merged `main` 451 µs/parse, follow-up 448 µs/parse (the prior PR head measured 7.8 ms vs 5.4 ms in CodSpeed)
- [x] `ruff check`, `ruff format`, full-project `pyrefly`, and `bandit` clean
- [ ] Full container/binary integration suite not run; the exact-head CI integration matrix covers it

## Checklist

- [x] Review findings addressed with real code/tests, not suppressions
- [x] No reviewer thread was resolved by the author
- [x] No tests disabled, skipped, or xfailed
- [x] Parser remains at 99% coverage
- [x] No external command/configuration behavior or documentation syntax changed

<a href="https://cubic.dev/pr/jainal09/envdrift/pull/622?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

This PR updates the dotenv key parser follow-up fixes. The main changes are:

- Splits key token parsing and binding preparation into smaller helpers.
- Restores the regex fast path for ordinary `KEY=value` lines.
- Tracks bare-key interpolation shadowing with a dedicated set.
- Adds regression tests for malformed bare lines and ambiguous `export` lines.

This looks safe to merge.

| Filename | Overview |
|----------|----------|
| src/envdrift/core/parser.py | Refactors key parsing and interpolation state while preserving the expected dotenv-compatible behavior. |
| tests/unit/test_parser_keys.py | Adds regression tests for malformed key lines, process environment fallback, and ambiguous export parsing. |

<sub>Reviews (4): Last reviewed commit: ["fix(parser): reject export prefix withou..."](https://github.com/jainal09/envdrift/commit/a36d924e3738faf81c4080fcb9ce3f2087dd5935) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43491968)</sub>

### DIFF
--- a/src/envdrift/core/parser.py
+++ b/src/envdrift/core/parser.py
@@ -142,6 +142,16 @@ class _KeyBinding:
     dropped: bool = False
 
 
+@dataclass(frozen=True)
+class _KeyToken:
+    """A key plus the unparsed text following its closing token."""
+
+    key: str
+    tail: str
+    consumed: int
+    quoted: bool
+
+
 class EnvParser:
     """Parse .env files with multi-backend encryption awareness.
 
@@ -321,10 +331,10 @@ class EnvParser:
         env_file = EnvFile(path=Path())
         content, env_file.leading_bom = self._strip_leading_bom(content)
         lines = self.LINE_BOUNDARY_PATTERN.split(content)
-        # Includes bare bindings as ``None`` for sequential interpolation.
-        # EnvFile.variables intentionally contains only values the Pydantic
-        # dotenv source can pass to model validation (#573).
-        interpolation_values: dict[str, str | None] = {}
+        pattern = self.LENIENT_LINE_PATTERN if lenient else self.LINE_PATTERN
+        # Bare bindings are omitted from EnvFile.variables but must shadow
+        # os.environ during sequential interpolation (#573).
+        bare_keys: set[str] = set()
         index = 0
         while index < len(lines):
             line_num = index + 1
@@ -341,29 +351,27 @@ class EnvParser:
                 env_file.comments.append(line)
                 continue
 
-            # Lex the key before the value. Unlike LINE_PATTERN, this accepts
-            # python-dotenv's single-quoted keys and consumes a quoted key
-            # across physical lines. That consumption is important even for
-            # bare bindings: interior ``K=1`` text belongs to the key and must
-            # never re-parse as a phantom assignment (#573).
-            binding = self._parse_key_binding(original_line, lines, index, lenient=lenient)
-            if binding is None:
-                continue
-            raw_lines = [original_line]
-            if binding.consumed:
-                raw_lines.extend(lines[index : index + binding.consumed])
-                index += binding.consumed
-            if binding.dropped or binding.key is None:
-                continue
-
-            key = binding.key
-            if binding.raw_value is None:
-                # A later bare duplicate wins in dotenv_values (A=1; A ->
-                # A=None), and pydantic-settings then filters that None out.
-                interpolation_values[key] = None
-                env_file.variables.pop(key, None)
-                continue
-            raw_value = binding.raw_value
+            # Keep the established regex as the hot path for ordinary
+            # KEY=value files. Quoted/bare bindings fall through to the richer
+            # key lexer; bypass a lenient match whose captured key starts with
+            # a quote, because python-dotenv removes those key quotes (#573).
+            match = pattern.match(original_line.lstrip())
+            if match is not None and not match.group(1).startswith("'"):
+                key = match.group(1)
+                raw_value = match.group(2)
+                raw_lines = [original_line]
+            else:
+                prepared, index = self._prepare_nonstandard_binding(
+                    original_line,
+                    lines,
+                    index,
+                    lenient,
+                    env_file,
+                    bare_keys,
+                )
+                if prepared is None:
+                    continue
+                key, raw_value, raw_lines = prepared
 
             # Quoted multiline continuation (#458): when the RHS opens a quote
             # that closes on a later physical line, the continuation lines are
@@ -389,8 +397,8 @@ class EnvParser:
             # the loader pydantic-settings uses — so validate/diff judge the
             # value the application receives, not the unexpanded literal.
             if "${" in value:
-                value = self._interpolate(value, interpolation_values)
-            interpolation_values[key] = value
+                value = self._interpolate(value, env_file.variables, bare_keys)
+            bare_keys.discard(key)
 
             # Determine encryption status and backend. Detection runs on the
             # interpolated value, so an alias of an encrypted variable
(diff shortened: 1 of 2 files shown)


src/envdrift/core/parser.py
```diff
@@ -414,6 +422,43 @@ def parse_string(self, content: str, *, lenient: bool = False) -> EnvFile:
...
+        index: int,
+        lenient: bool,
+        env_file: EnvFile,
+        bare_keys: set[str],
+    ) -> tuple[tuple[str, str, list[str]] | None, int]:
+        """Lex and prepare a quoted, bare, or malformed fallback binding."""
+        binding = self._parse_key_binding(original_line, lines, index, lenient=lenient)
+        if binding is None:
+            return None, index
+        return self._prepare_binding(binding, original_line, lines, index, env_file, bare_keys)
```
jainal09: Fixed in 9a361e8: _prepare_nonstandard_binding now receives one _ParseContext containing the shared lines, EnvFile, and bare-key set. Its explicit argument count is reduced to four without changing the standard-line hot path.

src/envdrift/core/parser.py
```diff
@@ -414,6 +422,43 @@ def parse_string(self, content: str, *, lenient: bool = False) -> EnvFile:
...
+        index += binding.consumed
+        if binding.dropped or binding.key is None:
+            return None, index
+        if binding.raw_value is None:
+            # A later bare duplicate wins in dotenv_values (A=1; A -> A=None),
+            # and pydantic-settings then filters that None out.
+            bare_keys.add(binding.key)
+            env_file.variables.pop(binding.key, None)
+            return None, index
+        return (binding.key, binding.raw_value, raw_lines), index
```
jainal09, line 485: Fixed in 9a361e8: _prepare_binding now takes binding, original_line, index, and one _ParseContext, reducing the explicit argument count to four. Targeted parity tests, full pyrefly, and the complexity gate remain clean.

src/envdrift/core/parser.py
```diff
@@ -341,29 +361,25 @@ def parse_string(self, content: str, *, lenient: bool = False) -> EnvFile:
...
-                interpolation_values[key] = None
-                env_file.variables.pop(key, None)
-                continue
-            raw_value = binding.raw_value
+            # Keep the established regex as the hot path for ordinary
+            # KEY=value files. Quoted/bare bindings fall through to the richer
+            # key lexer; bypass a lenient match whose captured key starts with
+            # a quote, because python-dotenv removes those key quotes (#573).
+            match = pattern.match(original_line.lstrip())
+            if match is not None and not match.group(1).startswith("'"):
```
jainal09: **Blocking: the hot-path regex backtracks on `export` and regresses #620's parity.**

`LINE_PATTERN`'s `^(?:export\s+)?([A-Za-z_][A-Za-z0-9_]*)\s*=` is not atomic: on a line where nothing valid follows the `export ` prefix, `re` retries with the optional group **unmatched** and captures `export` itself as the key. python-dotenv's lexer consumes the export prefix separately (no backtracking) and errors the line — and so did #620's `_parse_key_binding`. Verified on real python-dotenv 1.2.2:

| input | dotenv | #620 (`10ba096`, main) | this head (`a0e6ad1`) |
|---|---|---|---|
| `export =v` | `{}` | `{}` ✅ | `{'export': 'v'}` ❌ |
| `export  =B=x` | `{}` | `{}` ✅ | `{'export': 'B=x'}` ❌ |
| `export \t =` | `{}` | `{}` ✅ | `{'export': ''}` ❌ |

(For contrast, `export=v` — no space — is a legitimate assignment named `export` in dotenv, and the #620 lexer already handles it correctly; a fix must keep that case working.)

**Minimal fix:** route export-prefixed lines around the hot path — they're the only inputs where the regex can backtrack:
```python
match = pattern.match(text)
if (
    match is not None
    and not match.group(1).startswith("'")
    and not (match.group(1) == "export" and self.EXPORT_PATTERN.match(text))
):
    ... hot path ...
else:
    ... richer lexer (already correct for every export form) ...
```
Ordinary `KEY=value` lines are untouched, so the CodSpeed win is preserved; only genuine `export ...` lines take the lexer.

**Also please:**
1. Add the `export =v` / `export  =B=x` / `export=v` family to the parity suite — the exact backtracking trap, so the hot path can never silently drift from the lexer again.
2. Extend your differential generator so `export ` can be immediately followed by `=` — your 1,000-case check reported zero mismatches while this divergence existed, which means the generator never produced this adjacency. Mine caught it in 800 cases with `"export "` and `"="` as independent alphabet tokens.

Everything else in this PR verified clean on my side — fix this one seam and I'll merge.

jainal09 (commented):
Dogfooded the refactor with a fresh independent differential fuzz (new seed, `export`-adjacent alphabet). The complexity work, the `KEY junk` regression, and the perf restoration are all good — but the restored hot path **reintroduces a parity divergence that #620 had (silently) fixed**. One blocking inline comment with the repro and a minimal fix. Everything else verified: behavior identical to dotenv on all my targeted cases, full suite green locally.

src/envdrift/core/parser.py
jainal09: Fixed in 87c6642. _standard_line_match now detects the regex backtrack shape where group 1 is export while EXPORT_PATTERN matched a real prefix, and routes only that case through the full lexer. Added parity regressions for export =v, export  =B=x, export-tab-space-=, and the valid export=v assignment. I also reran an expanded 2,000-case differential generator with export-space and equals as independently adjacent tokens: zero mismatches against python-dotenv 1.2.2. Full non-integration gate: 3,355 passed, 6 skipped, 1 xpassed; static/security gates clean.

jainal09: Rebased onto current main release commit 585422b; the same fix is now a36d924. Exact-head targeted parity/static gates remain green (234 tests), with the full 3,355-test non-integration run already green before the release-only rebase.

jainal09: **Resolved — verified on `87c66429`.** `_standard_line_match` implements exactly the suggested guard, and the 4-case regression family (incl. the `export=v` legitimate-key contrast) landed. My re-verification:

| input | dotenv | this head |
|---|---|---|
| `export =v` | `{}` | `{}` ✅ |
| `export  =B=x` | `{}` | `{}` ✅ |
| `export \t =` | `{}` | `{}` ✅ |
| `export=v` | `{'export':'v'}` | `{'export':'v'}` ✅ |
| `export A=1` | `{'A':'1'}` | `{'A':'1'}` ✅ |

Plus a fresh independent 800-case differential fuzz (new seed, bare `export` token added to the alphabet): **clean**. Full suite 3355 passed; C901 ≤ 9 clean.

Perf note for the record (500-line ordinary file, my machine): pre-#620 737µs → #620/main 1049µs → this head 989µs. So this PR improves on main; the residual gap vs the old baseline is the inherent cost of the parity lexer's bookkeeping and is immaterial in absolute terms (sub-ms, once per invocation) — CodSpeed's main-relative gate is the arbiter. Merging on green CI.

src/envdrift/core/parser.py
jainal09: Verified on `87c66429` as part of the full pass: `_ParseContext` bundling is in place, full non-integration suite 3355 passed, pyrefly 0 errors, C901 ≤ 9 clean, and my independent differential fuzz confirms no behavior drift from the restructure. Resolving.

src/envdrift/core/parser.py
jainal09: Verified on `87c66429` as part of the full pass: `_ParseContext` bundling is in place, full non-integration suite 3355 passed, pyrefly 0 errors, C901 ≤ 9 clean, and my independent differential fuzz confirms no behavior drift from the restructure. Resolving.

PR #716, opened 2026-08-02
fix(deps): bring every pin to latest and repair the Renovate tracking that let them freeze

jainal09:
## What

Brings every pinned dependency to latest **and repairs the mechanisms that let them silently freeze in the first place**. The version bumps are the smaller half of this PR; the important half is that three separate "Renovate is watching this" claims turned out to be false.

Full suite green end to end: **4414 passed, 12 skipped**, run in one process against the real container stack and real binaries.

### 1. Renovate tracked the wrong Infisical repository

`renovate.json` watched `Infisical/infisical` with `extractVersionTemplate: ^infisical-cli/v(?<version>.*)$`. That repo's **last CLI release was 0.41.90 — exactly our pin** — so Renovate reported "up to date" forever with zero diagnostics. Its newer `v0.162.x` tags are *platform* releases carrying no CLI assets.

Fixing only the regex would have been worse than leaving it: it would have re-pinned us to a dead stream. The CLI moved to `Infisical/cli`, and three things had to change together or the installer 404s:

| | old | new |
|---|---|---|
| repo | `Infisical/infisical` | `Infisical/cli` |
| tag | `infisical-cli/v{version}` | `v{version}` |
| asset | `infisical_{v}_{os}_{arch}` | `cli_{v}_{os}_{arch}` |

**macOS installs would have broken silently.** Upstream now splits digests: `checksums.txt` has 42 entries covering linux/windows/bsd and **zero** darwin entries; macOS digests live in `checksums-darwin.txt`. Since `verify_download` fails closed, resolving the generic file on macOS aborts every install — and CI is Linux-only, so no lane would have caught it. `get_checksums_url()` is now platform-aware with a fallback to the generic URL, so a future upstream re-consolidation degrades gracefully.

### 2. `lockFileMaintenance` was disabled

Absent from `renovate.json` means **disabled** under `config:recommended`. That is why `uv.lock` had drifted **48 packages** behind. Now enabled on the repo's existing weekend window with automerge off.

### 3. LocalStack had become account-gated

Since 2026.03.0 the image exits(55) with "License activation failed" without `LOCALSTACK_AUTH_TOKEN`; the tokenless community image is gone and the bypass expired 2026-04-06. `4.14` was the last release that ran unattended.

[...]

PR #720, opened 2026-08-03
fix(renovate): track Go indirect deps, pin CI tooling, close audit gaps

### LINKED ISSUE #721: Go agent has no vulnerability gate: govulncheck missing from agent CI
Found during the Renovate coverage audit behind #720.

## Problem

`agent-ci.yml` runs build, test, and golangci-lint — but never `govulncheck`. Combined with three silent channels, this let a real vulnerability ship unnoticed:

- **GO-2026-5024 / CVE-2026-39824** (integer overflow in `NewNTUnicodeString`) was present in `golang.org/x/sys@v0.30.0`, and under `GOOS=windows GOARCH=amd64` govulncheck reports it in **Package Results** — the vulnerable package is genuinely linked into the released `envdrift-agent-windows-amd64.exe` (not proven symbol-reachable, but present).
- Renovate skipped it (indirect deps disabled by default — fixed in #720).
- Dependabot had **zero** Go alerts for this repo.
- GitHub's advisory DB has no GHSA covering v0.30.0 for this CVE.

So all three automated channels were blind simultaneously. #720 fixes the Renovate channel; this issue is the defense-in-depth gate.

## Proposed fix

Add to `agent-ci.yml` (one matrix leg, after tests):

```yaml
- name: govulncheck
  run: |
    go install golang.org/x/vuln/cmd/govulncheck@latest
    govulncheck ./...
    GOOS=windows GOARCH=amd64 govulncheck ./...
```

Notes:
- The `GOOS=windows` pass matters: the CVE above only surfaces in Package Results under Windows analysis.
- Pin govulncheck via the repo's conventions rather than `@latest` (Renovate-visible), consistent with #720.
- Decide whether the step should be required; per CLAUDE.md, Agent Lint is deliberately non-required today, so a red govulncheck would currently be UNSTABLE-but-mergeable.

## Verification

`cd envdrift-agent && GOOS=windows GOARCH=amd64 govulncheck ./...` — after #720's x/sys bump lands, GO-2026-5024 should be absent; the gate then keeps it that way.

jainal09:
## What

Closes the confirmed gaps from an adversarially-verified audit of what Renovate actually tracks (asked by the maintainer after #716: "should the dashboard track more things now?"). 40 candidate gaps were generated, then each was independently re-verified against Renovate's source and this repo's live state: **5 confirmed, 7 refuted**. Only confirmed gaps are fixed here; the refuted ones are recorded below so nobody re-raises them.

### 1. All 12 indirect Go modules were invisible to Renovate (high)

Renovate's gomod extractor stamps `enabled: false` on every `// indirect` require (absent a Go 1.24+ `tool` directive) — no updates, no dashboard entry, no diagnostic. Measured rot: `golang.org/x/sys` at v0.30.0, **17 minors / ~18 months behind**, carrying **GO-2026-5024 / CVE-2026-39824** — and `govulncheck` under `GOOS=windows` shows the vulnerable package genuinely linked into the shipped `envdrift-agent-windows-amd64.exe`. Silent in *every* channel: no Renovate entry, zero Go Dependabot alerts, no GHSA covering v0.30.0. The Infisical failure mode, again.

Fix: `matchDepTypes: ["indirect"]` rule (grouped — first weekend run opens 1 PR, not 12), `postUpdateOptions: ["gomodTidy"]`, `go mod tidy` committed here (prunes exactly the 6 stale go.sum hashes prior bumps left), and a `go mod tidy -diff` CI gate on one matrix leg.

**Why the Go floor moved 1.23 → 1.25.0 in the same PR:** x/sys ≥ v0.44.0 (the CVE fix) declares `go 1.25.0`. Without the floor bump the indirect PR can never merge; the agent-ci matrix leg moves with it. Deliberately **not** using `constraintsFiltering: "strict"` — it would cap x/sys below the CVE fix forever, which is silent non-coverage in new clothes. Agent verified on go1.25: build green, all package tests pass, tidy clean.

### 2. `uv` installed as `latest` on every CI run (high)

Under all 12 required checks *and* `uv build`/`uv publish` on the release path. All 10 `setup-uv` sites now pin `version: 0.12.1` via the input **Renovate's built-in github-actions manager already extracts** as `astral-sh/uv`. Deliberately no `# renovate:` annotations — they would double-track the dep through the regex manager (a guard now rejects them if reintroduced).

### 3. `npx markdownlint-cli2` unpinned inside the required Lint check (high)

Fresh tarball from npm on every run; an upstream release could redden a required gate with no PR and no warning. Pinned as an annotated Makefile variable with a new Makefile customManager. Datasource is `npm` on purpose: the project tags but publishes **zero GitHub releases**, so `github-releases` would resolve nothing and never bump — the exact Infisical shape.

### 4–5. gomodTidy + PR-body QoL

`prBodyTemplate` regains `{{{warnings}}}`/`{{{controls}}}` — honest framing: **QoL, not protection**. Verification proved the Infisical freeze was `extractVersionTemplate` filtering (silent by design, no lookup error), so no warning would have fired; the dashboard renders dep warnings regardless of PR template. The rebase checkbox is the real win.

Plus: `envdrift-agent/CONTRIBUTING.md` told contributors to `go install` golangci-lint from the **v1 module path** (tops out at 1.64.8) while CI runs v2.12.2 — the v2 path needs the `/v2/` segment. Now pinned to CI's version.

## Refuted — do not re-raise

| Claim | Why it's wrong |
|---|---|
| "uv.lock not covered by lockFileMaintenance" (PR #719 touched only npm) | pep621 exports `supportsLockFileMaintenance=true` with `uv.lock`; #719 was a legitimate no-op — #716 had regenerated uv.lock 11 min earlier; `uv lock --upgrade --dry-run` → "No lockfile changes detected" |
| "go 1.23 vs CI 1.26 is drift" | Compatibility floor vs build toolchain — intentional (though the floor moves here for the CVE fix, not for parity) |
| "commitlint ruleset invisible" | Dashboard lists `wagoid/commitlint-github-action v6` explicitly |
| "floating `@vN` action tags are a gap" | Deliberate, tested policy; repo is its own natural experiment |
| "Makefile `uv run` without `--locked` rewrites the lock" | CI runs `--locked`; local re-resolve is loud in `git status`, not silent |
| "prBodyTemplate suppression enabled the Infisical freeze" | The freeze produced **no warning to suppress** — `applyExtractVersion` drops non-matching releases with no logging; warnings also render on the dashboard, which the template can't touch |
| "need a repo-side uv.lock liveness assertion" | `uv sync --locked` in CI already fails on a stale lock |

## Guards

Four new tests in `test_dev_stack_hygiene.py`, each verified to fail when its invariant is broken (two by actually breaking it): `test_go_indirect_dependencies_are_tracked`, `test_go_floor_supports_current_x_sys`, `test_setup_uv_steps_pin_the_uv_version`, `test_makefile_tool_pins_are_renovate_visible`.

## Verification

- ruff / format / pyrefly / bandit / markdownlint all clean; every workflow parses
- non-integration suite **3905 passed**; hygiene module 22 passed
- Go agent: `go build` + all package tests green on go1.25, `go mod tidy -diff` clean
- `make lint-docs` green through the pinned markdownlint-cli2@0.23.2

## Expected after merge

Dashboard #37: `regex (9)` → `regex (10)` with a `Makefile (1)` block; `astral-sh/uv 0.12.1` under github-actions; first weekend run opens one grouped "go indirect dependencies" PR. If any of those don't appear, the corresponding guard's premise is wrong — file it, don't ignore it.

## Also closes #721 — the govulncheck gate

Fixes #721

Originally filed as a follow-up, folded in at the maintainer's direction. The gate proved itself before landing: govulncheck v1.6.0's newer DB surfaced a **second** vulnerability nothing else had found — **GO-2026-4602** (stdlib `os`, symbol-reachable via fsnotify → `os.ReadDir`), fixed here with a Renovate-tracked `toolchain go1.25.12` directive. `agent-ci.yml` now runs both govulncheck passes (native + `GOOS=windows`) each proven necessary by one of the two incidents, pinned and Renovate-visible, with guard-the-guard tests keeping both gate conditions alive across matrix rotations. Both passes: "No vulnerabilities found."

<a href="https://cubic.dev/pr/jainal09/envdrift/pull/720?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Chores**
  * Updated supported Go tooling to version 1.25.
  * Standardized CI environments on uv 0.12.1.
  * Pinned documentation linting to markdownlint-cli2 0.23.2.
  * Updated contributor guidance for the current linting tool version.
  * Improved automated dependency update tracking and Go module maintenance.

* **Tests**
  * Added checks to verify toolchain versions, dependency updates, and development-tool pinning across CI configuration.

<h3>Greptile Summary</h3>

The PR expands Renovate coverage and makes CI tooling reproducible while updating the Go agent’s dependency and vulnerability-audit baseline.
- Enables grouped Renovate updates and tidy maintenance for indirect Go modules.
- Raises the agent’s Go floor to 1.25.12 and updates `golang.org/x/sys`.
- Adds native and Windows-binary `govulncheck` gates.
- Pins uv and markdownlint tooling across CI and local automation.
- Adds regression tests for Renovate coverage and CI-tooling invariants.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| envdrift-agent/go.mod | Raises the required Go version to 1.25.12 and updates the vulnerable indirect `x/sys` dependency. |
| .github/workflows/agent-ci.yml | Updates the Go matrix and adds tidy plus native/Windows vulnerability checks. |
| renovate.json | Adds custom managers, indirect Go dependency tracking, grouped updates, tidy maintenance, and restored PR controls. |
| tests/unit/test_renovate_coverage.py | Consolidates and expands regression coverage for Renovate extraction and CI-tool pinning. |
| envdrift-agent/CONTRIBUTING.md | Aligns the documented Go prerequisite and golangci-lint installation command with the updated agent toolchain. |

<sub>Reviews (9): Last reviewed commit: ["fix(ci): audit a built windows binary, n..."](https://github.com/jainal09/envdrift/commit/02e2c66e55a25ffd83ae8cf21f6f7d09787e83fb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49574871)</sub>

**Context used:**

- Knowledge Base — [The envdrift-agent Go Daemon](https://app.greptile.com/my-company-org-6979/-/custom-context/knowledge-base/jainal09/envdrift/-/docs/go-agent-daemon.md)


.github/workflows/agent-ci.yml
```diff
@@ -43,6 +43,13 @@ jobs:
       - name: Download dependencies
         run: go mod download
 
+      - name: Verify go.mod/go.sum are tidy
+        # One leg only: -diff output could differ across toolchains, and a
+        # split failure across 6 legs is noise. Renovate's gomodTidy keeps
+        # this green for bot PRs; contributors run `go mod tidy` locally.
+        if: matrix.os == 'ubuntu-latest' && matrix.go == '1.26'
+        run: go mod tidy -diff
```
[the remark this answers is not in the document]
jainal09: Confirmed and fixed in `e8a85d6` — excellent find, and your repro holds up here too.

Reproduced on this repo before changing anything (go1.25.4, a `go.sum` line removed):

```
tidy-diff BEFORE `go mod download`: exit=1
tidy-diff AFTER  `go mod download`: exit=0
```

`go mod download` wrote the missing checksum on the runner, and the gate then blessed files that are untidy as committed — exactly your mechanism.

Took the first of your two options: the step now runs **before** "Download dependencies" (`go mod tidy -diff` itself never mutates files, so the ordering is safe), with the masking behaviour documented in the workflow comment so nobody "optimises" the order back. Resolving.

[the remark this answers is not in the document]
jainal09: Declining — the premise doesn't hold, verified empirically on go1.25.4 in this repo:

```
$ go mod tidy -diff            # clean tree
exit=0
$ printf 'github.com/fsnotify/fsnotify v1.9.0 h1:AAA...=\n' >> go.sum
$ go mod tidy -diff >/dev/null; echo exit=$?
exit=1
```

Go's documentation agrees: `-diff` "exits with a non-zero code if updates are needed" (added in Go 1.23). So the step does gate untidied files.

That said, the *adjacent* codex-review thread found a real flaw in the same step — `go mod download` running first could repair the runner's go.sum and mask the check — which is fixed in `e8a85d6` by moving the gate before the download. So the step is now both failing-on-diff (as it already did) and unmaskable. Resolving.

tests/unit/test_dev_stack_hygiene.py
```diff
@@ -670,3 +670,110 @@ def test_every_manager_matches_a_real_upstream_tag() -> None:
             "cannot extract a version, so this pin is frozen with no diagnostic "
             "— exactly how infisical sat at 0.41.90 after the CLI moved repos."
         )
+
```
jainal09: Fixed in `e8a85d6` by splitting, not suppressing — the finding was fair: the module had accreted eight concerns.

The Renovate-vs-upstream guards (manager regex/tool coverage, repo/tag matching, the live upstream-tag regression test, and the 2026-08 audit invariants) moved to a new `tests/unit/test_renovate_coverage.py`, which owns exactly one concern and carries the Infisical incident as its module docstring. `test_dev_stack_hygiene.py` keeps the container-stack parity, token and automerge-copy guards. All 22 tests pass across both modules. Resolving.

tests/unit/test_dev_stack_hygiene.py
```diff
@@ -670,3 +670,110 @@ def test_every_manager_matches_a_real_upstream_tag() -> None:
...
+            "annotation creates a duplicate dependency"
+        )
+        doc = yaml.safe_load(text)
+        for job in (doc.get("jobs") or {}).values():
+            for step in job.get("steps") or []:
+                if "astral-sh/setup-uv" in str(step.get("uses", "")):
+                    version = (step.get("with") or {}).get("version")
+                    (pinned if version else bare).append(f"{wf.name}")
+    assert pinned, "no setup-uv steps found — did the workflows move?"
+    assert not bare, f"setup-uv steps without a pinned uv version: {bare}"
```
jainal09: Fixed in `e8a85d6`. The three nested loops + conditional are now a flat comprehension helper (`_setup_uv_steps()` yielding `(workflow, step)` pairs), and the test body is two flat asserts — one for unpinned steps, one for duplicate-tracking annotations. Behaviour identical; verified it still fails when a `version:` input is removed. Resolving.

envdrift-agent/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/jainal09/envdrift-agent
 
-go 1.23
+go 1.25.0
```
jainal09: [codex-review] **[P1] Land the x/sys security update with this prerequisite**

This raises the Go floor specifically so `x/sys >= v0.44.0` can land, but the same committed `go.mod` still selects `golang.org/x/sys v0.30.0` at line 24. The [official Go advisory](https://pkg.go.dev/vuln/GO-2026-5024) marks every version before v0.44.0 affected. The new `test_go_floor_supports_current_x_sys` checks only the Go directive and CI matrix, so it passes while the known-vulnerable module remains selected; issue #721 likewise refers to “#720's x/sys bump,” but this diff contains no such bump. Update x/sys and its sums to at least v0.44.0 in this PR and assert the selected module version, rather than waiting for a weekend Renovate run after merging a known-vulnerable shipped dependency.

envdrift-agent/CONTRIBUTING.md
```diff
@@ -49,7 +49,10 @@ go test -v ./internal/config/...
 
 ```bash
 # Install golangci-lint
-go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+# v2 module path — the /cmd/... path without /v2 silently installs the
+# abandoned v1 line (tops out at 1.64.x) while CI runs v2; keep the version
+# in sync with .github/workflows/agent-ci.yml
+go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2
```
jainal09, line 55: [codex-review] **[P2] Make the new contributor-tool pin Renovate-visible**

This replaces `@latest` with a second literal `v2.12.2`, but `envdrift-agent/CONTRIBUTING.md` is outside every custom manager's `managerFilePatterns` (the relevant managers scan workflows and the Makefile only). When Renovate bumps the annotated `version:` in `agent-ci.yml`, this command remains stale despite the “keep in sync” comment—recreating the silent copied-pin drift this PR is intended to close. Add a Renovate-visible annotation/manager for this command, or add a regression guard that extracts and compares it with the workflow pin so a bot bump cannot merge only one copy.

envdrift-agent/go.mod
jainal09: Confirmed and fixed in `c811d61` — you were right, and this was the correct call over my original sequencing.

I had raised the floor as the prerequisite and deferred the actual bump to Renovate's weekend run, which — as you say — merges a known-vulnerable module in the shipped Windows binary and contradicts both the repo's no-known-broken-merges rule and #721's own wording.

`golang.org/x/sys` is now **v0.47.0** in this PR. Verified end to end:

```
go: upgraded golang.org/x/sys v0.30.0 => v0.47.0
BUILD OK · all package tests pass · go mod tidy -diff clean
$ GOOS=windows GOARCH=amd64 govulncheck ./...
No vulnerabilities found.        (previously: GO-2026-5024 in Package Results)
```

And your point about the guard passing while the vulnerable module stayed selected is addressed exactly as suggested: `test_go_floor_supports_current_x_sys` now also asserts the **selected** x/sys is ≥ v0.44.0 (the advisory's fixed version), not just the directive and matrix. Resolving.

envdrift-agent/CONTRIBUTING.md
jainal09: Confirmed and fixed in `c811d61` — took your first option, plus the guard as belt-and-braces.

A new customManager targets the exact install command in `envdrift-agent/CONTRIBUTING.md` with the same `depName` (`golangci/golangci-lint`, github-releases), so a Renovate bump lands in **one PR that updates both copies** — the doc can no longer be stranded when the agent-ci.yml annotation moves.

The parity guard (`test_golangci_lint_doc_pin_matches_the_workflow_pin`) asserts the two pins are equal AND that the new manager's regex still matches the doc — a manager whose regex matches nothing reports up-to-date forever, which is the Infisical failure mode this PR exists to close.

One honest note: wiring the new manager in exposed an over-broad assumption in my own guards — the constants.json URL-consistency checks tried to demand download URLs from it (a doc-pin manager has none by design). Scoped them to `_constants_managers()`; the live upstream-tag test still sweeps every github-releases manager. Resolving.

.github/workflows/agent-ci.yml
```diff
@@ -40,6 +40,17 @@ jobs:
...
+      - name: Verify go.mod/go.sum are tidy
+        # BEFORE 'Download dependencies': `go mod download` writes missing
+        # go.sum checksums on the runner, which would mask untidiness of the
+  # COMMITTED files (verified: a dropped go.sum line fails -diff before
+  # download, passes after). One leg only: -diff output could differ
+  # across toolchains, and a split failure across 6 legs is noise.
+  # Renovate's gomodTidy keeps
+        # this green for bot PRs; contributors run `go mod tidy` locally.
+        if: matrix.os == 'ubuntu-latest' && matrix.go == '1.26'
+        run: go mod tidy -diff
```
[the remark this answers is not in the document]
jainal09: Already fixed — this review ran against the pre-`e8a85d6` head. The masking you describe is real (I reproduced it: a dropped go.sum line fails `-diff` before `go mod download`, passes after), and that commit moved the gate before the download step for exactly that reason:

```
steps: [Checkout code, Set up Go, Verify go.mod/go.sum are tidy, Download dependencies, Run tests, ...]
```

The workflow comment documents the masking mechanism so the order doesn't get "optimised" back. Same finding as the codex thread on this line, resolved there with the repro. Resolving as duplicate/already-fixed.

envdrift-agent/CONTRIBUTING.md
jainal09, line 55: **Stale prerequisite a few lines up: line 9 of this file still says `Go 1.23+`, but this PR raises `go.mod` to `go 1.25.0` and the CI floor leg to 1.25.**

A contributor who installs Go 1.23/1.24 per the prerequisites gets a working build only if `GOTOOLCHAIN=auto` quietly downloads 1.25 for them; with `GOTOOLCHAIN=local` (common in corporate and hermetic-CI setups) every `go` command fails with `go.mod requires go >= 1.25.0`. Greptile's summary called this out and it has survived two review rounds — it can't be anchored inline because line 9 isn't in the diff, hence this comment on the nearest hunk.

Fix is one line:

```markdown
- Go 1.25+
```

(Repo convention: docs updated in the same PR as the behavior change — and this is the same file the PR already edits 40 lines below.)

.github/workflows/agent-ci.yml
```diff
@@ -40,6 +40,17 @@ jobs:
...
           cache-dependency-path: envdrift-agent/go.sum
 
+      - name: Verify go.mod/go.sum are tidy
+        # BEFORE 'Download dependencies': `go mod download` writes missing
+        # go.sum checksums on the runner, which would mask untidiness of the
+  # COMMITTED files (verified: a dropped go.sum line fails -diff before
+  # download, passes after). One leg only: -diff output could differ
+  # across toolchains, and a split failure across 6 legs is noise.
+  # Renovate's gomodTidy keeps
+        # this green for bot PRs; contributors run `go mod tidy` locally.
```
jainal09: Nit (polish): four of these comment lines are dedented to column 3 and the sentence breaks mid-clause (“Renovate's gomodTidy keeps” / “this green”). YAML ignores comment indentation so nothing breaks, but it reads like a patch that half-applied — worth reflowing before it fossilizes:

```suggestion
        # BEFORE 'Download dependencies': `go mod download` writes missing
        # go.sum checksums on the runner, which would mask untidiness of the
        # COMMITTED files (verified: a dropped go.sum line fails -diff before
        # download, passes after). One leg only: -diff output could differ
        # across toolchains, and a split failure across 6 legs is noise.
        # Renovate's gomodTidy keeps this green for bot PRs; contributors
        # run `go mod tidy` locally.
```

.github/workflows/agent-ci.yml
```diff
@@ -40,6 +40,17 @@ jobs:
...
+        if: matrix.os == 'ubuntu-latest' && matrix.go == '1.26'
```
jainal09, line 51: **The tidy gate dies silently when this matrix rolls forward.** `matrix.go == '1.26'` happens to keep the gate alive through the next rotation (`['1.26', '1.27']` still contains a `'1.26'` leg), but the rotation after that (`['1.27', '1.28']`) makes the condition permanently false — the only tidiness gate in CI evaporates with zero diagnostics, which is precisely the silent-guard-death failure mode this PR exists to close everywhere else.

Cheapest fix is a guard-the-guard assert in `test_go_floor_supports_current_x_sys` (it already parses this workflow's matrix):

```python
    tidy_steps = [
        s
        for s in ci["jobs"]["test"]["steps"]
        if "go mod tidy -diff" in str(s.get("run", ""))
    ]
    assert tidy_steps, "agent-ci.yml lost the go mod tidy gate"
    condition = tidy_steps[0].get("if", "")
    assert any(f"'{leg}'" in condition for leg in matrix_go), (
        "the tidy gate's `if:` pins a go version no longer in the matrix - "
        "the gate would silently never run again"
    )
```

(Alternative: hoist the tidy check into its own tiny job so it isn't coupled to a matrix leg at all — but the assert above is less workflow churn.)

tests/unit/test_renovate_coverage.py
```diff
@@ -0,0 +1,423 @@
...
+        "every version below v0.44.0 is affected by GO-2026-5024 "
+        "(package-imported in the shipped windows-amd64 binary)"
+    )
+
+
+def _setup_uv_steps() -> list[tuple[str, dict[str, Any]]]:
+    """Every (workflow-name, step) pair whose step uses astral-sh/setup-uv."""
+    return [
+        (wf.name, step)
+        for wf in sorted(_WORKFLOWS.glob("*.yml"))
```
jainal09: Hardening: this sweep (and the `annotated` sweep at line 353) globs `*.yml` only, while `renovate.json`'s own workflows `managerFilePatterns` accepts `ya?ml`. A future workflow saved as `.yaml` with a bare `setup-uv` step (or a stray `astral-sh/uv` annotation) escapes both asserts while Renovate half-tracks it — the guard reports full coverage it no longer has.

```suggestion
        for wf in sorted(p for ext in ("*.yml", "*.yaml") for p in _WORKFLOWS.glob(ext))
```

tests/unit/test_renovate_coverage.py
```diff
@@ -0,0 +1,423 @@
...
+    reports up-to-date forever).
+    """
+    makefile = (_REPO_ROOT / "Makefile").read_text(encoding="utf-8")
+    managers = [
+        m
+        for m in _renovate_config().get("customManagers", [])
+        if any("Makefile" in p for p in m.get("managerFilePatterns", []))
+    ]
+    assert managers, "renovate.json lost the Makefile customManager"
+    pattern = re.sub(r"\(\?<(\w+)>", r"(?P<\1>", managers[0]["matchStrings"][0])
```
jainal09: Nit: this validates only `managers[0]["matchStrings"][0]` (same pattern in the new parity guard at line 419). If a second Makefile pin ever gets its own manager entry or a second `matchStrings` pattern, the extra regex ships unvalidated — a regex matching nothing being exactly the failure mode the docstring names. Looping costs two lines:

```python
    matches: dict[str, str] = {}
    for manager in managers:
        for match_string in manager["matchStrings"]:
            pattern = re.sub(r"\(\?<(\w+)>", r"(?P<\1>", match_string)
            matches.update(
                (m.group("depName"), m.group("currentValue"))
                for m in re.finditer(pattern, makefile)
            )
```

renovate.json
```diff
@@ -128,8 +128,30 @@
...
+      ]
+    },
+    {
+      "customType": "regex",
+      "description": "golangci-lint install command in the agent contributor docs. Keeps the doc copy in the SAME Renovate PR as the agent-ci.yml pin, so the two cannot drift the way copied pins otherwise do.",
+      "managerFilePatterns": [
+        "/^envdrift-agent/CONTRIBUTING\\.md$/"
+      ],
+      "matchStrings": [
+        "go install github\\.com/golangci/golangci-lint/v2/cmd/golangci-lint@(?<currentValue>v\\d+\\.\\d+\\.\\d+)"
```
jainal09, line 151: Low, future-major footgun worth a `TODO`: both this regex and the doc command it matches hardcode the `/v2/` module path. When golangci-lint cuts v3, Renovate will bump `@v2.12.2` → `@v3.x` in **both** copies (majors are manual-review, so a human is in the loop) while the `/v2/cmd/` path stays — producing an install command that resolves the wrong module, and the new parity guard stays green because both copies broke identically. Nothing to change today (v2.12.2 is current); just flagging that the major-bump reviewer must also move the module path in `envdrift-agent/CONTRIBUTING.md` and in this `matchStrings` regex.

jainal09 (commented):
## Review — deps/renovate-coverage-gaps @ c811d61

Reviewed the full diff at head `c811d61` (the review scope moved twice while I worked — `e8a85d6` closed the tidy-gate masking window, `c811d61` shipped the x/sys bump + doc-pin manager — both re-verified here). **No correctness blockers.** The load-bearing claims all check out against upstream reality; what's left is one stale doc line and three guard-hardening items, each with a concrete fix inline.

### Verified against upstream (evidence, so nobody re-litigates)

- **Renovate really does track `setup-uv`'s `version:` input.** `lib/modules/manager/github-actions/community.ts` maps `astral-sh/setup-uv` → package `astral-sh/uv` (github-releases datasource, npm versioning). The pin-without-annotation design is correct, and the no-annotation guard is right to forbid double-tracking.
- **`uv 0.12.1`** is the latest release tag at astral-sh/uv — all 10 sites pin current.
- **`markdownlint-cli2 0.23.2`** is npm `latest`, and the repo publishes **zero** GitHub releases (`releases?per_page=5` → `0`) — the `npm` datasource choice is not just justified, `github-releases` would have been the Infisical freeze again.
- **`golang.org/x/sys v0.44.0`'s `.mod` declares `go 1.25.0`** (proxy.golang.org) — the floor-bump premise and the hardcoded `(1, 25)` in the guard are correct. GO-2026-5024 = CVE-2026-39824, fixed at 0.44.0 per vuln.go.dev; the tree now selects v0.47.0 (current latest), with the new selected-version assert closing the floor-without-fix hole.
- **`golangci-lint v2.12.2`** is the current latest; the `/v2/` module-path correction in CONTRIBUTING.md is real (the v1 path tops out at 1.64.x).
- **All 23 guards pass locally at `c811d61`**, tree clean; `ruff check` clean on both test modules. The tidy-before-download ordering fix is sound (`git diff` of the step order confirms it runs before `go mod download` under the job's `envdrift-agent` working directory).
- **`publish.yml` now exports `GH_TOKEN` for the whole-suite run** — the release-lane guard-skip flagged on #716 is closed; the upstream-tag test executes where `-m integration` runs.
- `prBodyTemplate`'s `{{{warnings}}}`/`{{{controls}}}` are valid default-template sections; the Makefile customManager regex was exercised end-to-end by the new guard (JS→Python named-group conversion + live match).

### Findings (all inline, each with a fix)

1. **`envdrift-agent/CONTRIBUTING.md` line 9 still says `Go 1.23+`** while this PR moves go.mod to 1.25.0 — the one Greptile summary item that survived both review rounds. One-line fix.
2. **agent-ci.yml comment block is mangled** (four lines dedented, sentence broken mid-clause) — cosmetic, suggestion attached.
3. **The tidy gate is keyed to `matrix.go == '1.26'`** and dies silently once the matrix rolls past 1.26 — the exact silent-guard-death pattern this PR closes elsewhere. Guard-the-guard assert suggested.
4. **`_WORKFLOWS.glob("*.yml")` misses `.yaml`** in both setup-uv sweeps; renovate.json's own file pattern accepts both. One-line hardening.
5. Nits: `managers[0]["matchStrings"][0]`-only validation (two sites); `/v2/` module path hardcoded in the new doc manager (major-bump footnote, nothing to change today).

The `c811d61` round itself was the right call on both codex findings — shipping the CVE fix inside the PR honors no-known-broken-merges, and the doc-pin customManager + parity guard is a better fix than the sync-by-comment it replaces.

envdrift-agent/CONTRIBUTING.md
jainal09: Fixed in `8721528` — and thank you for persisting with it across rounds; an un-anchorable line is exactly how a finding gets lost.

`- Go 1.23+` → `- Go 1.25+`. Your `GOTOOLCHAIN=local` scenario is the convincing part: with auto-download disabled, a contributor following the prerequisites literally cannot run a single `go` command against this module. Resolving.

.github/workflows/agent-ci.yml
jainal09: Confirmed and fixed in `8721528` — the sharpest find of the round. The gate I added to prevent silent guard death was itself wired to die silently: `matrix.go == '1.26'` survives one rotation by luck and goes permanently false on the next, taking the only tidiness gate in CI with it, no diagnostics.

Adopted your guard-the-guard essentially verbatim in `test_go_floor_supports_current_x_sys` (it already parses the matrix): the tidy step must exist and its `if:` condition must reference a leg present in the current matrix. A rotation that strands the gate now fails the suite instead of silently disarming CI. Resolving.

tests/unit/test_renovate_coverage.py
jainal09: Fixed in `8721528`. Right — the guard globbed `*.yml` while renovate.json's own `managerFilePatterns` accept `ya?ml`, so a future `.yaml` workflow would escape both asserts while Renovate half-tracks it: the guard reporting coverage it no longer has.

Both sweeps (the setup-uv pin check and the stray-annotation check) now go through a shared `_workflow_files()` helper covering both extensions, with the rationale in its docstring. Resolving.

tests/unit/test_renovate_coverage.py
jainal09: Fixed in `8721528`, in both places. Every manager and every `matchStrings` entry is now validated — and in the Makefile guard, an unmatched matchString is itself an assertion failure rather than silently contributing nothing, since a regex matching nothing is precisely the failure mode the docstring names. Same loop applied to the CONTRIBUTING.md parity guard. Resolving.

.github/workflows/agent-ci.yml
jainal09: Fixed in `8721528` — took your reflow verbatim. The half-applied look was real: an earlier scripted splice inserted new lines mid-sentence at the wrong column. Worth fixing before it fossilizes, as you say. Resolving.

renovate.json
jainal09: Agreed on all points, including that nothing changes today — recorded in `8721528` where the major-bump reviewer will actually see it: the CONTRIBUTING.md customManager's `description` now carries a MAJOR-BUMP NOTE stating that a golangci-lint v3 PR must move the `/v2/` module path in BOTH the doc command and the matchStrings regex, and — your key observation — that the parity guard stays green when both copies break identically, so the human reviewer is the control for that case. Majors are manual-review here, so the note sits directly in the config they will be reading. Resolving.

tests/unit/test_renovate_coverage.py
```diff
@@ -0,0 +1,460 @@
...
+        if m.get("datasourceTemplate") == "github-releases" and m.get("depNameTemplate")
+    ]
+
+
+def _constants_managers() -> list[dict[str, Any]]:
+    """github-releases managers that pin tools via constants.json.
+
+    Only these carry download URLs to cross-check. Managers that pin a command
+    in docs (the golangci-lint CONTRIBUTING.md one) have no artifact URLs by
+    design and are covered by the live upstream-tag test instead.
```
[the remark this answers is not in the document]
jainal09: Confirmed on the diagnosis, fixed differently on the prescription — `911b477`.

You were right that the doc-pin manager had **zero** upstream-reality coverage: the live test's `if not extract: continue` skipped it, and I had scoped it out of the URL guards. The one manager most structurally similar to the broken Infisical config was the one nothing watched. Good catch.

But the suggested `extractVersionTemplate: ^v(?<version>.*)$` would break the pin it protects: with an extract template, Renovate writes back the **extracted** (unprefixed) version, and Go rejects those —

```
$ go install golang.org/x/vuln/cmd/govulncheck@1.6.0
invalid version: unknown revision 1.6.0
```

That is precisely why this manager embeds the tag verbatim (`@v2.12.2`) with no template — the same reason constants.json stores unprefixed versions and its URL templates re-add `v{version}`.

So the fix is in the test: the live upstream-tag check no longer skips no-template managers; for them it asserts real release tags match the verbatim v-shape (`^v\d+\.\d+\.\d+`). A repo move, tag-format change, or dead stream now fails loud for this manager exactly as for the others, and the `_constants_managers` docstring now says what is actually true. Resolving.

tests/unit/test_renovate_coverage.py
```diff
@@ -0,0 +1,488 @@
...
+        and any("workflows" in pat for pat in m.get("managerFilePatterns", []))
+    ]
+    assert managers, "renovate.json lost the workflow go-install customManager"
+    workflow_text = (_WORKFLOWS / "agent-ci.yml").read_text(encoding="utf-8")
+    for manager in managers:
+        for match_string in manager["matchStrings"]:
+            pattern = re.sub(r"\(\?<(\w+)>", r"(?P<\1>", match_string)
+            found = re.search(pattern, workflow_text)
+            assert found, f"go-install manager regex matches nothing: {match_string!r}"
+            assert found.group("depName") == "golang.org/x/vuln"
```
jainal09, line 418: Fixed by splitting, not suppressing — `911b477`. The finding was accurate: that test had accreted five concerns across three review rounds (floor, matrix, selected x/sys, two gates' rotation-survival, the windows pass, and the manager regex).

Now four focused tests over three small helpers: `test_go_floor_supports_current_x_sys` (back to its original scope), `test_agent_gates_survive_matrix_rotation` (parametrised over both gates), `test_govulncheck_gate_analyses_the_windows_build`, and `test_go_install_manager_matches_the_workflow`. Behaviour identical — 27 guards green. Resolving.

tests/unit/test_renovate_coverage.py
```diff
@@ -0,0 +1,488 @@
+"""Renovate configuration must match upstream reality.
```
jainal09, line 1: Addressed by the same split as the Complex Method finding — `911b477`. The mean was dragged by one complexity-14 function; decomposing it into four flat tests (plus extracting `_agent_ci`/`_agent_matrix_go`/`_agent_step_running`, each trivial) brings the outlier down without gaming the metric. Resolving.

envdrift-agent/go.mod
```diff
@@ -1,6 +1,11 @@
...
+
+// Renovate-tracked (gomod manager, depType toolchain). Floor for the
+// GO-2026-4602 stdlib fix (os@go1.25.8): building with an older toolchain
+// ships the vulnerable os package inside the released agent binaries.
+toolchain go1.25.12
```
jainal09: [codex-review] **[P1] Enforce the patched standard library as the minimum Go version**

The `toolchain` directive is only a suggested toolchain; the `go` directive remains the mandatory minimum, and `GOTOOLCHAIN=local` uses the bundled compiler. Consequently this module still accepts Go 1.25.0–1.25.7 even though GO-2026-4602 affects versions before 1.25.8. I reproduced this with Go 1.25.4: `GOTOOLCHAIN=local go test -mod=readonly ./...` passes. The new govulncheck job runs only on the 1.26 matrix leg, and the static guard ignores both the patch version and this directive, so CI cannot catch a release built with the vulnerable toolchain. Raise the `go` directive to the patched floor (preferably 1.25.12 to match this selection) and update the contributor prerequisite/guard accordingly.

Go toolchain semantics: https://go.dev/doc/toolchain
Advisory: https://pkg.go.dev/vuln/GO-2026-4602

jainal09: Confirmed and fixed in `72f8371` — and this is the second time this PR's Go-floor work has been caught half-done, so thank you for the persistence. Reproduced your exact scenario first:

```
$ GOTOOLCHAIN=local go build ./...      # go1.25.4 bundled compiler
(builds — GO-2026-4602-vulnerable os package linked)
```

You are right on the semantics: the `toolchain` directive is only a suggestion, `GOTOOLCHAIN=local` uses the bundled compiler, and only the `go` directive is mandatory. My previous fix covered the default (`auto`) path and left hermetic/corporate builders exposed.

Taken as prescribed: the `go` directive is now **1.25.12** (matching the prior toolchain selection), the redundant `toolchain` line is dropped (Renovate tracks the go directive natively as depType `golang`), and the same tree now refuses the vulnerable compiler loudly:

```
go: go.mod requires go >= 1.25.12 (running go 1.25.4; GOTOOLCHAIN=local)
```

Default builds stay green — 11 packages, tidy clean, both govulncheck passes clean.

And your point about the guard was the root cause of the miss: it compared only `(major, minor)`, which is exactly how the hole survived it. It now asserts the full directive ≥ 1.25.8 (the advisory floor) with the toolchain-vs-directive rationale inline, and was verified to fail on a pretended `go 1.25.0`. Contributor prereq updated to 1.25.12+ with a note that older toolchains are refused. Resolving.

tests/unit/test_renovate_coverage.py
```diff
@@ -0,0 +1,488 @@
...
+    # zero diagnostics — the silent-guard-death mode this module exists to
+    # prevent.
+    for gate, needle in (
+        ("go mod tidy", "go mod tidy -diff"),
+        ("govulncheck", "govulncheck ./..."),
+    ):
+        steps = [step for step in ci["jobs"]["test"]["steps"] if needle in str(step.get("run", ""))]
+        assert steps, f"agent-ci.yml lost the {gate} gate"
+        condition = str(steps[0].get("if", ""))
+        assert any(f"'{leg}'" in condition for leg in matrix_go), (
```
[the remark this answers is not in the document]
jainal09: Fixed in `72f8371` — both points taken exactly as suggested.

You are right that the misleading direction was the worse half: an absent `if:` means the gate runs on **every** leg (the more robust form), and my guard failing that with "the gate would never run again" would have actively pushed future work away from the better design. An absent condition now passes with a comment saying why, and pinned legs are accepted in either quote style (`'1.26'` or `"1.26"`), so only genuinely stale pinned legs fail. Resolving.

envdrift-agent/go.mod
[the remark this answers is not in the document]
jainal09: Already fixed — this review ran against the pre-`72f8371` head, and that commit took exactly the option you recommend ("raising the `go` directive itself"):

- `go` directive is now **1.25.12**; the `toolchain` line is gone entirely (it was only a suggestion that `GOTOOLCHAIN=local` ignores — a codex thread reproduced that hole with a 1.25.4 bundled compiler).
- The guard is patch-aware and asserts the full directive ≥ **1.25.8** (the GO-2026-4602 floor), with the toolchain-vs-directive rationale inline. Verified it fails on a pretended `go 1.25.0`.
- Enforcement is loud: `go: go.mod requires go >= 1.25.12 (running go 1.25.4; GOTOOLCHAIN=local)`.

So both the directive and the guard now enforce the documented floor; there is no toolchain line left to remove or lower. Resolving as already-fixed.

.github/workflows/agent-ci.yml
```diff
@@ -40,13 +40,38 @@ jobs:
...
+        run: go mod tidy -diff
+
       - name: Download dependencies
         run: go mod download
 
       - name: Run tests
         shell: bash
         run: go test -v -race -coverprofile=coverage.out ./...
 
+      - name: govulncheck
```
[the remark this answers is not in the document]
jainal09: Correct, and fixed in `02e2c66` with the stronger of your two options. You are right that `GOOS=windows GOARCH=amd64 govulncheck ./...` is source-mode analysis under windows build constraints — my comment claiming it analysed "the released windows-amd64 binary" oversold it.

Since binary-level coverage was the honest goal, the gate now builds a windows-amd64 exe from `./cmd/envdrift-agent` — the same package `agent-release.yml` ships — and runs `govulncheck -mode=binary` against it. Verified locally: the exe builds cross-platform (8.4M) and the binary scan reports "No vulnerabilities found." The native source pass stays for call-graph reachability on the host.

The guard now also requires `-mode=binary`, so a future simplification back to source-only analysis fails loudly instead of quietly overselling what is audited. Resolving.

